PR #52, opened 2022-09-11
feat: switch to Go CLI for deployment

mnaser:
This is an initial branch to move some of the Ansible based stuff for deployment the service to a simply Go binary that can run as a cluster job or as a CLI.

It will also simplify testing the OpenStack parts since we can just setup KinD and deploy from there, there's a bunch of work left here:

- [ ] Add proper coverage for unit tests
- [ ] Add at least one other service
- [ ] Add `validate` command to do some sanity tests
- [ ] Add `diff` command to show diffs before applying
- [ ] Add `settle` command to wait for all pods to be ready


mnaser:
I think one of the challenges that we'll probably have to deal with now is how this is deployed.

I think we'll switch to using the `controller-runtime` and make it a controller that can run in a `--one-shot` option that will start up, reconcile and shut down -- or one that watches the config and keeps reconciling.

mnaser:
> I think one of the challenges that we'll probably have to deal with now is how this is deployed.
> 
> I think we'll switch to using the `controller-runtime` and make it a controller that can run in a `--one-shot` option that will start up, reconcile and shut down -- or one that watches the config and keeps reconciling.

Now the only thing with that is, we lose the ability to do things like `--diff` or something like that if we go towards a more server-based instead of CLI based.

mnaser:
I think the way to go is to stop using `Ansible-playbook` as an entry point and instead use `atmosphere` as an entry point, it will fork out to Ansible to run the playbooks and then do the missing pieces.

PR #54, opened 2022-09-13
Integrate Designate into Atmosphere

mpiscaer:
Integrate Designate into Atmosphere


roles/openstack_helm_designate/defaults/main.yml
```diff
@@ -0,0 +1,61 @@
...
+# .. envvar:: openstack_helm_designate_chart_name [[[
+#
+# Helm chart name (will also be used for release name)
+openstack_helm_designate_chart_name: designate
+
+                                                                   # ]]]
+# .. envvar:: openstack_helm_designate_image_repository [[[
+#
+# Image repository location to be prefixed for all images
+openstack_helm_designate_image_repository: "{{ atmosphere_image_repository | default('us-docker.pkg.dev/vexxhost-infra/openstack') }}"
```
mnaser, line 35: We've swapped this over to `quay.io/vexxhost` and the image builds are happening here now:

https://github.com/vexxhost/docker-openstack-designate

roles/openstack_helm_designate/defaults/main.yml
```diff
@@ -0,0 +1,61 @@
...
+
+                                                                   # ]]]
+# .. envvar:: openstack_helm_designate_image_tag [[[
+#
+# Image tag for container
+openstack_helm_designate_image_tag: 0.2.7
```
mnaser, line 41: I suggest that you update this to say `wallaby` instead, since that will be the newest promoted build.

roles/openstack_helm_designate/meta/main.yml
```diff
@@ -0,0 +1,29 @@
...
+  platforms:
+    - name: Ubuntu
+      versions:
+        - focal
+
+dependencies:
+  - role: helm_repository
+    vars:
+      helm_repository_name: "{{ openstack_helm_designate_chart_repo_name }}"
+      helm_repository_repo_url: "{{ openstack_helm_designate_chart_repo_url }}"
```
mnaser, line 29: We've moved away from this role, instead, we're using `HelmRelease` and `HelmRepository` which uses FluxCD and instead delegates the application of all the resources to the server side.  Therefore, this should be removed

roles/openstack_helm_designate/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+    openstack_helm_endpoints_chart: "{{ openstack_helm_designate_chart_name }}"
+
+- name: Deploy Helm chart
+  kubernetes.core.helm:
+    name: "{{ openstack_helm_designate_chart_name }}"
+    chart_ref: "{{ openstack_helm_designate_chart_repo_name }}/{{ openstack_helm_designate_chart_name }}"
+    chart_version: 0.2.7
+    release_namespace: openstack
+    kubeconfig: /etc/kubernetes/admin.conf
+    values: "{{ _openstack_helm_designate_values | combine(openstack_helm_designate_values, recursive=True) }}"
```
mnaser, line 30: With the change to Flux CD, things look a little different, see the following:

https://github.com/vexxhost/atmosphere/blob/d9e15a340da407775a1ce04877a7c85871dd72cb/roles/openstack_helm_barbican/tasks/main.yml#L23-L76

roles/openstack_helm_endpoints/vars/main.yml
```diff
@@ -299,6 +299,35 @@ _openstack_helm_endpoints_dns:
...
+  scheme:
+    default: 'tcp'
+  port:
+    ipc:
+      default: 5354
+
+_openstack_helm_endpoints_powerdns:
+  auth:
+    service:
+      token: chiave_segreta
```
mnaser, line 320: Shouldn't this be a secret value?

mnaser (changes requested):

roles/openstack_helm_endpoints/vars/main.yml
okozachenko1203: removed `_openstack_helm_endpoints_powerdns` because we don't install powerdns using osh for now.

mnaser (approved):

PR #61, opened 2022-09-16
feat: add simple controller to generate helm values

mnaser:
This is the initial setup to add a very simple controller on the server side which will generate secrets with Helm values.  The initial scope is to simply generate the `values.yaml` for Memcached and then it can slowly take on generating all of the endpoints.

- [x] Add GitHub actions for unit tests
- [x] Add GitHub actions for integration tests
- [x] Add GitHub actions for linters (black + pep8 or some other form of that)
- [x] Write Dockerfile for image
- [x] Add feature to allow for overrides
- [x] Write E2E tests using `kind`
- [x] Write GitHub actions for E2E tests
- [x] Figure out solution for running with Molecule (likely use [ttl.sh](https://ttl.sh))


mnaser:
pre-commit.ci run

PR #62, opened 2022-09-16
Set image_conversion_disable to true by default

ricolin:
Adopt config image_conversion_disable from [1].
Set it to true by default.

[1] https://review.opendev.org/c/openstack/cinder/+/839793


ricolin:
For https://github.com/vexxhost/system-config/issues/191

PR #93, opened 2022-09-28
[tox] Add toml as dependency for molecule

okozachenko1203:


mnaser:
@okozachenko1203 hmm, I've only been using `poetry` to run this, let me think about this. bit more.

mpiscaer:
This was needed for me inside the venv on Focal.

mnaser:
> This was needed for me inside the Tox venv on Focal.

indeed, but we've shifted things towards Poetry (granted, we haven't actually properly and cleanly documented the switch over).

I'll follow this up with some changes to docs as well as removing the leftover `tox` stuff, Poetry is much nicer to work with overall.

mnaser:
tox.ini is gone!

PR #120, opened 2022-10-10
feat: allow for a custom cluster IP address for the neutron coredns service

runlevel-six:
Default is set to '10.96.0.20' (matching current configuration). This will support Atmosphere running the 'openstack' playbook against a Kubernetes cluster that is not using the default `10.96.0.0/12` implemented with `kubeadm`.

### DIFF
--- a/roles/coredns/tasks/main.yml
+++ b/roles/coredns/tasks/main.yml
@@ -37,7 +37,7 @@
                 enabled: true
             service:
               name: neutron-coredns
-              clusterIP: 10.96.0.20
+              clusterIP: "{{ neutron_coredns_cluster_ip | default('10.96.0.20') }}"
             isClusterService: false
             servers:
               - port: 53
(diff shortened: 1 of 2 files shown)


mnaser (changes requested):
https://github.com/vexxhost/atmosphere/blob/cc2d497ae7546c2b33eb016f20f17f0c2b4b194c/roles/openstack_helm_neutron/vars/main.yml#L66

can you update that as well?  since if you modify it there now neutron won't be pointing to the right IP :)

runlevel-six:
> https://github.com/vexxhost/atmosphere/blob/cc2d497ae7546c2b33eb016f20f17f0c2b4b194c/roles/openstack_helm_neutron/vars/main.yml#L66
> 
> can you update that as well? since if you modify it there now neutron won't be pointing to the right IP :)

Thanks for catching that - I have updated the PR with that change.

mnaser (approved):

PR #190, opened 2022-11-22
chore: standardized hostname naming convention in Octavia role

runlevel-six:
changed octavia configuration to consistently use hostvars[item]['inventory_hostname_short'] instead of different hostvar variations.


runlevel-six:
Closing as this change can introduce a configuration that can break at least some load balancer configurations.

PR #191, opened 2022-11-22
feat: allow for custom clusterissuer and certificate secretName when atmosphere is not responsible for cert-manager

runlevel-six:
Defaults for `cert-manager.io/cluster-issuer` remains `atmosphere` and for `secretName` remains `{{ openstack_helm_ingress_service_name }}-certs`.

This allows for those defaults to be overridden by setting the `openstack_helm_ingress_secretname` and `openstack_helm_ingress_clusterissuer` variables in the cluster inventory.

### DIFF
--- a/roles/openstack_helm_ingress/tasks/main.yml
+++ b/roles/openstack_helm_ingress/tasks/main.yml
@@ -29,7 +29,7 @@
             http:
               paths: "{{ _openstack_helm_ingress_paths }}"
         tls:
-          - secretName: "{{ openstack_helm_ingress_secretname | default(openstack_helm_ingress_service_name + '-certs') }}"
+          - secretName: "{{ openstack_helm_ingress_secret_name | default(openstack_helm_ingress_service_name + '-certs') }}"
             hosts:
               - "{{ openstack_helm_endpoints[openstack_helm_ingress_endpoint]['host_fqdn_override']['public']['host'] }}"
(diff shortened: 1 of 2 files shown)


roles/openstack_helm_ingress/tasks/main.yml
```diff
@@ -29,7 +29,7 @@
             http:
               paths: "{{ _openstack_helm_ingress_paths }}"
         tls:
-          - secretName: "{{ openstack_helm_ingress_service_name }}-certs"
+          - secretName: "{{ openstack_helm_ingress_secretname | default(openstack_helm_ingress_service_name + '-certs') }}"
```
mnaser, line 32: Could we make this "openstack_helm_ingress_secret_name" instead?

roles/openstack_helm_ingress/vars/main.yml
```diff
@@ -13,7 +13,7 @@
 # under the License.
 
 _openstack_helm_ingress_annotations:
-  cert-manager.io/cluster-issuer: atmosphere
+  cert-manager.io/cluster-issuer: "{{ openstack_helm_ingress_clusterissuer | default('atmosphere') }}"
```
mnaser, line 16: Could we do "openstack_helm_ingress_cluster_issuer"?

mnaser (changes requested):
Small kits, otherwise looks good. If you don't manage to take care of them, I'll try to find time to finalize and land those changes.

mnaser (approved):

PR #192, opened 2022-11-23
feat: custom octavia manageament subnet cidr

runlevel-six:
Allow for a custom cidr for use by octavia's management subnet in case there is a conflict or overlap with the default. The default remains 172.24.0.0/22.

### DIFF
--- a/roles/openstack_helm_octavia/defaults/main.yml
+++ b/roles/openstack_helm_octavia/defaults/main.yml
@@ -56,7 +56,7 @@ openstack_helm_octavia_heartbeat_key: "{{ undef(hint='You must specify a Octavia
 # .. envvar:: openstack_helm_octavia_management_subnet_cidr [[[
 #
 # Octavia management subnet (CIDR)
-openstack_helm_octavia_management_subnet_cidr: "172.24.0.0/22"
+openstack_helm_octavia_management_subnet_cidr: "{{ octavia_management_subnet_cidr | default('172.24.0.0/22') }}"
 
                                                                    # ]]]
 # .. envvar:: openstack_helm_octavia_amphora_image_url [[[


mpiscaer:
Correct me if I'm wrong but if the variable is in set in default isn't it already changeable?

runlevel-six:
> Correct me if I'm wrong but if the variable is in set in default isn't it already changeable?

There are some variables in the defaults/main.yml files that have an override but this was not one of them. An example of one that has it is on line 17:

`openstack_helm_octavia_image_repository: "{{ atmosphere_image_repository | default('quay.io/vexxhost') }}"`

I just extended that same logic to the `openstack_helm_octavia_management_subnet_cidr` variable

roles/openstack_helm_octavia/defaults/main.yml
```diff
@@ -56,7 +56,7 @@ openstack_helm_octavia_heartbeat_key: "{{ undef(hint='You must specify a Octavia
 # .. envvar:: openstack_helm_octavia_management_subnet_cidr [[[
 #
 # Octavia management subnet (CIDR)
-openstack_helm_octavia_management_subnet_cidr: "172.24.0.0/22"
+openstack_helm_octavia_management_subnet_cidr: "{{ octavia_management_subnet_cidr | default('172.24.0.0/22') }}"
```
mnaser, line 59: You should be able to simply add "openstack_helm_octavia_management_subnet_cidr" into your variables and you should be good to go.

The reason we've used the other model of variable was because it was a variable used across multiple roles.

mnaser (changes requested):
I suspect this might not be necessary but I could be misunderstanding the root cause here

runlevel-six:
@mpiscaer and @mnaser are correct. This is an unnecessary change. Closing the PR.

PR #195, opened 2022-11-25
Tuning Ceph OSD sysctl

ricolin:
Tuning sysctl on Ceph OSD nodes.

The tuning factors and values are reference from:
https://opendev.org/openstack/charm-ceph-osd/src/branch/master/lib/charms_ceph/utils.py#L115-L144


mnaser (changes requested):
We actually want to run those tunes on all systems since the neutron routers benefit from this.

mnaser (changes requested):
since we're creating a new sysctl role instead, let's move the highlighted sysctls in here:

https://github.com/vexxhost/atmosphere/blob/3bb608b1ca03676411d73d36d4f1185801b74f4e/roles/kubernetes/defaults/main.yml#L49-L60

the highlighted ones from L49 to L60 are performance tweaks and not needed for k8s.  ALSO, while you're at it, can you remove this

https://github.com/vexxhost/atmosphere/blob/3bb608b1ca03676411d73d36d4f1185801b74f4e/roles/kubernetes/defaults/main.yml#L39-L42

L39-L42 was back when we were trying to do work with VRFs and it is not necessary since we don't use VRFs in the cloud anymore.

mnaser (approved):

PR #200, opened 2022-11-29
Remove helm default policy

ricolin:
This fix bump version of helm charts to get the fix for remove default policy.

Ref: https://review.opendev.org/c/openstack/openstack-helm/+/865347


mnaser (changes requested):
The following services need to be also bumped:

- Glance
- Neutron
-

mnaser:
handled by https://github.com/vexxhost/atmosphere/commit/bcdd25c47fc1f254b4375829b247dc7bc9243db7

PR #207, opened 2022-12-05
Remove rbd write-cache docs

guilhermesteinmuller:
After some usage of rbd write feature, we came to
the conclusion that it is not mature yet to work
with openstack.

The issue we currently see is when a vm with write-cache enabled is rebooted, it doesnt come back and error with https://paste.opendev.org/raw/817888/

We dont recomment the enablement of that feature for now.


mnaser:
sad but true

PR #217, opened 2022-12-12
feat: add cloud resource + magnum

mnaser:
Fix https://github.com/vexxhost/atmosphere/issues/14


mnaser:
@okozachenko1203 the kvm issue happens when an atmosphere env stays on for a while. Restarting libvirt fixes it. 

I just landed the change in magnum capi. I will make sure release goes out.

mnaser:
Can you add a masquerade rule to the Nolecule playbooks?

roles/openstack_helm_magnum/vars/main.yml
```diff
@@ -0,0 +1,85 @@
...
+      octavia_client:
+        endpoint_type: internalURL
+        region_name: "{{ openstack_helm_endpoints['identity']['auth']['magnum']['region_name'] }}"
+      oslo_messaging_notifications:
+        driver: noop
+      trust:
+        cluster_user_trust: true
+  manifests:
+    ingress_api: false
+    service_ingress_api: false
```
mnaser, line 85: drop this whole file, its not being used.

roles/openstack_helm_endpoints/vars/main.yml
```diff
@@ -438,3 +438,38 @@ _openstack_helm_endpoints_tempest:
...
+  container_infra_registry:
+    scheme:
+      public: https
+    host_fqdn_override:
+      public:
+        host: "{{ openstack_helm_endpoints_magnum_registry_host }}"
+    port:
+      api:
+        public: 443
+
```
mnaser, line 467: please remove this extra space

atmosphere/tasks/constants.py
```diff
@@ -440,3 +440,5 @@
 HELM_RELEASE_DESIGNATE_NAME = "designate"
 
 HELM_RELEASE_HEAT_NAME = "heat"
+
+HELM_RELEASE_MAGNUM_NAME = "magnum"
```
mnaser, line 444: is this being used anywhere?

mnaser (commented):
@okozachenko1203 ok so, new Magnum image are built with 0.2.5 CAPI that has the fix you mentioned, please address these

okozachenko1203:
> Can you add a masquerade rule to the Nolecule playbooks?

@mnaser https://github.com/vexxhost/atmosphere/pull/222/files

atmosphere/tasks/constants.py
okozachenko1203: https://github.com/vexxhost/atmosphere/pull/217/files#diff-8ae6ef538f496abc69c5cbbd4923c9dce3c4f7a54b985f851609fe3248b8b481R138-R140

atmosphere/flows.py
```diff
@@ -135,6 +135,9 @@ def get_deployment_flow(config):
         openstack_helm.ApplyRabbitmqClusterTask(
             name=constants.HELM_RELEASE_HEAT_NAME,
         ),
+        openstack_helm.ApplyRabbitmqClusterTask(
+            name=constants.HELM_RELEASE_MAGNUM_NAME,
+        ),
```
mnaser, line 140: this is not needed, we create this in the magnum flow.

atmosphere/operator/controllers/cloud.py
```diff
@@ -0,0 +1,79 @@
...
+        tasks.GenerateImageTagsConfigMap(provides="image_tags"),
+        tasks.GenerateSecrets(provides="secrets"),
+    )
+
+    if spec["magnum"].get("enabled", True):
+        flow.add(
+            tasks.InstallClusterApiTask(),
+            tasks.ApplyRabbitmqClusterTask(
+                inject={"chart_name": "magnum"}, provides="magnum_rabbitmq"
+            ),
```
mnaser, line 26: because we have this here. :)

atmosphere/operator/tasks.py
```diff
@@ -0,0 +1,618 @@
...
+        return {}
+
+    def _generate_container_infra(self):
+        return {}
+
+    def execute(self, *args, **kwargs):
+        endpoints = (
+            self._get_values().get("endpoints", {}).keys() - self.SKIPPED_ENDPOINTS
+        )
+        return {"endpoints": {k: getattr(self, "_generate_" + k)() for k in endpoints}}
```
mnaser, line 618: sorry this is something I added but not needed anymore.

atmosphere/tasks/constants.py
mnaser: no need for that, since we are creating it elsewhere, see comment above.

roles/atmosphere/templates/cluster_role_binding.yml
```diff
@@ -9,5 +9,5 @@ subjects:
     namespace: openstack
 roleRef:
   kind: ClusterRole
-  name: atmosphere
+  name: cluster-admin
```
mnaser, line 12: what's the reasoning behind this?  if we give atmosphere `cluster-admin` then all the RBAC stuff we do is meaningless

okozachenko1203: @mnaser we install capi in this operator and it means it requires permissions(at least get, patch, create) for resource types  required to install capi. It includes many resource types such as rbac stuff. So i thought it is better to use cluster-admin

mnaser:
looks like we need to actually wait for wsrep to be up to avoid race conditions.. but we're good for now.

PR #221, opened 2022-12-13
chore: Set nova virt_type as qemu in molecule env

okozachenko1203:
To fix `qemu-system-x86_64: failed to initialize KVM: Permission denied` error in nova-compute


mnaser (changes requested):
This will slow things down majorly since it will not use KVM anymore. 

The solution for this issue in my experience is to restart libvirt pods.

PR #222, opened 2022-12-13
fix: Add masquerade rule in molecule env

okozachenko1203:


molecule/default/prepare.yml
```diff
@@ -19,3 +19,11 @@
...
       become: yes
       command: apt-get update
+
+- hosts: controllers
+  tasks:
+    - name: Set masquerade rule
+      become: yes
+      ansible.builtin.shell: |
+        iptables -t nat -A POSTROUTING -s  10.96.250.0/24  -o ens3 -j MASQUERADE
+        iptables -t nat -A POSTROUTING -s  10.96.250.0/24  -o ens4 -j MASQUERADE
```
mnaser, line 29:
```suggestion
      ansible.builtin.iptables:
        table: nat
        chain: POSTROUTING
        source: 10.96.250.0/24
        out_interface: "{{ ansible_default_ipv4.interface }}"
        jump: MASQUERADE
```

This will be cleaner overall since it will create it only if needed, and also it will use the default interface (with default route) and masquerade that only instead.

mnaser (changes requested):

mnaser:
This does the job, but we need to figure out why the cert manager stuff isn't working, landed this.

PR #232, opened 2022-12-18
feat: allow airgapped installs

mnaser:


atmosphere/operator/controllers/cloud.py
```diff
@@ -91,6 +91,15 @@ def create_fn(namespace: str, name: str, spec: dict, **_):
             ),
         )
 
+    flow.add(
+        tasks.ApplyNamespaceTask(
+            name=constants.NAMESPACE_OPENSTACK,
+            provides="openstack_namespace",
+        ),
```
mnaser: no need for this, we already have `namespace` which is created earlier :)

atmosphere/operator/controllers/cloud.py
```diff
@@ -91,6 +91,15 @@ def create_fn(namespace: str, name: str, spec: dict, **_):
...
+        tasks.ApplyHelmReleaseTask(
+            config=constants.HELM_RELEASE_INGRESS_NGINX,
+        ),
```
mnaser, line 97: this creates a regression, it is not possible to "disable" ingress-nginx with this...

atmosphere/operator/controllers/cloud.py
```diff
@@ -177,11 +186,16 @@ def create_fn(namespace: str, name: str, spec: dict, **_):
         flow.add(tasks.ApplyRabbitmqClusterTask("heat"))
 
     if spec["monitoring"]["enabled"]:
+        # TODO(oleks): Enable opsgenie
```
mnaser, line 185: and make sure there's a way to do overrides too

atmosphere/operator/controllers/cloud.py
mnaser: and make sure there's a way to do overrides too

okozachenko1203:
@mnaser i cannot resolve these conflicts because it doesn't update head

mnaser:
this is already implemented via the `image_manifest` role:

https://github.com/vexxhost/atmosphere/blob/90f91b3dda9fed24bad292cf232cce9259d70635/roles/image_manifest/README.md#L4

PR #243, opened 2022-12-28
feat: Support offline helm repositories

okozachenko1203:


mnaser:
@okozachenko1203 

- can you make sure to only install 1 instance of `HelmRepository` inside `openstack` namespace?
- you can use `namespace` inside the `sourceRef` for `HelmRelease` which will allow you t ouse one only :)

mnaser:
@okozachenko1203 can you please finish this up? I think the integration won't work anymore since the charts are no longer hosted on-line, so you'll have to rely on the e2e tests only.

lets get this landed quickly as I don't think there's much left here.

okozachenko1203:
duplicated one with https://github.com/vexxhost/atmosphere/pull/252

PR #244, opened 2022-12-28
fix: Fix test failures

okozachenko1203:
- install extra dependency in CI
- add exception for config file missing
- use jinja2_ansible_filters for ansible template


mnaser:
@okozachenko1203 can you remove the jinja2 base64 filters then?

okozachenko1203:
> @okozachenko1203 can you remove the jinja2 base64 filters then?

I think i did it already, no?

okozachenko1203:
> > @okozachenko1203 can you remove the jinja2 base64 filters then?
> 
> I think i did it already, no?

@mnaser ok, i will remove it from poetry

mnaser (approved):

PR #251, opened 2023-01-03
fix: Fix the conditional deployment of rabbitmq clusters

okozachenko1203:


mnaser:
this happens now inside the `openstack_helm_endpoints` role.

PR #252, opened 2023-01-03
feat: Use offline helm repositories

okozachenko1203:


mnaser:
Could you merge main to this?

mnaser:
I think we're going to close this, the idea is to do something like this instead:

https://github.com/vexxhost/atmosphere/commit/94e07e101211d55c405716bdfe7741495c810963

PR #261, opened 2023-01-12
fix: Add crane install step in custom registry doc

okozachenko1203:


mnaser:
Thanks for taking care of this but I've folded this change into the documentation change for the image manifest role, so we can close this.

PR #263, opened 2023-01-15
feat: Make atmosphere_images overrideble.

mpiscaer:
Make atmosphere_images overrideble. This can be used to downgrade images  or build local images


mnaser:
@mpiscaer do I understand that the main purpose of this change is to allow you to manually control the images that are being mirrored so you can do it in-band with atmosphere instead of doing it out of band with your own toolkit and then updating the manifest !

PR #270, opened 2023-01-20
Fix: Specify the gallaxy collection for the filter.

mpiscaer:
The filter openstack_helm_image_tags & docker_image now not specifying the galaxy and uses the global filter set. When not including this in the ANSIBLE_FILTER_PLUGINS ENV variable and only do the ansible gallaxy install it will never find these filters.


mnaser:
thanks for this catch.

Issue #272, opened 2023-01-23
Name change in Openstack-helm at endpoints for Cinder


mpiscaer:
The change https://review.opendev.org/c/openstack/openstack-helm/+/869994/3/cinder/values.yaml#1262 at 20 jan they merged this change. With this change the tasks for creating endpoint at cinder failes:
```
TASK [vexxhost.atmosphere.openstack_helm_endpoints : Generate OpenStack-Helm endpoints] ***************************************************************************************************************************************************************************************
ok: [fctrko1.maas] => (item=identity)
ok: [fctrko1.maas] => (item=image)
fatal: [fctrko1.maas]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: No variable found with this name: _openstack_helm_endpoints_volume. No variable found with this name: _openstack_helm_endpoints_volume\n\nThe error appears to be in '/home/gitlab-runner/.ansible/collections/ansible_collections/vexxhost/atmosphere/roles/openstack_helm_endpoints/tasks/main.yml': line 74, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Generate OpenStack-Helm endpoints\n  ^ here\n"}
```

mnaser:
solved by https://github.com/vexxhost/atmosphere/commit/553e675132b32f495d0020f536dec335941e8d8c since we're not using the latest values for now.

PR #276, opened 2023-01-23
fix: var field name for openstack volume endpoint

okozachenko1203:
fix https://github.com/vexxhost/atmosphere/issues/272


mnaser:
this is now technically broken since https://github.com/vexxhost/atmosphere/commit/553e675132b32f495d0020f536dec335941e8d8c -- we should adjust once we have that updated commit in :)

PR #279, opened 2023-01-24
chore: move designate to offline-install

guilhermesteinmuller:


okozachenko1203:
we don't deploy designate in atmosphere yet. i will merge it without test for now

PR #280, opened 2023-01-24
chore: move glance to offline-install

guilhermesteinmuller:


roles/openstack_helm_glance/meta/main.yml
```diff
@@ -25,4 +25,10 @@ galaxy_info:
 
 dependencies:
   - role: defaults
-  - role: openstacksdk
```
mnaser, line 28: This needs to be restored.

mnaser (changes requested):

mnaser (approved):

PR #284, opened 2023-01-24
chore: move nova to offline-install

guilhermesteinmuller:


roles/openstack_helm_nova/meta/main.yml
```diff
@@ -25,5 +25,11 @@ galaxy_info:
...
   - role: defaults
-  - role: atmosphere
+  - role: openstack_helm_endpoints
+    vars:
+      openstack_helm_endpoints_chart: nova
+  - role: upload_helm_chart
+    vars:
+      upload_helm_chart_src: "{{ openstack_helm_nova_helm_chart_path }}"
+      upload_helm_chart_dest: "{{ openstack_helm_nova_helm_chart_ref }}"
   - role: openstacksdk
```
okozachenko1203, line 35: `openstacksdk` should be executed before `openstack_helm_endpoints` because `openstacksdk` itself includes   `openstack_helm_endpoints` role with `openstack_helm_endpoints_list: ["identity"]` var as subtask. And it will regenerate `openstack_helm_endpoints` which includes only `identity` section.

```suggestion
  - role: openstacksdk
  - role: openstack_helm_endpoints
    vars:
      openstack_helm_endpoints_chart: nova
  - role: upload_helm_chart
    vars:
      upload_helm_chart_src: "{{ openstack_helm_nova_helm_chart_path }}"
      upload_helm_chart_dest: "{{ openstack_helm_nova_helm_chart_ref }}"
```

PR #286, opened 2023-01-24
chore: move ceph_provisioners to offline-install

guilhermesteinmuller:


roles/openstack_helm_endpoints/vars/main.yml
```diff
@@ -13,6 +13,7 @@
 # under the License.
 
 _openstack_helm_endpoints_ignore:
+  - ceph_mon
```
okozachenko1203, line 16: @mnaser i ignored this

PR #294, opened 2023-01-26
feat: Support coredns offline-install

okozachenko1203:


mnaser:
Can you make sure you create a coredns_helm_values overrides and move things to vars

mnaser (changes requested):
We should make the values overridable

mnaser (approved):

PR #303, opened 2023-01-27
chore: Migrate cluster-api from atmosphere operator to magnum role

okozachenko1203:


roles/openstack_helm_magnum/tasks/main.yml
```diff
@@ -37,6 +37,19 @@
...
+    kubeconfig: /etc/kubernetes/admin.conf
+    state: present
+    template:
+      - path: 'templates/capi-bootstrap.yml.j2'
+      - path: 'templates/capi-control-plane.yml.j2'
+      - path: 'templates/capi-core.yml.j2'
+      - path: 'templates/capi-provider.yml.j2'
+      - path: 'templates/capi-rbac.yml.j2'
+  run_once: true
+  changed_when: false
```
okozachenko1203, line 51: There are some fields changed whenever apply so need to set this false manually.

- CRDs include conversion.webhook.clientConfig.caBundle which is injected by cert-manager after deployment. So it is always different.
- Rules of aggregated clusterRoles are rendered after clusterRole is deployed. So it is always different.

roles/openstack_helm_magnum/tasks/main.yml
```diff
@@ -358,15 +371,6 @@
               nodeSelector:
                 openstack-control-plane: enabled
 
-- name: Create Ingress
```
mnaser, line 361: Is this accidentally removed?

roles/openstack_helm_magnum/tasks/main.yml
mnaser: I think there might be a Kubernetes setting to ignore fields on change?  Or was that Terraform?

mnaser (changes requested):

roles/openstack_helm_magnum/tasks/main.yml
okozachenko1203: no, there were 2 duplicated tasks

roles/openstack_helm_magnum/tasks/main.yml
okozachenko1203: TF has but ansible doesn't

roles/openstack_helm_magnum/tasks/main.yml
mnaser: Are you sure?  One is for registry and one is for magnum..

okozachenko1203: yes, i am sure. there were 2 ingresses for magnum before

mnaser (approved):

PR #306, opened 2023-01-27
chore: migrate pushgateway from flux

okozachenko1203:


mnaser (changes requested):
Let's please use prometheus_pushgateway to match the role name as that is needed by ansible lint

mnaser (approved):

PR #326, opened 2023-02-15
fix: Fix image ref in kube_prometheus_stack role

okozachenko1203:
Also update the chart version


mnaser:
@okozachenko1203 why did we bump kube-prom-stack?  not against it, but just want to know why

okozachenko1203:
> @okozachenko1203 why did we bump kube-prom-stack? not against it, but just want to know why

well, nothing special :)
Just want to keep the latest.

PR #327, opened 2023-02-17
feat: allow using reader role

guilhermesteinmuller:


mnaser:
So for context, all the services we deploy that interact with Keystone:

- Keystone
- Barbican
- RadosGW (via Swift)
- Glance
- Cinder
- Placement
- Nova
- Neutron
- Senlin
- Designate
- Heat
- Octavia
- Magnum
- Horizon

The following are left after this change:

- Keystone
- RadosGW (via Swift)
- Placement
- Neutron
- Senlin
- Designate
- Octavia
- Magnum
- Horizon

We'll keep updating this list as we get it all locked up.

PR #328, opened 2023-02-17
feat: always disable swap on node startup

fitbeard:
Current bootstrap configuration for swap does not survive after node reboot. This change introduces new `systemd` service which runs `swapoff` before `kubelet` start.

### DIFF
--- a/roles/kubernetes/tasks/main.yml
+++ b/roles/kubernetes/tasks/main.yml
@@ -89,6 +89,15 @@
     - swap
     - none
 
+- name: Create noswap systemd service config file
+  ansible.builtin.copy:
+    src: noswap.service
+    dest: /etc/systemd/system/noswap.service
+    owner: root
+    group: root
+    mode: 0644
+  notify: Enable noswap service
+
 - name: Configure short hostname
   ansible.builtin.hostname:
     name: "{{ inventory_hostname_short }}"
--- a/roles/kubernetes/files/noswap.service
+++ b/roles/kubernetes/files/noswap.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Disable swap
+Before=kubelet.service
+After=local-fs.target
+
+[Service]
+Type=oneshot
+User=root
+ExecStart=/sbin/swapoff -a
+
+[Install]
+WantedBy=default.target
--- a/roles/kubernetes/handlers/main.yml
+++ b/roles/kubernetes/handlers/main.yml
@@ -1,0 +1,20 @@
+# Copyright (c) 2023 VEXXHOST, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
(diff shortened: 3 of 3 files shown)


roles/kubernetes/tasks/main.yml
```diff
@@ -89,6 +89,15 @@
     - swap
     - none
```
mnaser, line 90: we remove the swap from `/etc/fstab` here, just curious from where the swap would be coming in?

mnaser (changes requested):
thanks for your first contribution!

roles/kubernetes/tasks/main.yml
fitbeard: So I carefully tested again in isolated environment:
```
root@pxe:~# cat /etc/fstab
# /etc/fstab: static file system information.
#
# Use 'blkid' to print the universally unique identifier for a
# device; this may be used with UUID= as a more robust way to name devices
# that works even if disks are added and removed. See fstab(5).
#
# <file system> <mount point>   <type>  <options>       <dump>  <pass>
# / was on /dev/sda3 during curtin installation
/dev/disk/by-uuid/aeaf00c0-5326-498d-b85a-617be767807e / ext4 defaults 0 1
# /boot/efi was on /dev/sda1 during curtin installation
/dev/disk/by-uuid/B976-C874 /boot/efi vfat defaults 0 1
root@pxe:~# free -m
              total        used        free      shared  buff/cache   available
Mem:           7926         281        7220           1         424        7403
Swap:          2047           0        2047
root@pxe:~# systemctl status dev-sda2.swap
● dev-sda2.swap - Swap Partition
     Loaded: loaded (/run/systemd/generator.late/dev-sda2.swap; generated)
     Active: active since Fri 2023-02-17 23:49:40 UTC; 5min ago
       What: /dev/sda2
       Docs: man:systemd-gpt-auto-generator(8)
      Tasks: 0 (limit: 9405)
     Memory: 536.0K
     CGroup: /system.slice/dev-sda2.swap

Feb 17 23:49:40 pxe systemd[1]: Activating swap Swap Partition...
Feb 17 23:49:40 pxe systemd[1]: Activated swap Swap Partition.
root@pxe:~# systemctl stop dev-sda2.swap
root@pxe:~# systemctl disable dev-sda2.swap
root@pxe:~# free -m
              total        used        free      shared  buff/cache   available
Mem:           7926         713        6508           2         704        6964
Swap:             0           0           0
```
if `systemd` finds a SWAP partition on any drive it will anyway automount it (Ubuntu). On cloud instances this is not the case but on real hardware this can happen.
I think creating this `noswap` service  will be better approach than masking swap service because this should work the same on all Linux distributions (thinking about Rocky).

fitbeard: @mnaser removing swap records from fstab is not enough:
https://unix.stackexchange.com/questions/671940/disabling-swap-on-debian-permanently
https://rageagainstshell.com/2019/12/disable-swap-in-systemd/
https://wiki.archlinux.org/title/Swap

mnaser (approved):

PR #332, opened 2023-02-22
Update percona backup image

mpiscaer:
With the 8.0 image you can create backups, but you can not restore the backups.

Atmosphere uses Percona 5.7. With a 8.0 backup you can not restore to a 5.7 Percona cluster.

### DIFF
--- a/roles/percona_xtradb_cluster/README.md
+++ b/roles/percona_xtradb_cluster/README.md
@@ -27,7 +27,7 @@ example below.
 ```yaml
 percona_xtradb_cluster_spec:
   backup:
-    image: perconalab/percona-xtradb-cluster-operator:main-pxc8.0-backup
+    image: percona/percona-xtradb-cluster-operator:1.12.0-pxc5.7-backup
     storages:
       fs-pvc:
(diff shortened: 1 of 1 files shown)


roles/percona_xtradb_cluster/README.md
```diff
@@ -27,7 +27,7 @@ example below.
 ```yaml
 percona_xtradb_cluster_spec:
   backup:
-    image: perconalab/percona-xtradb-cluster-operator:main-pxc8.0-backup
+    image: percona/percona-xtradb-cluster-operator:1.12.0-pxc5.7-backup
```
okozachenko1203, line 30: Can we go with the same version as the cluster operator?
```suggestion
    image: percona/percona-xtradb-cluster-operator:1.10.0-pxc5.7-backup
```

mpiscaer: I don't see a problem in that. So probably this is ok.

okozachenko1203: yeap, it should not include any breakage, just for version matching

Issue #184, opened 2022-11-16
Look into support RockyLinux as a base operating system


mnaser:

fitbeard:
@mnaser Why this issue was closed? Whats the story behind this? I'm asking because I want to start working on Rocky8/9 integration and to understand is it good time for this right now or this should wait because of some blocker. Thank you.

mnaser:
@fitbeard we just didn't have the internal priority to do that, there shouldn't be any blockers, you're free to push the changes to support that

mnaser:
for what it's worth, there has been progress in being able to deploy the k8s part of things entirely on top of other operating systems.  the next 'baremetal' step would be improving the ceph collection to handle EL based systems and I _think_ we can close this.

fitbeard:
@mnaser can we leave it open? I’m still not able to 100% deploy everything on EL without modifications. There is still few places with APT/Ununtu only logic. But you are right. The main blocker is Ceph and this is the hard one. Because having the same Ceph version from system packages on Ubuntu2004/2204 and/or EL8/EL9 is impossible. Pacific release is only available for focal/el8, quincy is only available for focal/el8/el9 but not for jammy and so on. Rook is the way.

mnaser:
@fitbeard Are you deploying targeting Rocky 8/9 by any chance at the moment?  I am asking because we can probably run a CI job targeting it if you are interested in helping maintain that upstream so we avoid breaking your stuff

fitbeard:
@mnaser from 2022 my goal is to make RockyLinux first class citizen in Atmosphere. Right now i'm running mixed lab based on RockyLinux running control-plane and pool of workers running Rocky/Jammy. Slowly i'm fixing small things in all ansible collections to make possible run whole setup without manual changes. I'm using this issue as place to mention PRs and changes to make this happen.

fitbeard:
> @fitbeard Are you deploying targeting Rocky 8/9 by any chance at the moment?  I am asking because we can probably run a CI job targeting it if you are interested in helping maintain that upstream so we avoid breaking your stuff

CI would be nice to have. This would help a lot in bug hunting. I suggest focus only on Rocky9 and skip 8 like we are now doing with Focal.

PR #333, opened 2023-02-22
feat: switch to binary runc and containerd install

fitbeard:
- add multipurpose `binary_download` role
- add`runc` role 
- rewrite `containerd` role and switch to binary installation
- add better and more detailed version control support for Atmosphere components

This also introduces first batch of changes required for multi Linux distribution support https://github.com/vexxhost/atmosphere/issues/184


mnaser:
@fitbeard this is really neat, I have a few questions if you don't mind me asking:

- What happens when you are running this on an existing deployment of Atmosphere?  When those binaries disappear or the services get related, do things go bork?
- Do you think it's possible for you to create a Molecule scenario which uses Docker that runs for both CentOS/RHEL & Ubuntu only for testing the containerd/runc roles?  It feels like it would be good to have that (at least for the components that are OS-specific) and those can run quickly on a separate molecule scenario (similar to image_manifest)

Thoughts?

mnaser:
Also I'm curious if there is any reasoning behind why you're pulling into the Ansible controller and then uploading again, we've been more on the behaviour of pulling directly into the control plane nodes and avoiding things to be on the Ansible controller

fitbeard:
> Also I'm curious if there is any reasoning behind why you're pulling into the Ansible controller and then uploading again, we've been more on the behaviour of pulling directly into the control plane nodes and avoiding things to be on the Ansible controller

Fixed. Now working on molecule part.

fitbeard:
> @fitbeard this is really neat, I have a few questions if you don't mind me asking:
> 
>     * What happens when you are running this on an existing deployment of Atmosphere?  When those binaries disappear or the services get related, do things go bork?
> 
I tried the `containerd` installed with the package manager by replacing it with a binary and restarting the service. All the containers on the hosts seem to be operational. So I think the behavior is the same as if it was updated by the package manager.
>     * Do you think it's possible for you to create a Molecule scenario which uses Docker that runs for both CentOS/RHEL & Ubuntu only for testing the containerd/runc roles?  It feels like it would be good to have that (at least for the components that are OS-specific) and those can run quickly on a separate molecule scenario (similar to image_manifest)
> 
> 
> Thoughts?

@mnaser I added needed molecule tests. Also additional role for `crictl` <- later I plan to start refactoring `kubernetes`. Right now `crictl` role is only used for this molecule scenario.

mnaser:
i tested on my side, things seem fine, nothing happened running this on a prod cluster.

PR #338, opened 2023-02-24
fix: disallow embedded of horizon in iframe

mnaser:
OpenStack Helm enables the Horizon service to be embedded in an iframe
by default. This is a security risk and should be disabled.

This patch disables the embedding of Horizon in an iframe.


mnaser:
cc @bmoliff @runlevel-six

PR #362, opened 2023-03-17
feat: enable autoscaler for magnum

okozachenko1203:
To remove flux dependency from magnum cluster-api driver, we are gonna use helm to install cluster autoscaler and it requires helm cli.

### DIFF
--- a/internal/pkg/image_repositories/template/Dockerfile
+++ b/internal/pkg/image_repositories/template/Dockerfile
@@ -30,5 +30,6 @@ RUN <<EOF /bin/bash
   rm -rf /var/lib/apt/lists/*
 EOF
 {{- else if eq .Project "magnum" }}
+COPY --from=docker.io/alpine/helm:3.11.2 /usr/bin/helm /usr/local/bin/helm
 COPY --from=gcr.io/go-containerregistry/crane /ko-app/crane /usr/local/bin/crane
 {{- end }}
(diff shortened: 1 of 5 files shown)


internal/pkg/image_repositories/template/Dockerfile
```diff
@@ -31,4 +31,10 @@ RUN <<EOF /bin/bash
 EOF
 {{- else if eq .Project "magnum" }}
 COPY --from=gcr.io/go-containerregistry/crane /ko-app/crane /usr/local/bin/crane
+RUN <<EOF /bin/bash
+  curl -Lo /tmp/helm.tar.gz "https://get.helm.sh/helm-v3.10.3-linux-amd64.tar.gz"
+  tar -zxvf /tmp/helm.tar.gz -C /tmp
+  mv /tmp/linux-amd64/helm /usr/local/bin/helm
+  rm -rf /tmp/helm.tar.gz /tmp/linux-amd64/
+EOF
```
mnaser, line 39: Is there no official Helm image that we can just copy directly from like we do in Crane?

mnaser (changes requested):

internal/pkg/image_repositories/template/Dockerfile
okozachenko1203: unfortunately, i couldn't find it in dockerhub and quay.io.
Maybe we can use `hub.docker.com/r/alpine/helm` ?

mnaser: that one works, because it seems like it's just copying the binary:

https://github.com/alpine-docker/helm/blob/master/Dockerfile#L20-L23

PR #369, opened 2023-03-23
feat: Add manila role

okozachenko1203:


roles/manila/vars/main.yml
```diff
@@ -0,0 +1,41 @@
...
+      api: 3
+      data: 3
+      scheduler: 3
+      share: 3
+  conf:
+    manila:
+      oslo_messaging_notifications:
+        driver: noop
+
+  dependencies:
```
mnaser, line 30: Shouldn’t this live in the Helm chart?

roles/defaults/defaults/main.yml
```diff
@@ -167,3 +167,8 @@ atmosphere_images:
   skopeo: quay.io/skopeo/stable:latest
   tempest_run_tests: us-docker.pkg.dev/vexxhost-infra/openstack/tempest:30.1.0-4
   vector: docker.io/timberio/vector:0.27.0-debian
+  manila_db_sync: quay.io/vexxhost/manila:zed
```
mnaser, line 170: Can you please put these in alphabetical order?

docs/developer/services.md
```diff
@@ -37,7 +38,6 @@ list of services that are planned to be included in the future.
 
 - [Magnum](https://docs.openstack.org/magnum/latest/)
```
mnaser, line 39: Maybe remove magnum here tok

mnaser (changes requested):
Small nuts.

roles/manila/vars/main.yml
okozachenko1203: well, i didn't make this as default in the chart and just commented with description.

roles/defaults/defaults/main.yml
okozachenko1203: ok

docs/developer/services.md
okozachenko1203: ok

roles/manila/vars/main.yml
okozachenko1203: because users can select other driver types or even they can use share service without driver

PR #378, opened 2023-04-04
fix: update RBAC for magnum cluster api

okozachenko1203:


mnaser:
I got done with this so I just gave clusteradmin for now...

PR #412, opened 2023-05-12
Add manila-service-security-group to limit manila ports

ricolin:


ricolin:
@mnaser validated

mnaser (approved):

mnaser (changes requested):
@ricolin -- i see a few problems here:

1. We modified the `charts` which are not meant to manually modified, if there are any changes, they should be done via `hack/sync-charts`
2. I prefer that we use the OpenStack Ansible modules instead, they seem to be much better at this, I would use this as an example: https://github.com/vexxhost/atmosphere/blob/main/roles/octavia/tasks/generate_resources.yml#L30-L48

The only thing is you'll have to make sure that you point to the correct project when using the Ansible modules.

ricolin:
will send another patch asap

mnaser (approved):

PR #415, opened 2023-05-18
chore: bump charts

ricolin:
* Update chart versions
* Apply patch openstack-helm/+/883168


mnaser:
@ricolin how did you bump the chart versions? we have a script that takes care of doing that to make sure all of our patches are applied, check out inside `hack`

okozachenko1203:
duplicated to https://github.com/vexxhost/atmosphere/pull/416

PR #421, opened 2023-05-29
chore(molecule): Update controller instance type

okozachenko1203:


mnaser (changes requested):
Can you please update the documentation in the README as well for this?

Also, to make things backwards compatible, can you make them default to ATMOSPHERE_INSTANCE_TYPE and then if that's not set it will set the overwritten values we setup?

mnaser (approved):

PR #422, opened 2023-05-30
fix(nova): Remove nova bootstrap job

okozachenko1203:


roles/nova/tasks/main.yml
```diff
@@ -40,6 +40,17 @@
   ansible.builtin.import_tasks:
     file: generate_public_key.yml
 
+# NOTE(okozachenko1203): We cannot set helm hook for this bootstrap job.
+# As a workaround, remove this job before upgrade nova helm release.
+- name: Remove nova-bootstrap job
+  failed_when: false
```
mnaser, line 46: I think we might need to add changed_when: false too so we can avoid the idempotence issues.

mnaser (changes requested):

Issue #428, opened 2023-05-31
Set `[barbican]/barbican_endpoint_type` to `internal` for Nova


mnaser:
In order to avoid going through the ingress, we should set `[barbican]/barbican_endpoint_type` to `internal` for Nova in order to force things to go over the internal service mesh.

PR #429, opened 2023-06-01
fix(nova): Set barbican_endpoint_type as internal

ricolin:
This avoid nova-barbican communication go cross ingress

fix https://github.com/vexxhost/atmosphere/issues/428


mnaser (changes requested):
i think we're trying to keep all the keys of `conf.nova` in alphabetical order (other than `DEFAULT`) so could you put `barbican` section right above `cache`?

mnaser (approved):

PR #433, opened 2023-06-01
fix(manila): update before install dependencies

ricolin:


ricolin:
This should fix manila image 1.4.0 build job https://github.com/vexxhost/atmosphere/actions/runs/5125491986/jobs/9266061946

mnaser (approved):

Issue #427, opened 2023-05-31
Allow for overriding image manifest


mnaser:
At the moment, we 'force' the user to supply all of the images in the `atmosphere_images` section, however, this is not that useful as it may be possible that a user wants to only override one or two images.

We should introduce `atmosphere_image_overrides` variable which will be combined into `atmosphere_images`.

PR #435, opened 2023-06-02
Add atmosphere_images_override options

ricolin:
This is for ability to provide override values for atmosphere_images


ricolin:
fix https://github.com/vexxhost/atmosphere/issues/427

mnaser (changes requested):
I think there's a cleaner way to do this.

I think you can create `roles/defaults/vars/main.yml` and move `atmosphere_images` from `roles/defaults/defaults/main.yml` to `_atmosphere_images` in `roles/defaults/vars/main.yml`.

Then, you can create `atmosphere_image_overrides: {}` in `defaults` as you already have, and then inside `vars`, you would do: `atmosphere_images: "{{ _atmosphere_images | combine(atmosphere_image_overrides, recursive=True) }}"`

mnaser:
We need to update these two as well:

https://github.com/vexxhost/atmosphere/blob/main/Earthfile#L63-L65
https://github.com/vexxhost/atmosphere/blob/main/build/pin-images.py

Just make sure those two are working too please! :)

mnaser (changes requested):
need some small cleanups.

PR #436, opened 2023-06-02
(WIP)feat: Build tempest image using atmosphere

okozachenko1203:


okozachenko1203:
@mnaser i think the git repository `docker-openstack-tempest` should be created manually at the initial step, right? I can see `atmosphere-ci` has `init` subcommand but that is not executed inside atmosphere gh workflows.
If i understood correctly, could you create `docker-openstack-tempest` repository?

mnaser:
already exists :)

Issue #399, opened 2023-05-02
Possible to create duplicate subnet names in public network


m49sdgn230n:
Using the member role, a user is able to create additional subnets in the shared public network. They are unable to create an overlapping subnet, but they can define a unique CIDR range and give it the same name. This is then shown across all domains when creating an instance or attaching an interface:

![image](https://user-images.githubusercontent.com/33265662/235770031-15f0283e-6431-4977-a95b-46765392f8e9.png)

If a subnet with a duplicate name is created, it will block future redeployments:

`failed: [controller] (item=[{'external': True, 'mtu_size': 1500, 'name': 'public', 'port_security_enabled': True, 'provider_network_type': 'flat', 'provider_physical_network': 'external', 'shared': True}, {'allocation_pool_start': 'x.x.x.x', 'allocation_pool_end': 'x.x.x.x', 'cidr': 'x.x.x.x/22', 'dns_nameservers': ['x.x.x.x', 'x.x.x.x'], 'enable_dhcp': True, 'gateway_ip': 'x.x.x.x', 'name': 'public-subnet'}]) => {"ansible_loop_var": "item", "attempts": 60, "changed": false, "extra_data": {"data": null, "details": "None", "response": "None"}, "item": [{"external": true, "mtu_size": 1500, "name": "public", "port_security_enabled": true, "provider_network_type": "flat", "provider_physical_network": "external", "shared": true}, {"allocation_pool_end": "x.x.x.x", "allocation_pool_start": "x.x.x.x", "cidr": "x.x.x.x/22", "dns_nameservers": ["x.x.x.x", "x.x.x.x"], "enable_dhcp": true, "gateway_ip": "x.x.x.x", "name": "public-subnet"}], "msg": "Multiple matches found for public-subnet"}`

okozachenko1203:
This issue is caused by the changes outside of the atmosphere.
In this case, we have 2 options
- add a custom policy to disallow the subnet creation with a specific name(the name used in atmosphere)
- Or use a special name which is not used for users normally
@mnaser  what do you think?

okozachenko1203:
> @okozachenko1203 i think the issue here is that it's the possiblity of creating a subnet on a network you don't own (in this a case, a member role can create the subnet on the public network which they dont own)...
> 
> maybe you can try and see if you can replicate it first @okozachenko1203 ?

@mnaser  yes, a member of admin project can create a subnet in the public network

okozachenko1203:
[...]

PR #472, opened 2023-07-03
Bump images for _atmosphere_images

ricolin:
This introduce fixes for public subnet.

fix https://github.com/vexxhost/atmosphere/issues/399

So after this fix, only member in admin project (where public network created) can add subnets to `public`


mnaser:
@ricolin could you remove the change to the magnum capi images?

PR #473, opened 2023-07-05
feat: add ironic

mnaser:
- chore: add initial ironic impl.
- chore: native ironic + neutron integration
- chore: enable ironic in nova
- fix(nova): removed bootstrap jobs
- fix(nova): drop un-unused api-paste for nova-compute-ironic
- fix(nova): stop using log_config_append


mnaser:
we should probaly move the ironic port into its own ns as it seems to conflict with the host networking, had to delete it since it conflicted with default networking.

Issue #471, opened 2023-07-03
Add secure RBAC for Magnum


ricolin:
Address the needs for secure RBAC for Magnum to let users aware the possible upcoming changes

https://review.opendev.org/c/openstack/magnum/+/875625/14
https://github.com/vexxhost/magnum/pull/10

ricolin:
We already have rbac released in vexxhost/magnum:zed and have the container tested.
The openstack community patch still requires some fixes before heat drivers are usable, but because we use vexxhost/magnum-cluster-api as driver, it will not be a problem.
it's added in PR: https://github.com/vexxhost/atmosphere/pull/487
What this means is, whoever not using magnum-cluster-api driver (aka not using atmosphere) will not be able to use that vexxhost/magnum:zed image.
For enable magnum rbac
simply add magnum config like follow:
```
conf:
     magnum:
       oslo_policy:
          enforce_new_defaults: True
          enforce_scope: True
```

ricolin:
https://github.com/vexxhost/magnum/pull/11 will fix the rest drivers which also have community WIP patch https://review.opendev.org/c/openstack/magnum/+/889144

ricolin:
secure RBAC is added as we bump image version in https://github.com/vexxhost/atmosphere/issues/471

ricolin:
trace https://review.opendev.org/c/openstack/magnum/+/893018

PR #487, opened 2023-07-14
Bump magnum image to support magnum rbac

ricolin:
This introduce magnum rbac to atmoshpere, to enable it add follow config to magnum:
```
conf:
  magnum:
    oslo_policy:
      enforce_new_defaults: True
      enforce_scope: True
````

[1] was added to vexxhost/magnum. But notice the magnum image only usable for vexxhost/magnum-cluster-api driver user.
fix https://github.com/vexxhost/atmosphere/issues/471

[1] https://review.opendev.org/c/openstack/magnum/+/875625/14


ricolin:
There is already something more image version advanced merged, that means this already merged and fixed.

PR #489, opened 2023-07-17
Bump neutron to 0.3.15

ricolin:
This sync asap2 patch to community work


ricolin:
fix https://github.com/vexxhost/atmosphere/pull/489

roles/neutron/vars/main.yml
```diff
@@ -62,6 +62,8 @@ _neutron_helm_values:
           tunnel_id_ranges: 1:1000
         ml2_type_vlan:
           network_vlan_ranges: external:1:4094
+    netoffload:
+      enabled: true
```
mnaser, line 66: I think it's totally ok to keep this at `false` since most folk will not be using hardware offload, so you can just remove this section.

mnaser (changes requested):

mnaser (approved):

Issue #490, opened 2023-07-18
Create an alert monitoring logs for list failures


mnaser:
We have seen a major issue with ports disappearing from Magnum clusters when the cell is not responding "for some reason".

We should use Loki's alert rules and issue an alert if we see something like this message:

```
2023-07-15 09:06:44.022 23 WARNING nova.compute.multi_cell_list [req-52e0b57e-7d9c-456c-98cd-69755cc2b15e 1b23ea79cd454800094ea06e0c319d2eb2585879745d43a2787a2f004bd86950 27f0c0244502435bbde078259cea6201 - b9b8e05b25c64b03abc10069c08b6217 b9b8e05b25c64b03abc10069c08b6217] Cell 1dd95335-7be1-47df-a076-a4fac1cba152 is not responding and hence is being omitted from the results
```

ref: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1612

PR #495, opened 2023-07-20
[ATMOSPHERE-103] chore: Add loki rule to delect Nova cell down

okozachenko1203:
fix https://github.com/vexxhost/atmosphere/issues/490

### DIFF
--- a/roles/loki/vars/main.yml
+++ b/roles/loki/vars/main.yml
@@ -80,20 +80,20 @@ _loki_helm_values:
   lokiCanary:
     enabled: false
   extraObjects:
-  - apiVersion: v1
-    kind: ConfigMap
-    metadata:
-      name: loki-alerting-rules
-      labels:
-        loki_rule: "atmosphere"
-    data:
-      loki-alerting-rules.yaml: |-
-        groups:
-          - name: additional-loki-rules
-            rules:
-              - alert: NovaCellNotResponding
-                expr: 'count_over_time({pod_label_component="compute"} |= "not responding and hence is being omitted from the results" [1m]) > 0'
-                labels:
-                  severity: critical
-                annotations:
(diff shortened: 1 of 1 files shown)


roles/loki/vars/main.yml
```diff
@@ -32,6 +32,16 @@ _loki_helm_values:
         installOperator: false
     lokiCanary:
       enabled: false
+    rules:
+      additionalGroups:
+        - name: additional-loki-rules
+          rules:
+            - alert: NovaCellNotResponding
+              expr: 'kubernetes.labels.component="compute" and log =~ "not responding and hence is being omitted from the results"'
```
okozachenko1203, line 40:
```suggestion
              expr: 'kubernetes.labels.component="compute" and log =~ /nova\.compute\.multi_cell_list.*Cell [0-9a-fA-F-]+ is not responding/'
```
I think we can use regex here but it will be more or less resource-intensive.

mnaser:
@okozachenko1203 I think the main issue here is that Loki is not integrated with the AlertManager deployed by the operator, so the alerts will be going to `/dev/null`.

Can you either make a new PR or update this one to send the alerts to AlertManager (as mentioned in https://grafana.com/docs/loki/latest/alert/#scheduling-and-best-practices)

Once we have that setup, then the alerts will actually "make sense" otherwise now they won't be going anywhere.

mnaser:
Looks like a merge conflict and needs some work to pass CI

mnaser:
recheck

okozachenko1203:
ok, i found that ruler sidecar is not deployed in singleBInary deployment mode.
https://github.com/grafana/loki/issues/10967

okozachenko1203:
https://github.com/grafana/loki/pull/13572

Issue #496, opened 2023-07-20
Keycloak support


mnaser:
One of the things that we need to accomplish is having proper SSO and authentication within OpenStack.  The plan is to enable Keycloak to be the primary source of truth for authentication for the cloud and all the different components, such as OpenStack dashboard, Grafana, etc.

There are a few components that we need to do in order to successfully complete this:

- [x] Deploy Keycloak (either using the operator or the Helm charts)
- [x] Setup an "Atmosphere" realm within Keycloak out of the box
- [x] Create clients for the different components (such as OpenStack dashboard, Grafana, etc)
- [x] Integrate all of the components with Keycloak for authentication
- [x] Add documentation to integrate external identity providers (such as Azure AD)

There are a few things that I think we need to be careful about:

We need to find a way to make sure that if a user gets deactivated/disabled, we do that in the other services.   We need to make sure if we disable a user inside of Keycloak, it should deactivate it in all the other systems such as Grafana & OpenStack.

We should also make sure Keycloak is opt-in only as well, because we don't want to break existing deployments, but I believe the long term plan is for things to MOVE towards native Keycloak auth only (probably at some point when we have a major release).

okozachenko1203:
- I agree, I bet once we use keycloak operator, such realm, clients, etc. can be managed by CRDs.
- This auth integration includes keystone OIDC with keycloak. i.e. not only GUIs such as grafana and horizon, all openstack API requests will be authed by keycloak using OIDC. I think it is feasible to deactivate already-issued oidc tokens as long as their lifespan is not expired.

mnaser:
I've been thinking about how we can get this done cleanly over the past few days and I think I have reached to a pretty good spot about how it can be handled:

- [ ] Build a new Keystone identity backend: `keycloak` 

    This should use the [Keycloak Admin REST API](https://www.keycloak.org/docs-api/18.0/rest-api/) in order to pull in all of the users/groups from Keycloak, the primary reasons for this would be for the purposes of disabling a user in Keycloak and it would automatically disable in the other OpenStack.  It would also mean that the second a user "exists" in Keycloak, it will automatically exist in OpenStack and we don't need wonky syncing.

- [ ] Deploy Keycloak using the operator inside Atmosphere

    With this, we'll be able to deploy the initial deployment of Keycloak and it will allow us to pave the path to be integrated, we should also probably create a realm called `atmosphere` to get all of this working properly.

- [ ] Integrate OpenStack with Keycloak

[...]

PR #510, opened 2023-08-01
feat: Add keycloak

okozachenko1203:
ref: https://github.com/vexxhost/atmosphere/issues/496


mnaser (changes requested):
@okozachenko1203 

FYI, two things:

- You can use Keycloak with MySQL just fine, we're actually running it in a few places already, so we can use the existing XtraDB cluster for this.
- Any reason why we're going for the Helm chart instead of the operator?  I think it might be better to go for the operator so we can create Realm and other things using CRs.

okozachenko1203:
> @okozachenko1203
> 
> FYI, two things:
> 
> * You can use Keycloak with MySQL just fine, we're actually running it in a few places already, so we can use the existing XtraDB cluster for this.
> * Any reason why we're going for the Helm chart instead of the operator?  I think it might be better to go for the operator so we can create Realm and other things using CRs.

The [new operator](https://github.com/keycloak/keycloak/tree/main/operator) is not perfect yet i think. It doesn't provide CRD for client. Ofc, we can define realm and clients using `KeycloakRealmImport` CRD or we can install [realm operator](https://github.com/keycloak/keycloak-realm-operator) as sidecar.

Main concern is there is no official helm chart for keycloak operator, just a few from some orgs but not popular. I am not sure how much we can benefit from operator to justify the effort to maintain the keycloak operator helm chart ourselves.

mnaser:
> > @okozachenko1203
> > FYI, two things:
> > 
> > * You can use Keycloak with MySQL just fine, we're actually running it in a few places already, so we can use the existing XtraDB cluster for this.
> > * Any reason why we're going for the Helm chart instead of the operator?  I think it might be better to go for the operator so we can create Realm and other things using CRs.
> 
> The [new operator](https://github.com/keycloak/keycloak/tree/main/operator) is not perfect yet i think. It doesn't provide CRD for client. Ofc, we can define realm and clients using `KeycloakRealmImport` CRD or we can install [realm operator](https://github.com/keycloak/keycloak-realm-operator) as sidecar.
> 
> Main concern is there is no official helm chart for keycloak operator, just a few from some orgs but not popular. I am not sure if we can benefit from operator much enough to justify the effort to maintain the keycloak operator helm chart ourselves.

You're right, it does not seem very pretty from looking at it.  I think we're actually better off using something like an [Ansible Module](https://docs.ansible.com/ansible/latest/collections/community/general/keycloak_client_module.html) to handle the creation of clients instead, that's far more simple and easy.

roles/keycloak/defaults/main.yml
```diff
@@ -0,0 +1,28 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+keycloak_helm_release_name: keycloak
+keycloak_helm_chart_path: "../../charts/keycloak/"
+keycloak_helm_chart_ref: /usr/local/src/keycloak
+
+keycloak_helm_release_namespace: openstack
```
mnaser, line 19: let's maybe use something like `auth-system` in here? since we use keycloak in several different components/places like grafana (monitoring ns), openstack (openstakc ns), etc

roles/keystone/meta/main.yml
```diff
@@ -32,3 +32,5 @@ dependencies:
     vars:
       upload_helm_chart_src: "{{ keystone_helm_chart_path }}"
       upload_helm_chart_dest: "{{ keystone_helm_chart_ref }}"
+  - role: keycloak
+    when: atmosphere_keycloak_enabled | default(true)
```
mnaser, line 36: since keycloak is an "independent" thing for many services, i think it should be added to the playbook directly rather than a dependency for keystone.

mnaser (changes requested):

roles/keycloak/vars/main.yml
```diff
@@ -0,0 +1,76 @@
...
+            "directAccessGrantsEnabled": true,
+            "protocol": "openid-connect",
+            "defaultClientScopes": [
+              "email",
+              "offline_access",
+              "profile",
+              "roles"
+            ]
+          }
+        ]
```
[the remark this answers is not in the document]
okozachenko1203: @mnaser what do you mean by using ansible module here?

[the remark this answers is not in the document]
okozachenko1203: nvm, i missed you above comments. I will check myself

roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,157 @@
...
+# under the License.
+
+- name: Create a Keycloak namespace
+  kubernetes.core.k8s:
+    state: present
+    definition:
+      apiVersion: v1
+      kind: Namespace
+      metadata:
+        name: "{{ keycloak_helm_release_namespace }}"
```
mnaser, line 22: we can remove this in favour of the `create_namespace` option in the helm module.

roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,157 @@
...
+  ansible.builtin.pip:
+    name: PyMySQL
+
+- name: Create Keycloak database
+  run_once: true
+  community.mysql.mysql_db:
+    login_host: "{{ _pxc_service.resources[0].spec.clusterIP }}"
+    login_user: root
+    login_password: "{{ openstack_helm_endpoints.oslo_db.auth.admin.password }}"
+    name: keycloak
```
mnaser, line 42: lets make this a configurable thing in defaults

roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,157 @@
...
+
+- name: Check if the Keycloak database secret exists
+  kubernetes.core.k8s_info:
+    api_version: v1
+    kind: Secret
+    name: keycloak-db-credentials
+    namespace: "{{ keycloak_helm_release_namespace }}"
+  register: _keycloak_db_secret
+  ignore_errors: true
```
mnaser, line 51: run_once ?

mnaser (changes requested):
partial review

roles/defaults/vars/main.yml
```diff
@@ -74,6 +74,8 @@ _atmosphere_images:
   ingress_nginx_default_backend: registry.k8s.io/defaultbackend-amd64:1.5
   ingress_nginx_kube_webhook_certgen: registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.1.1@sha256:64d8c73dca984af206adf9d6d7e46aa550362b1d7a01f3a0a91b20cc67868660 # noqa: yaml[line-length]
   keepalived: us-docker.pkg.dev/vexxhost-infra/openstack/keepalived:2.0.19
+  keycloak: docker.io/bitnami/keycloak:22.0.1-debian-11-r0
+  keycloak_config_cli: docker.io/bitnami/keycloak-config-cli:5.8.0-debian-11-r0
```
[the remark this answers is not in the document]
okozachenko1203: well, I tested a bit using the official image. But we rely on bitnami helm chart so it required almost all env var overrides.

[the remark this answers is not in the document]
okozachenko1203: i will look into this once other things all done

playbooks/generate_workspace.yml
```diff
@@ -379,6 +380,13 @@
             egrep '(_kek)' |
               cut -d':' -f2
 
+    - name: Generate global/common secrets
+      ansible.builtin.set_fact:
+        secrets: "{{ secrets | default({}) | combine({item: lookup('password', '/dev/null chars=ascii_lowercase,ascii_uppercase,digits length=32')}) }}"
+      when: item not in secrets
+      loop:
+        - atmosphere_grafana_oauth_client_secret
+
```
okozachenko1203, line 389: @mnaser do you agree on this way? There is no shared secret between roles still. For some shared/common vars, we used to set it as vars when call dependent roles in the role's meta so all are in one role's scope.
But i bet we will have this kind of common vars/secrets more in the future.

roles/rook_ceph_cluster/meta/main.yml
```diff
@@ -27,7 +27,8 @@ dependencies:
   - role: defaults
   - role: openstacksdk
   - role: openstack_helm_endpoints
-    openstack_helm_endpoints_list: ["rook_ceph_cluster"]
+    vars:
+      openstack_helm_endpoints_list: ["rook_ceph_cluster"]
```
okozachenko1203, line 31: this is just a drive-by change/fix out of the keycloak scope

roles/kube_prometheus_stack/vars/main.yml
```diff
@@ -31,6 +31,24 @@ _kube_prometheus_stack_helm_values:
...
+        client_secret: "{{ atmosphere_grafana_oauth_client_secret }}"
+        scopes: openid email profile offline_access roles
+        email_attribute_path: email
+        login_attribute_path: username
+        name_attribute_path: full_name
+        auth_url: "https://{{ keycloak_host }}/realms/{{ atmosphere_keycloak_realm | default('Atmosphere') }}/protocol/openid-connect/auth"
+        token_url: "https://{{ keycloak_host }}/realms/{{ atmosphere_keycloak_realm | default('Atmosphere') }}/protocol/openid-connect/token"
+        api_url: "https://{{ keycloak_host }}/realms/{{ atmosphere_keycloak_realm | default('Atmosphere') }}/protocol/openid-connect/userinfo"
+        tls_skip_verify_insecure: true
+        role_attribute_path: "contains(roles[*], 'admin') && 'Admin' || contains(roles[*], 'editor') && 'Editor' || 'Viewer'"
```
okozachenko1203, line 51: @mnaser this role mapping is comming from grafana official doc. I don't have any other idea on this but have a hunch that we have to follow the Openstack role naming convention like member or reader. How about your idea?

okozachenko1203 (commented):
@mnaser plz recheck these

hack/sync-charts.sh
```diff
@@ -238,3 +238,7 @@ curl 'https://review.opendev.org/changes/openstack%2Fopenstack-helm~883168/revis
   | filterdiff -p2 -x 'Chart.yaml' \
   | filterdiff -p1 -i 'manila/*' \
   | patch -p2 -d ${ATMOSPHERE}/charts/manila
+
+KEYCLOAK_VERSION=16.0.0
```
mnaser, line 242: its at 16.0.2 so lets bump that.

playbooks/generate_workspace.yml
mnaser: I go down about this later, I think it's easier/cleaner to refactor things out into a role.

roles/defaults/vars/main.yml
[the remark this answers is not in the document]
mnaser: Let's investigate this again, I would rather us try to make sure we use the upstream images.

roles/keycloak/defaults/main.yml
```diff
@@ -0,0 +1,31 @@
...
+
+keycloak_helm_release_namespace: auth-system
+keycloak_helm_values: {}
+
+keycloak_host: "{{ undef('You must specify a Keycloak host using keycloak_host') }}"
+keycloak_ingress_annotations: {}
+keycloak_ingress_class_name: "{{ atmosphere_ingress_class_name }}"
+
+keycloak_admin_username: admin
+keycloak_admin_password: "{{ undef(hint='You must specify a Keycloak admin password') }}"
```
mnaser, line 27: I've found that this is not very useful to troubleshoot to be honest, when this alert comes, it's hard to know the exact missing variable.  We have two alternatives, we either include the variable name in this alert, or we replace it with a simple comment that will be easy for the user to understand:

```suggestion
# keycloak_admin_password: 
```

roles/keycloak/vars/main.yml
```diff
@@ -0,0 +1,78 @@
...
+  auth:
+    adminPassword: "{{ keycloak_admin_password }}"
+    adminUser: "{{ keycloak_admin_username }}"
+  externalDatabase:
+    host: "{{ openstack_helm_endpoints.oslo_db.hosts.default }}.openstack"
+    port: 3306
+    database: "{{ keycloak_database_name }}"
+    existingSecret: "keycloak-db-credentials"
+    existingSecretUserKey: username
+    existingSecretPasswordKey: password
```
mnaser, line 35: we can replace this by using `externalDatabase.user` and `externalDatabase.password`

```suggestion
    user: "{{ keycloak_database_username }}"
    password: "{{ keycloak_database_password }}"
```

I know I suggested auto-generating earlier, but I think it would be better that we do not auto-generate this stuff and keep it for the user to create (the password at least).  That way, the user can rotate passwords if they want easily.

roles/keycloak/tasks/main.yml
mnaser: I think we're better off not auto-generating and asking the user to create it, so let's drop this now and drop the secret removal too.

roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,174 @@
...
+        namespace: "{{ keycloak_helm_release_namespace }}"
+      type: Opaque
+
+- name: Grab Keycloak database secret
+  kubernetes.core.k8s_info:
+    api_version: v1
+    kind: Secret
+    name: keycloak-db-credentials
+    namespace: "{{ keycloak_helm_release_namespace }}"
+  register: _keycloak_db_secret
```
mnaser, line 75: this can be also removed them

roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,174 @@
...
+
+- name: Create a Keycloak user
+  community.mysql.mysql_user:
+    login_host: "{{ _pxc_service.resources[0].spec.clusterIP }}"
+    login_user: root
+    login_password: "{{ openstack_helm_endpoints.oslo_db.auth.admin.password }}"
+    name: "{{ _keycloak_db_secret.resources[0]['data']['username'] | b64decode }}"
+    password: "{{ _keycloak_db_secret.resources[0]['data']['password'] | b64decode }}"
```
mnaser, line 83: This would be simply getting it from the defaults, where the username would default to `keycloak` and the password would be requested by the user.

roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,174 @@
...
+        standard_flow_enabled: "{{ item.1.standard_flow_enabled }}"
+        direct_access_grants_enabled: "{{ item.1.direct_access_grants_enabled }}"
+        protocol: "{{ item.1.protocol }}"
+      with_subelements:
+        - "{{ _keycloak_realms + keycloak_realms }}"
+        - clients
+      retries: 60
+      delay: 5
+      register: _result
+      until: _result is not failed
```
mnaser, line 174: I think it would be neat to move this to a role called `keycloak_client` similar to how we have `glance_image` .. in there, we would have defaults:

```yaml
keycloak_client_admin_username: "{{ keycloak_admin_username | default("admin") }}"
keycloak_client_admin_password: "{{ keycloak_admin_password }}"
```

by doing this, then we can just call that role whenever we need and dont need to have to put credentials into each service.

roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,174 @@
...
+        auth_username: "{{ keycloak_admin_username }}"
+        auth_password: "{{ keycloak_admin_password }}"
+        id: "{{ item.id }}"
+        realm: "{{ item.id }}"
+        enabled: true
+      loop: "{{ _keycloak_realms + keycloak_realms }}"
+      retries: 60
+      delay: 5
+      register: _result
+      until: _result is not failed
```
mnaser, line 143: i think it would be neat to actually create a role called `keycloak_realm` which would create a realm (similar to the way we `glance_image`) as well, same concept as `keycloak_client` described below..

We can then create another role called `keycloak_atmosphere_realm` which calls `keycloak_realm` with the values needed to create the atmosphere realm, and this would be a dependency for those services that need to use it.

roles/keycloak/vars/main.yml
```diff
@@ -0,0 +1,78 @@
...
+          - profile
+          - roles
+        redirect_uris:
+          - "https://{{ kube_prometheus_stack_grafana_host }}/*"
+        root_url: "https://{{ kube_prometheus_stack_grafana_host }}"
+        web_origins:
+          - "https://{{ kube_prometheus_stack_grafana_host }}"
+        standard_flow_enabled: true
+        direct_access_grants_enabled: true
+        protocol: openid-connect
```
mnaser, line 78: we can move this into the seperate roles as discussed

also, i think since the realm goes into the url/etc, it would be best if we lower case it.. so id `atmosphere`.

roles/kube_prometheus_stack/vars/main.yml
```diff
@@ -31,6 +31,24 @@ _kube_prometheus_stack_helm_values:
         openstack-control-plane: enabled
   grafana:
     adminPassword: "{{ kube_prometheus_stack_grafana_admin_password }}"
+    grafana.ini:
+      server:
+        root_url: https://{{ kube_prometheus_stack_grafana_host }}
+      auth.generic_oauth:
+        enabled: true
+        name: Keycloak-OAuth
```
mnaser, line 39: Let's switch this to "Atmosphere"

roles/kube_prometheus_stack/vars/main.yml
```diff
@@ -31,6 +31,24 @@ _kube_prometheus_stack_helm_values:
...
+        allow_sign_up: true
+        client_id: "{{ atmosphere_grafana_oauth_client_id | default('grafana') }}"
+        client_secret: "{{ atmosphere_grafana_oauth_client_secret }}"
+        scopes: openid email profile offline_access roles
+        email_attribute_path: email
+        login_attribute_path: username
+        name_attribute_path: full_name
+        auth_url: "https://{{ keycloak_host }}/realms/{{ atmosphere_keycloak_realm | default('Atmosphere') }}/protocol/openid-connect/auth"
+        token_url: "https://{{ keycloak_host }}/realms/{{ atmosphere_keycloak_realm | default('Atmosphere') }}/protocol/openid-connect/token"
+        api_url: "https://{{ keycloak_host }}/realms/{{ atmosphere_keycloak_realm | default('Atmosphere') }}/protocol/openid-connect/userinfo"
```
mnaser, line 49: If we are actually using the role, we can actually set a fact from creating the client and get that information directly back from it!

roles/kube_prometheus_stack/vars/main.yml
mnaser: I think it's just a suggestion, I think what we need to do is the following:

- Create *application* roles inside Keycloak (so not global roles, roles that are application specific)
- User can assign users/groups access to Grafana with the admin/member/reader roles
- Grafana will not give Viewer permission, but 0 permissions if they do not have any of those roles.  This will stop any random cloud users of being able to access it.

mnaser (changes requested):

roles/kube_prometheus_stack/vars/main.yml
okozachenko1203: the `keycloak_realm` module doesn't output openid config url https://docs.ansible.com/ansible/latest/collections/community/general/keycloak_realm_module.html#return-values.
Just we can use the realm id from the `proposed` output.

roles/kube_prometheus_stack/vars/main.yml
okozachenko1203: @mnaser  so it means we will create the roles like `grafana_admin`, `grafana_editor`, and `grafana_viewer` for Grafana, 
or
we will have `admin`, `member`, `reader` roles which present the perm levels and app-specific roles like `grafana`, etc.
Then we will check if the user has the set of roles like [grafana, admin], then grant grafana admin role, if the user has roles like `[grafana, viewer]` then grant viewer role. If he has role only `admin`,we will not grant grafana access.

Which one is perferrable?

Also are we gonna create those roles in the ansible playbook?

roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,79 @@
...
+- name: Create Keycloak database
+  run_once: true
+  community.mysql.mysql_db:
+    login_host: "{{ _pxc_service.resources[0].spec.clusterIP }}"
+    login_user: root
+    login_password: "{{ openstack_helm_endpoints.oslo_db.auth.admin.password }}"
+    name: "{{ keycloak_database_name }}"
+
+- name: Create a Keycloak user
+  community.mysql.mysql_user:
```
mnaser, line 36: `run_once`

roles/keycloak/vars/main.yml
```diff
@@ -0,0 +1,74 @@
...
+    registry: "{{ atmosphere_images['keycloak'] | vexxhost.kubernetes.docker_image('domain') }}"
+    repository: "{{ atmosphere_images['keycloak'] | vexxhost.kubernetes.docker_image('path') }}"
+    tag: "{{ atmosphere_images['keycloak'] | vexxhost.kubernetes.docker_image('tag') }}"
+  ingress:
+    enabled: true
+    ingressClassName: "{{ keycloak_ingress_class_name }}"
+    annotations:
+      cert-manager.io/cluster-issuer: atmosphere
+    hostname: "{{ keycloak_host }}"
+    tls: true
```
mnaser, line 63: my concern with using their built in ingress thing, is that users cannot override this with common overrides

we are using `openstack_helm_ingress` and users are kinda using the overrides there, .. maybe we can create role `ingress` and then `openstack_helm_ingress` uses it and then we can make this role us it as well?

roles/keycloak_client/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+    base_url: "{{ keycloak_client_base_url }}"
+    enabled: true
+    client_authenticator_type: client-secret
+    public_client: false
+    secret: "{{ keycloak_client_secret }}"
+    redirect_uris: "{{ keycloak_client_redirect_uris }}"
+    web_origins: "{{ keycloak_client_web_origins }}"
+    standard_flow_enabled: "{{ keycloak_client_standard_flow_enabled }}"
+    direct_access_grants_enabled: "{{ keycloak_client_direct_access_grants_enabled }}"
+    protocol: "{{ keycloak_client_protocol }}"
```
mnaser, line 37: for the non-required items, i think we should have `| default(omit)`

roles/kube_prometheus_stack/meta/main.yml
```diff
@@ -29,3 +29,17 @@ dependencies:
     vars:
       upload_helm_chart_src: "{{ kube_prometheus_stack_helm_chart_path }}"
       upload_helm_chart_dest: "{{ kube_prometheus_stack_helm_chart_ref }}"
+  - role: keycloak_atmosphere_realm
```
mnaser, line 32: so we can technically remove this cause we can make `keycloak_client` depeend on `keycloak_realm` pointing to `keycloak_client_realm_id`

mnaser (changes requested):
almost there

roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,79 @@
...
+    values: "{{ _keycloak_helm_values | combine(keycloak_helm_values, recursive=True) }}"
+
+- name: Disable pxc strict mode
+  community.mysql.mysql_query:
+    login_host: "{{ _pxc_service.resources[0].spec.clusterIP }}"
+    login_user: root
+    login_password: "{{ openstack_helm_endpoints.oslo_db.auth.admin.password }}"
+    query: "set global pxc_strict_mode='ENFORCING'"
+
+- name: Wait until Keycloak ready
```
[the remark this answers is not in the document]
okozachenko1203: yes, there is.
Once we move keycloak health check into `keycloak_client` and `keycloak_realm` role, we cannot rely on the ready status of K8s resource(sts here). We should check if keycloak server is functioning using curl to `keycloak_host` endpoint.
ofc, we can check keycloak health check endpoints https://www.keycloak.org/server/health  in those roles in advance. But health check endpoints are disabled in default so it should be enabled explicitly.

I thought it is more reasonable to confirm keycloak is ready in keycloak role and in those client and realm roles, it should be fail if the provided keycloak endpoint is not responding or unhealthy instead of waiting.
One more thing is keycloak server bootstrapping takes a bit long time during db migration task.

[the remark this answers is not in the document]
mnaser: @okozachenko1203  but since we have retries in the task, wouldn't the retry just eventually work?

roles/keycloak/vars/main.yml
okozachenko1203: ok.

But i think managing ingresses out of the helm chart has a downside.
In osh helm charts, resource names are static, not generated from the release name, you know. But in other normal helm charts, resource names are generated by their own template and includes helm release name.
It means we have to check how names are generated in the helm chart manifest and follow the same template to get the service name required for ingress spec.
Also when we upgrade the helm chart version, if there is any changes in the template or ingress version or sth related to ingress or service or annotation, we have to mirror them by using our own ingress role.

Btw, I created a new role for ingress and used it. Not a swiss-army knife for now. I think we can improve to support more complex specs like extraHosts, extraPaths, etc. in the future when we need it.

okozachenko1203:
i filed a new issue for this ovn ci failure not related to keycloak stuff. https://github.com/vexxhost/atmosphere/issues/527
So lets ignore ovn ci failure here.

mnaser:
ovn failure is unrelated, we should get around fixing it.

PR #512, opened 2023-08-02
feat: more EL compatibility

fitbeard:
New batch of changes for: https://github.com/vexxhost/atmosphere/issues/184

### DIFF
--- a/playbooks/kubernetes.yml
+++ b/playbooks/kubernetes.yml
@@ -22,6 +22,7 @@
     - role: ethtool
       tags:
         - ethtool
+      # (fitbeard) networkd-dispatcher is not available for EL distros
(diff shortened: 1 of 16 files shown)


mnaser:
@fitbeard those failures are interesting, but i think part of what's frustrating a little bit here is the weird openssl issues, you might need to look into those!

i wonder if they are updating certain system dependencies that are breaking things.. what do you think perhaps about building a venv and adding it to the PATH of the system to avoid messing with the system?

fitbeard:
It’s not weird at all. Github is full of gits with this topic: https://gist.github.com/kekru/deabd57f0605ed95d5c8246d18483687 - different distros different patches.

fitbeard:
Good point about dedicated venv. Will do.

Issue #529, opened 2023-08-17
CVE-2023-2088: Use service tokens


okozachenko1203:
https://bugs.launchpad.net/nova/+bug/2004555 (https://bugs.launchpad.net/nova/+bug/2004555/comments/57, https://bugs.launchpad.net/nova/+bug/2004555/comments/63)
All upstream codebases introduce the logic which require service tokens for the communication between openstack components.
Here is how to configure service tokens. https://docs.openstack.org/cinder/latest/configuration/block-storage/service-token.html

Fortunately, osh upstream already introduced service tokens https://opendev.org/openstack/openstack-helm/commit/91c8a5baf2cf2f0dddded57d88f00ea11dd4ff4a

Affected projects: Cinder, neutron, nova, glance, placement

What we have to do is
- [ ] bump the chart versions.
- [x] Sync forked codebases with upstream (we use forked repo for cinder only)
- [x] ~rebuild container images with the latest commits for all releases~ (we will not rebuild images for now. Just update charts to support service tokens, so we make sure new images will not break when they are released and used)

okozachenko1203:
@mnaser could you sync this forked repo? https://github.com/vexxhost/cinder

mnaser:
> @mnaser could you sync this forked repo? https://github.com/vexxhost/cinder

I'm trying to check why we had a fork there in the first place.. checking...

mnaser:
It was primarily because of https://review.opendev.org/c/openstack/cinder/+/876226 -- that backport was merged a while back.

@okozachenko1203 we can flip to upstream repository instead :)

okozachenko1203:
> It was primarily because of https://review.opendev.org/c/openstack/cinder/+/876226 -- that backport was merged a while back.
> 
> @okozachenko1203 we can flip to upstream repository instead :)

https://github.com/vexxhost/atmosphere/pull/535

PR #530, opened 2023-08-17
fix: CVE-2023-2088: Use service tokens

okozachenko1203:
fix https://github.com/vexxhost/atmosphere/issues/529

### DIFF
--- a/charts/neutron/templates/bin/_neutron-netns-cleanup-cron.py.tpl
+++ b/charts/neutron/templates/bin/_neutron-netns-cleanup-cron.py.tpl
@@ -1,0 +1,157 @@
+#!/usr/bin/env python
+
+import sys
+import os
+import time
+import socket
+from neutron.common import config
+from oslo_config import cfg
+from oslo_concurrency import processutils
+from neutron.agent.linux import dhcp
+from neutron.agent.l3 import namespaces
+from neutron.agent.l3 import dvr_snat_ns
+from neutron.agent.l3 import dvr_fip_ns
+from neutron.cmd.netns_cleanup import setup_conf
+from neutron.cmd.netns_cleanup import unplug_device
+from neutron.cmd.netns_cleanup import eligible_for_deletion
+from neutron.conf.agent import common as agent_config
+from neutron.agent.linux import ip_lib
+from keystoneauth1.identity import v3
+from keystoneauth1 import session
+from neutronclient.neutron import client as neutron_client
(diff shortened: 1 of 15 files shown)


charts/octavia/Chart.yaml
```diff
@@ -9,4 +9,4 @@ name: octavia
 sources:
 - https://opendev.org/openstack/octavia
 - https://opendev.org/openstack/openstack-helm
-version: 0.2.9
+version: 0.2.8
```
ricolin, line 12: we need to bump this to 0.2.9

charts/libvirt/templates/bin/_libvirt.sh.tpl
```diff
@@ -16,32 +16,6 @@ limitations under the License.
 
 set -ex
 
-# NOTE(mnaser): This will move the API certificates into the expected location.
```
ricolin, line 19: are we meant to delete this bdde64dad769ae05302b241c5c0db3b98c5f8272?

ricolin (commented):
I think we touched some codes that was from Atmoshphere side

charts/octavia/Chart.yaml
okozachenko1203: ok, i can see that patch merged. Let me remove the patch also

okozachenko1203:
@mnaser i can see that your patch to improve ovn support merged to upstream.
But there are some differences in neutron chart manifests between atmosphere and upstream. Could you review all good?

mnaser:
@okozachenko1203  we need to clean up the code that is cherry picking things...

okozachenko1203:
> @okozachenko1203 we need to clean up the code that is cherry picking things...

@mnaser i checked again and all patches are ok. Only missing part is https://github.com/vexxhost/atmosphere/commit/43677f6f3b8745179039ad0dee3f684c8413e9c1. This was not reflected to upstream. Could you make a patch of this commit or i do?

charts/neutron/templates/bin/_neutron-netns-cleanup-cron.py.tpl
okozachenko1203, line 1: this file is not used anymore in the neutron chart. Our patch commands doesn't apply `deleted files`  from patchsets and that is why this file is created whenever we run `sync-charts.sh` despite it is never used in neutron chart.
i will leave it as is. It will be removed once the upstream patch is merged.

PR #537, opened 2023-08-21
fix: Set cinder auth in nova

okozachenko1203:
drive-by: bump nova chart version and apply patches instead of local changes


mnaser (changes requested):
@okozachenko1203 i think this wont work, it will break vnc auth because i think we dont have the issuer wired in the libvirt side of things?  can you verify this in a lab env?   since im not sure where we are writing to use `libvirt-vnc` as the issuer and we need to make sure that issuer is used.

okozachenko1203:
> @okozachenko1203 i think this wont work, it will break vnc auth because i think we dont have the issuer wired in the libvirt side of things? can you verify this in a lab env? since im not sure where we are writing to use `libvirt-vnc` as the issuer and we need to make sure that issuer is used.

ok, i can see that one patch was merged to upstream for vencrypt support in libvirt, but includes some diff from ours. https://opendev.org/openstack/openstack-helm-infra/commit/f234218dba3cfad36513e7701cabf18a7fde217a#diff-1dfa1babae53c0b9cbe57e6a10f42919786c0a0f

Let me sync with that

okozachenko1203:
Lab Test result
- Tempest smoke success with OVN network backend
```sh

        "======",
        "Totals",
        "======",
        "Ran: 132 tests in 394.6475 sec.",
        " - Passed: 127",
        " - Skipped: 5",
        " - Expected Fail: 0",
        " - Unexpected Success: 0",
        " - Failed: 0",
        "Sum of execute time for each test: 467.9648 sec.",
        "",
        "==============",
        "Worker Balance",
        "==============",
        " - Worker 0 (30 tests) => 0:06:34.647495",
        " - Worker 1 (33 tests) => 0:04:48.689663",
        " - Worker 2 (33 tests) => 0:03:57.783281",
        " - Worker 3 (36 tests) => 0:05:50.116100",
        ""

```
- vnc vencrypt auth scheme works as expected
```sh
2023-09-05 18:13:28.534 136 INFO nova.console.websocketproxy [None req-5fe9bf9f-e91f-4b84-8d07-d00aac8b38a2 - - - - - -]  116: connecting to: 10.96.240.29:5900
<}<}2023-09-05 18:13:29.084 136 INFO nova.console.rfb.authvencrypt [None req-5fe9bf9f-e91f-4b84-8d07-d00aac8b38a2 - - - - - -] VeNCrypt security handshake accepted
2023-09-05 18:13:29.084 136 INFO nova.console.securityproxy.rfb [None req-5fe9bf9f-e91f-4b84-8d07-d00aac8b38a2 - - - - - -] Finished security handshake, resuming normal proxy mode using secured socket
```

okozachenko1203:
> @okozachenko1203 i think this wont work, it will break vnc auth because i think we dont have the issuer wired in the libvirt side of things? can you verify this in a lab env? since im not sure where we are writing to use `libvirt-vnc` as the issuer and we need to make sure that issuer is used.

done

PR #538, opened 2023-08-21
chore(keystone): Sync keystone image sha

ricolin:
Keystone is failing because we remove the old build keystone image (sha 8214ccc92d7973451509701c5728d5af614762be3b2b89b29e486f28eb68167f) 
Which block jobs for PRs (like https://github.com/vexxhost/atmosphere/actions/runs/5891949296/job/15980160551?pr=530#step:6:169)

PR #540, opened 2023-08-21
feat: use env variable for custom prepare playbook.

mpiscaer:
Import a custom playbook in the molecule prepare playbook. This to  prepare the instances.


okozachenko1203:
https://github.com/vexxhost/atmosphere/pull/538 will fix CI failures

Issue #542, opened 2023-08-21
bug: Add information of image build project itself in the container image tag


okozachenko1203:
### Descriptioin
Now container image tags only include target project's release and commit hash value but no infor related to build project's.
https://github.com/vexxhost/atmosphere/blob/7b46b46bac204d0f9029d42513a61c87da0a768a/internal/pkg/image_repositories/build_workflow.go#L185-L191
We faced image overriding (i.e specific image digests have been deleted which are consumed by atmosphere head branch) when only build project manifests updated. (for example, https://github.com/vexxhost/docker-openstack-keystone/commit/c427be0b01490c44df079a833ca345a08c3a0725, https://github.com/vexxhost/docker-openstack-keystone/commit/df83bc9f50198d5431872f274d70aadaeb21f53c. In these 2 commits, CI built images using a same tag).
As a result, we faced CI failures in atmosphere because of non-existing image digest. https://github.com/vexxhost/atmosphere/actions/runs/5926478084/job/16069881013?pr=537

### Solution
Add some information related to image build action (for example, image build time or image repository commit hash value) in the image tag.

mnaser:
This is now done natively via Zuul.

PR #543, opened 2023-08-21
Add build project's ref in container image tag

okozachenko1203:
fix https://github.com/vexxhost/atmosphere/issues/542

This PR could be considered as just a reference because we are gonna offloading the logic for building image build repository from this atmosphere project. Then just maintain Dockerfiles inside the image build repository itself.
In this case, we can replicate the same changes in the gh workflow manifest of build repositories manually.

### DIFF
--- a/internal/pkg/image_repositories/build_workflow.go
+++ b/internal/pkg/image_repositories/build_workflow.go
@@ -156,7 +156,7 @@ func NewBuildWorkflow(project string) *GithubWorkflow {
 					},
 					{
 						Name: "Setup environment variables",
-						Run:  "echo PROJECT_REF=$(cat manifest.yml | yq '.\"${{ matrix.release }}\".sha') >> $GITHUB_ENV",
+						Run:  "echo PROJECT_REF=$(cat manifest.yml | yq '.\"${{ matrix.release }}\".sha') >> $GITHUB_ENV\necho BUILD_PROJECT_REF=$(git rev-parse --short HEAD) >> $GITHUB_ENV",
 					},
 					{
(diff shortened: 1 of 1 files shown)


internal/pkg/image_repositories/build_workflow.go
```diff
@@ -156,7 +156,7 @@ func NewBuildWorkflow(project string) *GithubWorkflow {
 					},
 					{
 						Name: "Setup environment variables",
-						Run:  "echo PROJECT_REF=$(cat manifest.yml | yq '.\"${{ matrix.release }}\".sha') >> $GITHUB_ENV",
+						Run:  "echo PROJECT_REF=$(cat manifest.yml | yq '.\"${{ matrix.release }}\".sha') >> $GITHUB_ENV\necho BUILD_PROJECT_REF=$(git rev-parse --short HEAD) >> $GITHUB_ENV",
```
ricolin, line 159: I know in direct command `\n` here will just make things echo to file `$GITHUB_ENVecho`, maybe gitaction perform differently?
If not, we can sue && instead. :)

okozachenko1203: I think it will output like
```yaml
run: |
  echo PROJECT_REF=$(cat manifest.yml | yq '.\"${{ matrix.release }}\".sha') >> $GITHUB_ENV
  echo BUILD_PROJECT_REF=$(git rev-parse --short HEAD) >> $GITHUB_ENV
```

mnaser:
These labels are now coming naturally via Zuul here.

https://github.com/vexxhost/atmosphere/pull/1090

Issue #518, opened 2023-08-08
Do not delete `pvc` inside Glance


mnaser:
If a user accidentally change the storage mode from pvc to another mode, the PVC is deleted since it's not in the new list of manifests.

This is not really good experience for users, so it would be good to eliminate this and keep the pvc to survive no matter what.

ricolin:
resolved, but let's keep tracking on community patch  https://review.opendev.org/c/openstack/openstack-helm/+/892704

PR #549, opened 2023-08-24
Use keep as resoruce policy for glance-images

ricolin:
Make sure Once PVC glance-images never get deleted.
fix https://github.com/vexxhost/atmosphere/issues/518

### DIFF
--- a/charts/glance/templates/pvc-images.yaml
+++ b/charts/glance/templates/pvc-images.yaml
@@ -20,6 +20,8 @@ kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
   name: glance-images
+  annotations:
+    helm.sh/resource-policy: keep
 spec:
   accessModes: [ "ReadWriteOnce" ]
   resources:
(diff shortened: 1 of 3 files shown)


ricolin:
propose a patch to openstack helm, https://review.opendev.org/c/openstack/openstack-helm/+/892704
let's see how's the discussion there

okozachenko1203:
> propose a patch to openstack helm, https://review.opendev.org/c/openstack/openstack-helm/+/892704 let's see how's the discussion there

@ricolin i think this is very authentic and also very clear. But it is always to good to listen from community so I suggest that we can merge this and just keep eye on the upstream patch.

okozachenko1203 (approved):

mnaser:
I disagree with upstream.  99.99999% of the time the only reason your get a PVC deleted is because of a mistake.

ricolin:
@mnaser FYI, I update the upstream to direct the flag to default true. will update here once it's merged

Issue #550, opened 2023-08-24
Bump libvirt to Zed


mnaser:
This should be a trivial change, but we should bump and pin to the Zed image. :)

https://github.com/vexxhost/atmosphere/blob/17b6411a0b2ed58228e6bae099abb0977efed8b6/roles/defaults/vars/main.yml#L96

PR #554, opened 2023-08-30
feat: introduce aarch64 support for libvirt pod

fitbeard:
Closes https://github.com/vexxhost/atmosphere/issues/550

- Starting from this tag `libvirtd` build for `aarch64` was introduced.
- `bitnami/kubectl` is using different SHA per platform. Switch to TAG instead of SHA (`1.27.3` is the version which `atmoshere` is using now).

### DIFF
--- a/roles/defaults/vars/main.yml
+++ b/roles/defaults/vars/main.yml
@@ -17,7 +17,7 @@ _atmosphere_images:
   barbican_api: quay.io/vexxhost/barbican@sha256:fde302ee731cca6019feaf87400f5a377c3e38f459bc88d4c7677f2967e0939b # image-source: quay.io/vexxhost/barbican:zed
   barbican_db_sync: quay.io/vexxhost/barbican@sha256:fde302ee731cca6019feaf87400f5a377c3e38f459bc88d4c7677f2967e0939b # image-source: quay.io/vexxhost/barbican:zed
   bootstrap: quay.io/vexxhost/heat@sha256:755225f9a63c0968f1ceeda3a2f06c66dd8d247ff00308f549e66496aa8f59d0 # image-source: quay.io/vexxhost/heat:zed
-  ceph_config_helper: quay.io/vexxhost/libvirtd@sha256:d400204e0332dc815827e5902038a1c672446c58633ba97ede9e20f8ae9a2349 # image-source: quay.io/vexxhost/libvirtd:yoga-focal
+  ceph_config_helper: quay.io/vexxhost/libvirtd@sha256:4c6ce66f2b957eaf18d35c8236a038bd2289c78eec7c21addfa357b30f97b4ed # image-source: quay.io/vexxhost/libvirtd:zed-jammy
   ceph: quay.io/ceph/ceph:v16.2.11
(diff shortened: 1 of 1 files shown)


mnaser (dismissed):

mnaser:
@fitbeard ok i think we can run the pin images again to bump to the latest image which has the fixes to support `aarch64`

roles/defaults/vars/main.yml
```diff
@@ -92,8 +92,8 @@ _atmosphere_images:
   kube_proxy: registry.k8s.io/kube-proxy:v1.22.17
   kube_scheduler: registry.k8s.io/kube-scheduler:v1.22.17
   kube_state_metrics: registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.9.2
-  kubectl: docker.io/bitnami/kubectl@sha256:bd420268ae3424b3ab3174e26b895fd8dc464589a8cd62654b9aa739d00ff280 # image-source: docker.io/bitnami/kubectl:latest
-  libvirt: quay.io/vexxhost/libvirtd@sha256:d400204e0332dc815827e5902038a1c672446c58633ba97ede9e20f8ae9a2349 # image-source: quay.io/vexxhost/libvirtd:yoga-focal
+  kubectl: docker.io/bitnami/kubectl:1.27.3
+  libvirt: quay.io/vexxhost/libvirtd@sha256:4c6ce66f2b957eaf18d35c8236a038bd2289c78eec7c21addfa357b30f97b4ed # image-source: quay.io/vexxhost/libvirtd:zed-jammy
```
ricolin: as we don't build focal for zed anymore, I think we are save to directly using zed instead of `zed-jammy` :)

roles/defaults/vars/main.yml
```diff
@@ -17,7 +17,7 @@ _atmosphere_images:
   barbican_api: quay.io/vexxhost/barbican@sha256:fde302ee731cca6019feaf87400f5a377c3e38f459bc88d4c7677f2967e0939b # image-source: quay.io/vexxhost/barbican:zed
   barbican_db_sync: quay.io/vexxhost/barbican@sha256:fde302ee731cca6019feaf87400f5a377c3e38f459bc88d4c7677f2967e0939b # image-source: quay.io/vexxhost/barbican:zed
   bootstrap: quay.io/vexxhost/heat@sha256:755225f9a63c0968f1ceeda3a2f06c66dd8d247ff00308f549e66496aa8f59d0 # image-source: quay.io/vexxhost/heat:zed
-  ceph_config_helper: quay.io/vexxhost/libvirtd@sha256:d400204e0332dc815827e5902038a1c672446c58633ba97ede9e20f8ae9a2349 # image-source: quay.io/vexxhost/libvirtd:yoga-focal
+  ceph_config_helper: quay.io/vexxhost/libvirtd@sha256:4c6ce66f2b957eaf18d35c8236a038bd2289c78eec7c21addfa357b30f97b4ed # image-source: quay.io/vexxhost/libvirtd:zed-jammy
```
ricolin: same here, we can use `zed` directly

ricolin (commented):
Lgtm, just a nit worth fix :)

PR #570, opened 2023-09-07
feat: add cluster_issuer_type none

mpiscaer:
Add cluster_issuer_type none for staticly certs


roles/cluster_issuer/tasks/type/none/main.yml
```diff
@@ -0,0 +1 @@
+---
```
mnaser, line 1: can you please add the usual apache2 license header

roles/openstack_helm_ingress/README.md
```diff
@@ -24,6 +24,8 @@ with Atmosphere, you can simply configure it as follows;
 
    ```yaml
    openstack_helm_ingress_secret_name: wildcard-certs
+   cluster_issuer_type: none
```
mnaser, line 27: can you move this to the line above and move the trailing line?

mnaser (changes requested):

roles/cluster_issuer/tasks/type/none/main.yml
mpiscaer: fixed

mnaser (approved):

PR #573, opened 2023-09-08
feat: Enable libvirt exporter ootb

okozachenko1203:


roles/libvirt/tasks/main.yml
```diff
@@ -87,3 +87,27 @@
...
+          name: libvirt-exporter
+          namespace: "{{ libvirt_helm_release_namespace }}"
+        spec:
+          podMetricsEndpoints:
+            - path: /metrics
+              port: metrics
+          selector:
+            matchLabels:
+              application: libvirt
+              component: libvirt
```
mnaser, line 113: we should do this inside the `additionalPodMonitors` inside `kube-prometheu-stack`.

mnaser (changes requested):

PR #578, opened 2023-09-12
fix: Upgrade CAPO version

okozachenko1203:
Fixes https://github.com/vexxhost/magnum-cluster-api/issues/199

ref: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1653


mnaser:
@okozachenko1203 : I'm just testing if the newer upstream works: https://github.com/vexxhost/magnum-cluster-api/pull/212 -- if it does, then lets just jump straight and hit 'two birds with one stone' :)

mnaser:
replaced by #594

PR #579, opened 2023-09-12
WIP: Offline install

mpiscaer:


mpiscaer:
@mnaser @guilhermesteinmuller I'm not sure in the versions I use in [Create PIP mirror](https://github.com/mpiscaer/atmosphere/blob/offline_install/doc/source/deploy/airgapped.rst#create-pip-mirror) and or in [Using Earthly to build the images](https://github.com/mpiscaer/atmosphere/blob/offline_install/doc/source/deploy/airgapped.rst#using-earthly-to-build-the-images). Would be great if one of you can feedback on this.

PR #582, opened 2023-09-13
Switch to pxc_strict_mode to MASTER

ricolin:
This will give us sql lock support so we can use tooz with mysql backend.

Also fix task name in keycloak role

ref: https://docs.percona.com/percona-xtradb-cluster/8.0/strict-mode.html

### DIFF
--- a/roles/keycloak/tasks/main.yml
+++ b/roles/keycloak/tasks/main.yml
@@ -75,9 +75,9 @@
     ingress_annotations:
       cert-manager.io/cluster-issuer: atmosphere
 
-- name: Disable pxc strict mode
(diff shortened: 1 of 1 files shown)


mnaser:
I'm not sure how I feel about this change and the possible implications behind it.

Is there any other drivers that Tooz supports that we could leverage instead, @ricolin ?

ricolin:
I think it would be ZooKeeper or ETCD(etcd/etcd3/etcdgw)

ricolin:
here's one more possible solution, to directly use filesystem as we can share base on mount with kubernetes. make sure we mount same pvc for all nodes should make the work happen. I'm looking into it now

PR #587, opened 2023-09-15
fix: mcapi offline installs

mnaser:
- chore: bump to version of v.k8s with offline installs
- chore: refactor to v.k8s.cert_mgr


mnaser:
I've taken a subset of the changes and moved them here:

https://github.com/vexxhost/atmosphere/pull/590

Issue #586, opened 2023-09-15
staffeln support in Atmosphere


ricolin:
This track staffeln support in Atmoshpere
Few tasks to be done:
- [x] Staffeln image publish (we still have Unauthorized when publish staffeln image on quary.io after built one) (see https://github.com/vexxhost/staffeln/actions/runs/6181208321/job/16778839169#step:7:775 )
- [x] Staffeln charts https://github.com/vexxhost/staffeln/issues/107 (This can be consider as a follow up action)
- [x] https://github.com/vexxhost/atmosphere/issues/593 Staffeln roles and add chart for Vexxhost/staffeln ( https://github.com/vexxhost/atmosphere/compare/main...staffeln-support )
- [x] Change lock backend driver ( as content here https://github.com/vexxhost/atmosphere/pull/582 ) ( https://github.com/vexxhost/staffeln/issues/111 )

mnaser:
@ricolin The image has been published to GHCR:

https://github.com/vexxhost/staffeln/pkgs/container/staffeln

PR #588, opened 2023-09-18
feat: Staffeln support

ricolin:


ricolin:
relate https://github.com/vexxhost/atmosphere/issues/586

ricolin:
fix https://github.com/vexxhost/atmosphere/issues/593
fix https://github.com/vexxhost/atmosphere/issues/586

playbooks/generate_workspace.yml
```diff
@@ -229,6 +229,7 @@
           openstack_helm_endpoints_magnum_registry_host: "container-infra-registry.{{ domain_name }}"
           openstack_helm_endpoints_rgw_host: "object-store.{{ domain_name }}"
           openstack_helm_endpoints_manila_api_host: "share.{{ domain_name }}"
+          openstack_helm_endpoints_staffeln_api_host: "staffeln.{{ domain_name }}"
```
okozachenko1203, line 232: staffeln-api is only consumed by cinder, isn't it? I doubt if public endpoint is needed.

roles/cinder/tasks/main.yml
```diff
@@ -11,6 +11,12 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
+#
+- name: Set Staffeln Backup Delete Policey Rule
+  ansible.builtin.set_fact:
+    staffeln_backup_delete_policy:
+      backup:delete: "rule:admin_api or (project_id:%(project_id)s and {{ openstack_helm_endpoints_staffeln_api_host }}/v1/backup?backup_id=%(id)s)"
```
okozachenko1203, line 18: same here. We can just use staffeln k8s service namae?

roles/openstack_helm_endpoints/defaults/main.yml
```diff
@@ -521,3 +521,15 @@ openstack_helm_endpoints_manila_keystone_password: "{{ undef(hint='You must spec
 openstack_helm_endpoints_manila_mariadb_password: "{{ undef(hint='You must specify a Manila MariaDB password') }}"
 
                                                                    # ]]]
+# .. envvar:: openstack_helm_endpoints_staffeln_api_host [[[
+#
+# API hostname for Staffeln Service
+openstack_helm_endpoints_staffeln_api_host: "{{ undef(hint='You must specify a Staffeln API hostname') }}"
```
okozachenko1203, line 527: same here. I don't think we need to push users to set staffeln api host.

roles/staffeln/tasks/main.yaml
```diff
@@ -0,0 +1,31 @@
...
+    kubeconfig: /etc/kubernetes/admin.conf
+    values: "{{ _staffeln_helm_values | combine(staffeln_helm_values, recursive=True) }}"
+
+- name: Create Ingress
+  ansible.builtin.include_role:
+    name: openstack_helm_ingress
+  vars:
+    openstack_helm_ingress_endpoint: staffeln
+    openstack_helm_ingress_service_name: staffeln-api
+    openstack_helm_ingress_service_port: 8808
```
okozachenko1203, line 31: if we don't need public api, we have no need for ingress also.

charts/staffeln/templates/deployment-conductor.yaml
```diff
@@ -0,0 +1,156 @@
...
+{{- $mounts_staffeln_conductor := .Values.pod.mounts.staffeln_conductor.staffeln_conductor }}
+{{- $mounts_staffeln_conductor_init := .Values.pod.mounts.staffeln_conductor.init_container }}
+
+{{- $serviceAccountName := "staffeln-conductor" }}
+{{ tuple $envAll "conductor" $serviceAccountName | include "helm-toolkit.snippets.kubernetes_pod_rbac_serviceaccount" }}
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: staffeln
```
okozachenko1203, line 27: I think you forgot to remove this :)

roles/openstack_helm_endpoints/vars/main.yml
```diff
@@ -534,3 +534,18 @@ _openstack_helm_endpoints_sharev2:
...
+        password: "{{ openstack_helm_endpoints_staffeln_mariadb_password }}"
+  staffeln:
+    scheme:
+      public: https
+    host_fqdn_override:
+      public:
+        host: "{{ openstack_helm_endpoints_staffeln_api_host }}"
+    port:
+      api:
+        public: 443
```
okozachenko1203, line 551: same here. If we don't need public endpoint, we don't need to define these.

charts/staffeln/templates/secret-db.yaml
```diff
@@ -0,0 +1,33 @@
...
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ $secretName }}
+type: Opaque
+data:
+{{- if $envAll.Values.manifests.certificates }}
+  DB_CONNECTION: {{ (printf "%s?charset=utf8&ssl_ca=/etc/mysql/certs/ca.crt&ssl_key=/etc/mysql/certs/tls.key&ssl_cert=/etc/mysql/certs/tls.crt&ssl_verify_cert" $connection ) | b64enc -}}
+{{- else }}
```
okozachenko1203, line 29: I think there is a way to generate connection string using helm-toolkit template instead of generating here.

charts/staffeln/templates/openrc.tpl
```diff
@@ -0,0 +1,30 @@
...
+export OS_REGION_NAME="{{ .Values.conf.clouds.clouds.envvars.region_name }}"
+{{- end }}
+{{- if .Values.conf.clouds.clouds.envvars.auth.user_domain_name }}
+export OS_USER_DOMAIN_NAME={{ .Values.conf.clouds.clouds.envvars.auth.user_domain_name }}
+{{- end }}
+{{- if .Values.conf.clouds.clouds.envvars.auth.password }}
+export OS_PASSWORD="{{ .Values.conf.clouds.clouds.envvars.auth.password }}"
+{{- end }}
+{{- if .Values.conf.clouds.clouds.envvars.auth.username }}
+export OS_USERNAME="{{ .Values.conf.clouds.clouds.envvars.auth.username }}"
```
okozachenko1203, line 28: Could you check if we can use `helm-toolkit.snippets.keystone_secret_openrc` to generate openrc using htk?
Also instead of setting cloud config in `values.conf.clouds.clouds.envvars` , we fetch from `values.endpoints` and it is more helm-ly way i think.
You can refer to `secret-keystone.yaml` file of other charts.

charts/staffeln/templates/deployment-conductor.yaml
```diff
@@ -0,0 +1,151 @@
...
+metadata:
+  name: {{ $serviceAccountName }}-leases
+subjects:
+  - kind: ServiceAccount
+    name: {{ $serviceAccountName }}
+    namespace: {{ $envAll.Release.Namespace }}
+roleRef:
+  kind: ClusterRole
+  name: {{ $serviceAccountName }}
+  apiGroup: rbac.authorization.k8s.io
```
okozachenko1203, line 55: well, i think we don't need cluster-scope resources as long as we use the lease in the same namespace.

okozachenko1203:
If we can have a molecule scenario for staffeln (making a backup simply to test its basic functions) and run that in github workflow only when the stafflen-related files(role or chart) changed, it will be perfect

charts/staffeln/templates/deployment-conductor.yaml
ricolin: will change it to Role

charts/staffeln/templates/secret-db.yaml
ricolin: That code was directly from other services like nova, cinder, etc. we can do them all together, but didn't see other route for this

charts/staffeln/templates/openrc.tpl
ricolin: it does not goes in directly, but let me do some string replace and put it in followup pr

ricolin: I remember why I didn't use keystone_secret_openrc at first place. Because staffeln might use different account to format openrc here, and we need be allow it reset

mnaser (approved):

PR #602, opened 2023-09-28
fix(manila): fix host config to respect fqdn

ricolin:


ricolin:
some running result in my env:
root@ctl1:/home/ubuntu# kubectl -nopenstack exec -it manila-share-c7d7fb5b-pr4v2 -- /bin/bash
...
--------------------------------------------
manila@ctl1:/$ cat /tmp/pod-shared/manila-share-fqdn.conf 
[DEFAULT]
host = ctl1
--------------------------------------------
manila@ctl1:/$ cat /tmp/manila-share.sh 
#!/bin/bash



set -ex
exec manila-share \
     --config-file /etc/manila/manila.conf \
     --config-file /tmp/pod-shared/manila-share-fqdn.conf

migmeneses (approved):

mnaser:
@ricolin hmm, i think that added another issue indirectly..

before we were relying on `host` being the same everywhere so that things go up.. now, that won't be the case because if the pod moves from one host to another, the host will change

can you check if manila will automatically failover/move these things over..? otherwise we might be in trouble if a vm gets moved from one place to another

ricolin:
@mnaser  the host binding seems only work when manila share service exist on that host. So once we move pod from ctl1 to ctl2, existing share server will not be able to extend. 
Note: new share will honor new host and build a new share instance and use ctl2 as host bound (that can extend and create new)

@okozachenko1203 do you remember why share server only do 1 replica?

mnaser:
@ricolin the only thing I can imagine as a solution for this would be to run something that runs this:

> The manila-manage utility now has a new command to update the host attribute of shares. This is useful when the share manager process has been migrated to a different host, or if changes are made to the host config option or the backend section name in manila.conf. Execute manila-manage share update_host -h to see usage instructions.

That sounds very ugly though... we can run multiple replicas of `manila-share` but then how do we ensure HA, because if one of them dies (or is being failed over), then all shares on it aren't acccessible

Cinder solved this by adding https://docs.openstack.org/cinder/latest/contributor/high_availability.html -- maybe Manila has the same?

ricolin:
> @ricolin the only thing I can imagine as a solution for this would be to run something that runs this:
> 
> > The manila-manage utility now has a new command to update the host attribute of shares. This is useful when the share manager process has been migrated to a different host, or if changes are made to the host config option or the backend section name in manila.conf. Execute manila-manage share update_host -h to see usage instructions.
> 
> That sounds very ugly though... we can run multiple replicas of `manila-share` but then how do we ensure HA, because if one of them dies (or is being failed over), then all shares on it aren't acccessible
> 
> Cinder solved this by adding https://docs.openstack.org/cinder/latest/contributor/high_availability.html -- maybe Manila has the same?

I see there still related spec for active-active manila-share and not yet implemented https://blueprints.launchpad.net/manila/+spec/manila-share-ha

mnaser (approved):

PR #608, opened 2023-09-29
chore(manila): use uwsgi

okozachenko1203:


mnaser:
@okozachenko1203 we created a conflict :( can you fix please?

Issue #613, opened 2023-10-02
keycloak frequent fails in CI


ricolin:
Keycloak freqnently failed in CI this week https://github.com/vexxhost/atmosphere/actions/workflows/keycloak.yml

Few fails we can observed:

```
  TASK [vexxhost.atmosphere.horizon : Deploy Helm chart] *************************
  fatal: [keycloak]: FAILED! => {"changed": false, "command": "/usr/bin/helm upgrade -i --reset-values --create-namespace -f=/tmp/tmpw95t8501.yml horizon /usr/local/src/horizon", "msg": "Failure when executing Helm command. Exited 1.\nstdout: Release \"horizon\" does not exist. Installing it now.\n\nstderr: Error: failed post-install: timed out waiting for the condition\n", "stderr": "Error: failed post-install: timed out waiting for the condition\n", "stderr_lines": ["Error: failed post-install: timed out waiting for the condition"], "stdout": "Release \"horizon\" does not exist. Installing it now.\n", "stdout_lines": ["Release \"horizon\" does not exist. Installing it now."]}
```
from https://github.com/vexxhost/atmosphere/actions/runs/6378277920/job/17308527600

```
  TASK [vexxhost.atmosphere.keycloak : Disable pxc strict mode] ******************
  fatal: [keycloak]: FAILED! => {"changed": false, "msg": "unable to connect to database, check login_user and login_password are correct or /root/.my.cnf has the credentials. Exception message: (2013, 'Lost connection to MySQL server during query')"}
```
from https://github.com/vexxhost/atmosphere/actions/runs/6376458851/job/17309567121

```
Run poetry run molecule test -s keycloak
  poetry run molecule test -s keycloak
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.10.13/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.10.13/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.10.13/x64
[...]

PR #615, opened 2023-10-02
chore(CI): keycloak limit verify scope

ricolin:
fix https://github.com/vexxhost/atmosphere/issues/613


mnaser:
I think we can skip on this for now, we don't need to reduce coverage since things have become a lot more stable.

PR #616, opened 2023-10-02
CI: Enable debug by default

ricolin:
I wonder if this is too radical, but we definitely gonna benefit from it.


mnaser:
@ricolin I don't know if this will provide much value, since most failures happen within the cloud and we don't get much details from the pod logs / etc on the failures.

It would probably make sense for us to figure out some way of exporting all those logs on failures, but GHA doesn't give us a real nice way of getting logs out of it.

PR #617, opened 2023-10-02
fix: Use a larger runner for keycloak CI jobs

okozachenko1203:
fix https://github.com/vexxhost/atmosphere/issues/613

### DIFF
--- a/.github/workflows/keycloak.yml
+++ b/.github/workflows/keycloak.yml
@@ -22,7 +22,7 @@ on:
 
 jobs:
   molecule:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/keycloak.yml
```diff
@@ -22,7 +22,7 @@ on:
 
 jobs:
   molecule:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04-16-cores
```
mnaser, line 25: can you use `ubuntu-latest-4-cores` so we don't overspend 👍

mnaser (changes requested):

Issue #505, opened 2023-07-31
Add support for retrieving `KUBECONFIG` via dashboard


mnaser:
I'm a bit in awe I'm writing this but..

there's literally no way of getting the `KUBECONFIG` for a Magnum cluster using the OpenStack dashboard, we should fix this usability issue.

ricolin:
patches: 
* https://review.opendev.org/c/openstack/python-magnumclient/+/890893
* https://review.opendev.org/c/openstack/magnum-ui/+/890894
Will push python-magnumclient for a release first.

ricolin:
pushing release for magnum client https://review.opendev.org/c/openstack/releases/+/891565

ricolin:
The rest patch are all merged, once magnum client released (https://review.opendev.org/c/openstack/releases/+/891565) we can start pushing stuff into our codebase

mnaser:
@ricolin can we bump the image and test this please?

ricolin:
> @ricolin can we bump the image and test this please?

Sorry, totally fail to follow up this task, bumping it and testing now

PR #620, opened 2023-10-03
chore(magnum): bump magnum image

ricolin:
fix https://github.com/vexxhost/atmosphere/issues/505


ricolin:
it's not about magnum image version in this case so I closed this for now

PR #624, opened 2023-10-04
chore(selin): use wsgi

okozachenko1203:


okozachenko1203:
seems we don't have enough github runners ?

Issue #626, opened 2023-10-04
Services worker count inconsistent


mnaser:
At the moment, Cinder is using a single worker which is probably having some performance impact in high load scenarios.

We should move it to at least 8 like other services, and make sure that in Molecule's shared.yml file we set it to 2 to not use up a lot of resources.

https://github.com/vexxhost/atmosphere/blob/103c85edc6f8e40a2d44476ed879193f82336a6f/charts/cinder/values.yaml#L800

mnaser:
We should do an audit of all services and use 8 workers in prod / 2 workers in molecule

PR #628, opened 2023-10-05
chore(cinder): Increase osapi_volume_workers

ricolin:
fix: https://github.com/vexxhost/atmosphere/issues/626


mnaser (approved):

ricolin:
@mnaser weird, the check job for this pr always stuck in queued, because it's approved?

PR #631, opened 2023-10-06
fix(ovs): accept space separated string for ovs_options for dpdk bonds

thywyn:
We had a requirement for our DPDK bonds to have more than one option set during creation.  This change allows a space separated string to be used in the inventory.

```yaml
neutron_helm_values:
  conf:
    ovs_dpdk:
      bonds:
        - name: dpdkbond
        - ovs_options: 'bond_mode=balance-tcp lacp=active'
...
```

### DIFF
--- a/charts/neutron/templates/bin/_neutron-openvswitch-agent-init.sh.tpl
+++ b/charts/neutron/templates/bin/_neutron-openvswitch-agent-init.sh.tpl
@@ -25,8 +25,8 @@ OVS_CTL=/run/openvswitch/ovs-vswitchd.${OVS_PID}.ctl
 chown neutron: ${OVS_CTL}
 
 function get_dpdk_config_value {
-  values=$1
-  filter=$2
+  values=${@:1:$#-1}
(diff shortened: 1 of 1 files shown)


ricolin:
@thywyn I think this looks pretty valid and ready to merge. On the other hand, are you interested to add this to [1]? Or you like me to have the honor to added it to openstack/openstack-helm?

We can sync charts from there so we can make sure no regression for this fix. This little part I can help with [2].
In either way, I think this PR will be landed this week if no other concerns

[1] https://github.com/openstack/openstack-helm/blob/master/neutron/templates/bin/_neutron-openvswitch-agent-init.sh.tpl#L28-L29
[2] https://github.com/vexxhost/atmosphere/blob/main/hack/sync-charts.sh#L133-L141

thywyn:
@ricolin I've submitted an issue via this issue...   https://storyboard.openstack.org/#!/story/2010934 .. If you  can submit it upstream by all means!

PR #632, opened 2023-10-06
Add tasks to create projects and define project quotas

etimmons:


mpiscaer:
Whould it make sense to create its own role so that you can run only this without the risk to deploy config changes to keystone?

Issue #643, opened 2023-10-13
CI: image-manifest meet running out of space


ricolin:
see error here: https://github.com/vexxhost/atmosphere/actions/runs/6509784800/workflow

I guess that means it's time for us to either not cache for action or bump flavor

okozachenko1203:
hmm... i think we just cache poetry deps, not image blobs.

okozachenko1203:
```sh
runner@fv-az573-11:~/work/atmosphere/atmosphere$ df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        84G   67G   17G  80% /
tmpfs           3.4G  172K  3.4G   1% /dev/shm
tmpfs           1.4G  1.1M  1.4G   1% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
/dev/sdb15      105M  6.1M   99M   6% /boot/efi
/dev/sda1        14G  4.1G  9.0G  31% /mnt
tmpfs           693M   12K  693M   1% /run/user/1001
```

okozachenko1203:
```sh
Filesystem      Size  Used Avail Use% Mounted on
tmpfs           3.2G  1.1M  3.2G   1% /run
/dev/vda1       155G   62G   94G  40% /
tmpfs            16G  172K   16G   1% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
/dev/vda15      105M  6.1M   99M   6% /boot/efi
tmpfs           3.2G   12K  3.2G   1% /run/user/1000
```

okozachenko1203:
```sh
ubuntu@gha-gxzcy:/opt/github/actions-runner/2.309.0/_work/atmosphere/atmosphere$ df -h
Filesystem      Size  Used Avail Use% Mounted on
tmpfs           3.2G  1.2M  3.2G   1% /run
/dev/vda1       155G   80G   76G  52% /
tmpfs            16G  172K   16G   1% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
/dev/vda15      105M  6.1M   99M   6% /boot/efi
tmpfs           3.2G   12K  3.2G   1% /run/user/1000
```

okozachenko1203:
18GB is required for all images, but ubuntu runner has 17GB available space

PR #644, opened 2023-10-13
fix(CI): image-manifest and ceph(ovs/ovn)

ricolin:
This includes:
* Use larger disk space for image-manifest

fix: https://github.com/vexxhost/atmosphere/issues/643
fix: https://github.com/vexxhost/atmosphere/issues/648

### DIFF
--- a/.github/workflows/image-manifest.yml
+++ b/.github/workflows/image-manifest.yml
@@ -22,7 +22,7 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: v3-standard-8
     steps:
       - name: Checkout project
         uses: actions/checkout@v4
--- a/roles/keycloak/tasks/main.yml
+++ b/roles/keycloak/tasks/main.yml
@@ -55,7 +55,7 @@
     host: "%"
     priv: "{{ keycloak_database_name }}.*:ALL"
 
-- name: Enable pxc strict mode
+- name: Disable pxc strict mode
   community.mysql.mysql_query:
     login_host: "{{ _pxc_service.resources[0].spec.clusterIP }}"
     login_user: root
@@ -87,7 +87,7 @@
     ingress_annotations:
       cert-manager.io/cluster-issuer: atmosphere
 
-- name: Disable pxc strict mode
+- name: Eable pxc strict mode
   community.mysql.mysql_query:
     login_host: "{{ _pxc_service.resources[0].spec.clusterIP }}"
     login_user: root
--- a/roles/keycloak/tasks/main.yml
+++ b/roles/keycloak/tasks/main.yml
@@ -20,9 +20,21 @@
     namespace: openstack
   register: _pxc_service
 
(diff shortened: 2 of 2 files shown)


ricolin:
IIUC, this should provide us bigger disk space :)

roles/keycloak/tasks/main.yml
```diff
@@ -20,10 +20,22 @@
...
+    login_host: "{{ _pxc_service.resources[0].spec.clusterIP }}"
+    login_user: root
+    login_password: "{{ openstack_helm_endpoints.oslo_db.auth.admin.password }}"
+    filter:
+      - version
+  register: mysql_ready
+  until: mysql_ready is not failed
+  retries: 10
+  delay: 5
+
```
okozachenko1203, line 38: @ricolin i suggest making PRs separately for image-manifest fix and keycloak fix.

ricolin: I'm fine either way, but now both stuck at the same time though. It requires both to unblock IMO. Do we allow with force merging?

roles/keycloak/tasks/main.yml
```diff
@@ -43,7 +55,7 @@
     host: "%"
     priv: "{{ keycloak_database_name }}.*:ALL"
 
-- name: Disable pxc strict mode
+- name: Enable pxc strict mode
```
okozachenko1203, line 58: I think Disable is correct

okozachenko1203: https://github.com/vexxhost/atmosphere/blob/main/roles/keycloak/tasks/main.yml#L78
And this should be `Enable` i think

.github/workflows/image-manifest.yml
```diff
@@ -22,7 +22,7 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: v3-standard-8
```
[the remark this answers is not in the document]
mnaser: New flavor is a bit difficult, I'd rather stick to our existing ones.

roles/keycloak/tasks/main.yml
```diff
@@ -75,7 +87,7 @@
     ingress_annotations:
       cert-manager.io/cluster-issuer: atmosphere
 
-- name: Disable pxc strict mode
+- name: Eable pxc strict mode
```
mnaser, line 90:
```suggestion
- name: Enable pxc strict mode
```

mnaser (changes requested):

roles/keycloak/tasks/main.yml
okozachenko1203: how about v3-standard-6?

mnaser: I don't know if we want to add a whole another pool of that flavor, since we're running it inside our cloud, the cost isn't much to worry about that this point :)

mnaser (approved):

PR #650, opened 2023-10-17
Check pxc ready

ricolin:


roles/percona_xtradb_cluster/tasks/main.yml
```diff
@@ -47,3 +47,15 @@
     wait_condition:
       type: ready
```
ricolin, line 48: this test is more checking where mysql is not ready for keycloak db careate, as it should be ready by this wait condition

PR #663, opened 2023-10-19
fix: OVS 2.15 changed terminology slave to member

thywyn:
with new OVS installs the ovs-agent does not become ready.  Only applies if you are enabling DPDK


mnaser:
Nice catch @thywyn

PR #664, opened 2023-10-20
chore: pin ingress-nginx to the control plane

### LINKED ISSUE #657: The follow services are not running in the control plane
ingress-nginx-defaultbackend-5c44d84866-5hf42
coredns-788c76bc67-4sd6m
coredns-788c76bc67-bvgjf
loki-gateway-789b47cbd5-pjc6n

mpiscaer:
chore: pin ingress-nginx to the control plane to fix #657

### DIFF
--- a/roles/ingress_nginx/vars/main.yml
+++ b/roles/ingress_nginx/vars/main.yml
@@ -42,6 +42,8 @@ _ingress_nginx_helm_values:
     metrics:
       enabled: true
   defaultBackend:
+    nodeSelector:
+      openstack-control-plane: enabled
     enabled: true
     image:
       registry: "{{ atmosphere_images['ingress_nginx_default_backend'] | vexxhost.kubernetes.docker_image('domain') }}"


charts/ingress-nginx/values.yaml
```diff
@@ -824,6 +824,7 @@ defaultBackend:
   ##
   nodeSelector:
     kubernetes.io/os: linux
+    openstack-control-plane: enabled
```
mnaser, line 827: This will get lost the next time we sync the charts, so what we need to do for this is actually add it into `roles/ingress_nginx/vars/main.yml` into `_ingress_nginx_helm_values`.

mnaser (dismissed):

charts/ingress-nginx/values.yaml
mpiscaer: Thanks for pointing it out.

okozachenko1203 (approved):

PR #668, opened 2023-10-25
chore(manila): bump manila chart version

ricolin:
This will need to wait for https://review.opendev.org/c/openstack/manila/+/899151 to merged


mnaser:
We can close this, this is already updated to latest :)

PR #680, opened 2023-10-31
fix(neutron): Nit fix for config `neutron/ovn/enable_distributed_floating_ip`

ricolin:


mnaser (changes requested):
@ricolin We have explicitly disabled this because it has caused us a lot of network connectivity issues, i think it's not a very stable feature for now unfortunately.

ricolin:
@mnaser why this can't be disabled from roles? https://github.com/vexxhost/atmosphere/pull/680/commits/429020d9032dd3197fefc47c2d1c9fe800f7716f#diff-1c5c2ad5f0f97e5d10ff8f9d4cdf157e025be857679f01d78ac98891d293e063R79

mnaser:
Oh I see. The charts default to true and we are forcing false. Ok good

mnaser:
Can you re-open a new PR please?

Issue #681, opened 2023-11-01
Sync OSH charts version


ricolin:
checking list:
- [x] barbican
- [x] cinder
- [x] heat
- [x] keystone
- [ ] loki
- [x] neutron
- [x] openvswitch
- [ ] pxc-operator
- [x] senlin
- [ ] ceph-csi-rbd
- [ ] coredns
- [x] horizon
- [ ] kube-prometheus-stack
- [x] magnum
- [ ] node-feature-discovery
- [x] ovn 
- [ ] rabbitmq-cluster-operator
- [x] staffeln
- [ ] ceph-provisioners
- [x] designate
- [ ] ingress-nginx
- [x] libvirt
- [x] manila
- [x] nova
- [x] placement
- [ ] rook-ceph
- [x] tempest
- [x] cert-manager-webhook-infoblox-wapi
- [x] glance
- [ ] keycloak
- [x] local-path-provisioner
- [ ] memcached
- [x] octavia
- [ ] prometheus-pushgateway
- [ ] rook-ceph-cluster
- [ ] vector

ricolin:
I think we can leave the rest as what it is for now.

PR #685, opened 2023-11-01
chore(glance): bump glance charts version

ricolin:
relate to https://github.com/vexxhost/atmosphere/issues/681


ricolin:
moved to https://github.com/vexxhost/atmosphere/pull/688/commits

PR #694, opened 2023-11-03
fix: wait will the ingress has been created

mpiscaer:
fix: wait will the ingress has been created

At creating the ingress it needs to wait till it has been created. Especially when using ACME certificates this can take longer, where the next task will fail there the ingress does not exists yet.

### DIFF
--- a/roles/ingress/tasks/main.yml
+++ b/roles/ingress/tasks/main.yml
@@ -17,3 +17,4 @@
   kubernetes.core.k8s:
     state: present
     template: ingress.yml.j2
+    wait: true


mnaser:
@mpiscaer  historically we've had retries on the things that can possible not be ready such as Nova etc so we don't have to wait everywhere and slow down the deploy. 

Does it make sense to add retries instead for where you're trying to fix this for?

mnaser (changes requested):
I think it's better to fix this at the resource creation level rather than always waiting

mpiscaer:
Keystone was failing, i will see what i can do to check if the ingress is working before creating the domain.

I don't like to do retries for this on creating domain(for example).

PR #698, opened 2023-11-03
fix: LoadBalancer Health not coming online in OVN/DPDK environment

thywyn:
healthcheck security group was not forwarding 5555/udp traffic from amphora for it's health.  TCP is not needed for 5555. Does not work in OVN/DPDK environment without UDP specified.  My best guess is that in other environments the 5555/UDP traffic was leaking to the health-manager interfaces..


mnaser:
@thywyn looks like i caught/saw @ricolin fix at #699 which has coverage for both udp/tcp since logging traffic can be switched.  I've merged that one in the meantime.  Thank you

PR #704, opened 2023-11-08
chore(prometheus-pushgateway,ceph-csi-rbd):Bump reinstall required chartts

ricolin:
I separate these two charts as they requires reinstall helm chart or delete kubernetes/deploy to make it work

prometheus-pushgateway:
This will get `v1.labelselectorrequirement(nil)}: field is immutable` issue

ceph-csi-rbd got issue when upgrade to >3.7.0 https://github.com/ceph/ceph-csi/issues/3397
https://github.com/Syphdias/ceph-csi/blob/devel/charts/ceph-csi-rbd/README.md#known-issues-upgrading


mnaser:
@ricolin  the body of the PR doesn't seem to match with the content, did the content of the PR change?

ricolin:
> @ricolin the body of the PR doesn't seem to match with the content, did the content of the PR change?

@mnaser yeah, removed those overlapping commits

ricolin:
Convert it to draft because I plan to directly add the delete resource part in directly

PR #707, opened 2023-11-10
chore: staffeln values

mpiscaer:
Setting the values so that it backups only once a day.

### DIFF
--- a/charts/staffeln/values.yaml
+++ b/charts/staffeln/values.yaml
@@ -269,7 +269,8 @@ conf:
     conductor:
       backup_metadata_key: __staffeln_backup
       retention_metadata_key: __staffeln_retention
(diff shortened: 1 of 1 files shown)


charts/staffeln/values.yaml
```diff
@@ -269,6 +269,10 @@ conf:
     conductor:
       backup_metadata_key: __staffeln_backup
       retention_metadata_key: __staffeln_retention
+      backup_service_period: 86400
```
ricolin, line 272: If we like to make sure volume get backup around 86400 seconds, we can set `backup_min_interval: 86400` here and set `backup_service_period` to some smaller amounts. 


ref: https://github.com/vexxhost/staffeln/blob/main/staffeln/conf/conductor.py#L26

Issue #728, opened 2023-11-23
bug: renewed certificates for libvirt are not reflected


okozachenko1203:
## Context
We use TLS for vnc session encryption and auth, and for libvirt api. All these TLS certificates are generated per a libvirt pod (issued inside the initContainer) and managed by cert-manager.
Cert-manager does its job well so all certs are renewed automatically without a doubt but renewed certs are not reflected to libvirt.
So after 90 days (90d is the default cert expiry date in cert-manager), all libvirt pods are stuck because of expired certificates. This is a kinda huge incident because we cannot do any actions on hypervisors because of this issue.

## Reason
We fetch the cert content inside the initContainer of libvirt pods by using these scripts 
https://github.com/vexxhost/atmosphere/blob/main/charts/libvirt/values.yaml#L276-L321
https://github.com/vexxhost/atmosphere/blob/main/charts/libvirt/templates/bin/_libvirt.sh.tpl#L21-L42
As you can see, the cert contents are fetched at the generation time only once then copied to the desired paths.
So the renewed certs are not reflected.

## Workaround for now
Just rollout the libvirt so the new pods are created and also new certs are issued for them.

Need to find the way to reflect the renewed certs.

okozachenko1203:
@mnaser what is your opinion?

mnaser:
@okozachenko1203 i think there is a CSI for cert-manager but I don't believe this would work well because we need the pod name in order to be able to generate the certificate. 

Can cert-manager give us a dynamic certificate based on pod name using the CSI module?  If so we can get rid of the initContainer...

okozachenko1203:
the bad thing is cert-man csi-driver doesn't support POD_IP
```sh
  Warning  FailedMount  35s (x8 over 100s)  kubelet            MountVolume.SetUp failed for volume "api-tls" : rpc error: code = Unknown desc = generating certificate signing request: "csi.cert-manager.io/common-name": undefined variable "POD_IP", known variables: [POD_NAME POD_NAMESPACE POD_UID SERVICE_ACCOUNT_NAME]
```

mnaser:
@okozachenko1203 do we need the IP address for the cert?  could we start using hostname instead?

guilhermesteinmuller:
Hi -- I would prioritizer this. We just got another cloud which no vms were being created due to this

mnaser:
The errors:

```
[...]

PR #734, opened 2023-11-29
feat: use csi driver for libvirt certificates

okozachenko1203:
fix https://github.com/vexxhost/atmosphere/issues/728


mnaser:
We can close this since we use `libvirt-tls-sidecar`.

Issue #736, opened 2023-11-30
Cluster API proxy needs to run on all nodes with OVN


mnaser:
When OVN is involved, there's no DHCP agent running.  Therefore, the `qdhcp` namespaces do not exist on the control plane, therefore the only to jump into the network to create the isolated clusters is to go through `ovnmeta-`

If cluster is running OVN, we need to run Cluster API proxy across all OvS nodes, otherwise keep as is.

PR #750, opened 2023-12-11
fix: set nodeSelector for mcapi-proxy according to the network backen…

okozachenko1203:
…d type

fix https://github.com/vexxhost/atmosphere/issues/736

### DIFF
--- a/roles/magnum/tasks/main.yml
+++ b/roles/magnum/tasks/main.yml
@@ -96,13 +96,6 @@
     kubeconfig: /etc/kubernetes/admin.conf
     values: "{{ _magnum_helm_values | combine(magnum_helm_values, recursive=True) }}"
 
-- name: Set nodeSelector of magnum-cluster-api-proxy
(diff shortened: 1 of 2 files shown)


roles/magnum/tasks/main.yml
```diff
@@ -96,6 +96,13 @@
     kubeconfig: /etc/kubernetes/admin.conf
     values: "{{ _magnum_helm_values | combine(magnum_helm_values, recursive=True) }}"
 
+- name: Set nodeSelector of magnum-cluster-api-proxy
+  ansible.builtin.set_fact:
+    _magnum_cluster_api_proxy_ovs_node_selector:
+      openstack-control-plane: enabled
+    _magnum_cluster_api_proxy_ovn_node_selector:
+      openvswitch: enabled
```
mnaser, line 104: no need to set fact here.. this can go into vars

mnaser (approved):

PR #751, opened 2023-12-11
doc: update designate doc for bind9 integration

okozachenko1203:


mnaser (changes requested):
We have all the documentation now inside `docs` using Sphinx, can you please move this there? :)

Issue #747, opened 2023-12-06
Add VPNaaS for OVN


mnaser:
At the moment, upstream is fully stalled out on merging OVN support:

- https://review.opendev.org/c/openstack/neutron-tempest-plugin/+/847007
- https://review.opendev.org/c/openstack/neutron-vpnaas/+/765353
- https://review.opendev.org/c/openstack/neutron-tempest-plugin/+/889294
- https://review.opendev.org/c/openstack/neutron-vpnaas/+/895651

Since we've got users and customers that rely on this functionality, we're going to start maintaining a fork that has this fix.  We need the following work items:

- [x] Fork `neutron-vpnaas` and backport all the way to `stable/zed`
- [x] Build images with our fork of `neutron-vpnaas`
- [x] Update Helm charts to support the new agent for VPN with OVN

mnaser:
I merged the backports but I was able to +W the `neutron-vpnaas` change into master which is neat.

I kinda did a merge commit instead of rebase which means 2 commits got added on top... do you feel like recreating the repo and creating the PRs again so I can merge with rebase @ricolin ?

ricolin:
> I merged the backports but I was able to +W the `neutron-vpnaas` change into master which is neat.
> 
> I kinda did a merge commit instead of rebase which means 2 commits got added on top... do you feel like recreating the repo and creating the PRs again so I can merge with rebase @ricolin ?

I just reset repo and PRs

mnaser:
@ricolin good catch. I think we just need to test this manually now if possible.

ricolin:
> @ricolin good catch. I think we just need to test this manually now if possible.

workon build newest env for test now

PR #757, opened 2023-12-18
chore: use forked repo for neutron-vpnaas

ricolin:
relate to https://github.com/vexxhost/atmosphere/issues/747


internal/pkg/image_repositories/build_workflow.go
```diff
@@ -7,11 +7,12 @@ import (
...
-	"keystone":  true,
-	"magnum-ui": true,
-	"magnum":    true,
-	"neutron":   true,
+	"horizon":        true,
+	"keystone":       true,
+	"magnum-ui":      true,
+	"magnum":         true,
+	"neutron":        true,
+	"neutron-vpnaas": true,
```
mnaser, line 15: i dont think this is gonna work since vpnaas is installed as a pip package right now i believe in the same venv

mnaser (changes requested):

Issue #760, opened 2023-12-19
Disable `neutron-coredns` with OVN


mnaser:
Since #539 is a thing that won't be resolved easily or quickly, we should disable the installation of `neutron-coredns` if OVN is in place.

PR #761, opened 2023-12-19
chore: disable neutron coredns on ovn backend

ricolin:
fix https://github.com/vexxhost/atmosphere/issues/760


roles/coredns/tasks/main.yml
```diff
@@ -14,6 +14,7 @@
 
 - name: Uninstall the legacy HelmRelease
   run_once: true
+  when: atmosphere_network_backend != 'ovn'
```
mnaser, line 17: we can drop this here, its not necessary

mnaser (changes requested):
small nit but also we need to remove this here too:

https://github.com/vexxhost/atmosphere/blob/main/roles/neutron/vars/main.yml#L78

mnaser (approved):

PR #767, opened 2023-12-21
fix: configure ovn-bridge with netdev datapath type when dpdk is enabled

thywyn:
Needing to automatically configure a netdev datapath type for ovn-bridge (default: br-int) for DPDK enabled OVN/OVS


thywyn:
Reference for PR https://github.com/ovn-org/ovn/blob/ef45415a6148461c56f87523beb6865e13742f30/controller/ovn-controller.c#L542C32-L542C32

PR #775, opened 2023-12-27
feat: Adding option to use interface names instead of just pci_ids 

thywyn:
This script change helps alleviate the issue of heterogeneous server environments having a consistent pci_id.  This change will enable usage of host interface names which can be set to a consistent value despite pci_id differing when deploying the server.  Defaults to using pci_id parameter if present for backwards compatibility.

We are using this in our environment, use it if you'd like.


thywyn:
any way you all will consider  putting  this in?

mnaser:
recheck

thywyn:
Rebased and redid this PR as a patch file for the neutron chart.

thywyn:
fixed accidental pasted line

mnaser:
already landed

Issue #861, opened 2024-01-08
python-memcached 1.61 causes errors in Horizon


etimmons:
_deletetouch() has been removed from python-memcached as of 1.60 and is causing errors in horizon

```
AttributeError: 'Client' object has no attribute '_deletetouch'
```

see https://github.com/django/django/pull/17648

as noted in the link above, this only affects django 3.2

etimmons:
also: https://stackoverflow.com/a/77757428

mnaser:
fixed, thanks @etimmons

PR #862, opened 2024-01-08
fix: pin python-memcached==1.59 in upper-constraints.txt for 2023.2

etimmons:
python-memcached 1.60 made breaking changes, e.g. _deletetouch() has been removed

see https://github.com/django/django/pull/17648


etimmons:
Created issue here:
https://github.com/vexxhost/atmosphere/issues/861

Issue #880, opened 2024-01-12
Implement ssh key access to amphoras


guilhermesteinmuller:
There are cases in which the operator needs to inspect amphoras vms by logging into its respective virtual machines via ssh so it may be helpful to log into running amphorae when troubleshooting service problems.

PR #883, opened 2024-01-15
feat: add amphora jumpbox on k8s

okozachenko1203:
fix https://github.com/vexxhost/atmosphere/issues/880


roles/octavia/tasks/generate_resources.yml
```diff
@@ -152,3 +152,89 @@
...
+  run_once: true
+  block:
+    - name: Create amphora ssh key
+      kubernetes.core.k8s:
+        state: present
+        definition:
+          - apiVersion: secretgen.k14s.io/v1alpha1
+            kind: SSHKey
+            metadata:
+              name: amphora-ssh-key
```
mnaser, line 166: Let's make this `octavia-amphora-ssh-key` in order to make it clear that it's part of the Octavia things.

roles/octavia/tasks/generate_resources.yml
```diff
@@ -152,3 +152,89 @@
...
+        api_version: apps/v1
+        kind: SSHKey
+        name: amphora-ssh-key
+        namespace: openstack
+        wait_sleep: 1
+        wait_timeout: 60
+        wait: true
+        wait_condition:
+          type: ReconcileSucceeded
+          status: true
```
mnaser, line 181: We don't need a separate task to wait over here, see this:

https://github.com/vexxhost/atmosphere/blob/main/roles/rabbitmq/tasks/main.yml#L49-L53

roles/octavia/tasks/generate_resources.yml
```diff
@@ -152,3 +152,89 @@
...
+                        - name: sshkey
+                          mountPath: /sshkey
+                          readOnly: true
+                  nodeSelector:
+                    openstack-control-plane: enabled
+                  volumes:
+                    - name: sshkey
+                      secret:
+                        secretName: amphora-ssh-key
+                        defaultMode: 0400 # noqa: yaml[octal-values]
```
mnaser, line 220: Instead of adding a new deployment, why don't we just simply use the `volumeMounts` feature and add it into the `octavia-worker` pod?  I think that's an easier and more straight forward approach (for now).

roles/octavia/tasks/generate_resources.yml
```diff
@@ -152,3 +152,89 @@
...
+      kubernetes.core.k8s_info:
+        api_version: v1
+        kind: Secret
+        name: amphora-ssh-key
+        namespace: openstack
+      register: _octavia_ssh_key_secret
+
+    - name: Cache fact with Percona XtraDB password
+      ansible.builtin.set_fact:
+        _octavia_amphora_ssh_public_key: "{{ _octavia_ssh_key_secret.resources[0]['data']['ssh-authorizedkey'] | b64decode }}"
```
mnaser, line 232: I don't think we need to cache it because we actually just grab it once and only only, so no need to.

roles/secretgen_controller/tasks/main.yml
```diff
@@ -0,0 +1,921 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Deploy service
+  kubernetes.core.k8s:
+    state: present
+    definition:
+      # Note(Oleks): https://github.com/carvel-dev/secretgen-controller/releases/download/v0.16.0/release.yml
```
mnaser, line 19: Can you move the contents of this into `files/release.yml` and then include that into here instead to make this file cleaner?

mnaser (changes requested):

.ansible-lint
```diff
@@ -7,6 +7,7 @@ exclude_paths:
   - playbooks
   - roles/defaults/vars/main.yml
   - roles/kube_prometheus_stack/files/jsonnet
+  - roles/secretgen_controller/tasks/main.yml
```
mnaser, line 10: Shouldn't we remove this now?

roles/octavia/tasks/generate_resources.yml
```diff
@@ -152,3 +152,43 @@
...
+          type: ReconcileSucceeded
+          status: true
+
+    - name: Grab Amphora jumpbox public key
+      kubernetes.core.k8s_info:
+        api_version: v1
+        kind: Secret
+        name: octavia-amphora-ssh-key
+        namespace: openstack
+      register: _octavia_ssh_key_secret
```
mnaser, line 186: Do we actually need this?  Couldn't we register the task above?

roles/octavia/tasks/generate_resources.yml
```diff
@@ -152,3 +152,43 @@
...
+
+    - name: Create Amphora ssh key pair in Openstack
+      openstack.cloud.keypair:
+        cloud: atmosphere
+        state: present
+        name: octavia-amphora-ssh-key
```
mnaser, line 192: Let's call this `{{ octavia_helm_release_name }}-amphora-ssh-key`

roles/octavia/tasks/generate_resources.yml
```diff
@@ -152,3 +152,43 @@
...
+        wait_timeout: 60
+        wait_condition:
+          type: ReconcileSucceeded
+          status: true
+
+    - name: Grab Amphora jumpbox public key
+      kubernetes.core.k8s_info:
+        api_version: v1
+        kind: Secret
+        name: octavia-amphora-ssh-key
```
mnaser, line 184: Let's call this `{{ octavia_helm_release_name }}-amphora-ssh-key`

roles/octavia/tasks/generate_resources.yml
```diff
@@ -152,3 +152,43 @@
     cloud: atmosphere
     image: "{{ octavia_amphora_image_name }}"
   register: _octavia_amphora_image
+
+- name: Deploy Amphora jumpbox
+  run_once: true
+  block:
```
mnaser, line 158: Since this will become 2 tasks only, maybe just make two tasks here and that's it (with `run_once: true`)

roles/secretgen_controller/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Deploy secretgen-controller
+  kubernetes.core.k8s:
+    state: present
+    template: release.yml.tpl
```
mnaser, line 18: Is there any reason this is templated?  You could use `src` if it's not templated and I don't see any templates in there.

Also, is it possible to use `vendir` to sync that file?

mnaser (changes requested):
I got a few things to look at :)

roles/secretgen_controller/tasks/main.yml
okozachenko1203: for image ref, i used template

mnaser (approved):

PR #918, opened 2024-01-26
fix: mount atmosphere ca cert inside the magnum-conductor

okozachenko1203:


mnaser:
@okozachenko1203 : wouldn't this break for non-self signed environments?

Issue #921, opened 2024-01-27
Switch default amphora do active/standby


guilhermesteinmuller:
Currently amphoras are standalone, that's not reliable if we lose a compute node with amphoras within it.  We should change it to active/standby

PR #927, opened 2024-01-29
fix: change loadbalancer topology to active_standby

okozachenko1203:
fix https://github.com/vexxhost/atmosphere/issues/921


mpiscaer:
@mnaser @okozachenko1203 would this be a good idea to get this merge?

mnaser:
@mpiscaer the only thing about this is that this instantly doubles the amount of VMs needed, the goal/idea here I think was to actually try and figure out a method that we can integrate flavors natively with this approach.

Issue #925, opened 2024-01-29
Compress assets during build-time


mnaser:
At this point, Horizon is building all the assets during run-time which is not effective.  Also, for our Atmosphere use case, we don't actually have many changes that require re-compressing all the assets over and over again on each load.

We should compress and collect all the assets during build time and then stop doing it during run-time for a faster roll-out and avoiding the usual race condition that happens if any requests come in before the compression / collection is done.

mnaser:
@etimmons: FYI, this is what we talked about the other day.

okozachenko1203:
@mnaser i think running those commands during the image build time is not good because we mount local_settings.py file using k8s configmap and this file defines `STATIC_ROOT`. Once users change this value instead of using default one, the collected static files are not served correctly.

mnaser:
I see. But I think we should assume the users who are changing that setting know what they're doing. We've never had to change that ever for any reason

fitbeard:
This will add limitation for funcionality like panel remove: https://docs.openstack.org/horizon/latest/configuration/pluggable_panels.html#remove-panel

I have a plan to contribute upstream for horizon chart to add funcionality like we have already for `local_settings.d` where you can do something like this:

```
horizon_helm_values:
  conf:
    horizon:
      custom_panels_d:
        _5000_disable_project_vg_snapshots: |
          # Remove the volume group snapshot menu entry
          # They are curently not supported
          PANEL = 'vg_snapshots'
          PANEL_DASHBOARD = 'project'
          PANEL_GROUP = 'volumes'
          REMOVE_PANEL = True
```

PR #929, opened 2024-01-29
fix: run django collectstatic and compress in horizon image

okozachenko1203:
fix https://github.com/vexxhost/atmosphere/issues/925


mnaser:
@okozachenko1203 : Can you take care of the linter issues?

Issue #964, opened 2024-02-14
Containerize Openstack CLI


fitbeard:
Right now `openstack_cli` role is the only role which relies on OS package manager/debian packages. This can be safely containerized and later used as shell alias with any distro and any Openstack release:

```
# cat /etc/profile.d/atmosphere.sh
alias osc='nerdctl run --rm --network host --volume $PWD:/opt --volume /tmp:/tmp --volume /etc/openstack:/etc/openstack:ro --env-file <(env | grep OS_) 10.13.55.223:5000/openstack-cli:2023.1'
alias openstack='osc openstack'
alias nova='osc nova'
alias neutron='osc neutron'
alias cinder='osc cinder'
alias glance='osc glance'
```

- [x] Create small role in `vexxhost.kubernetes` collection for `nerdctl`. (`nerdctl` is smart enough to automatically pull images, or switch to insecure/http registry if needed, or mount local paths inside container. Unlike `ctr` or `crictl`.)
- [x] Add `openstack-cli` build step for Earthly images
- [x] Rewrite `openstack_cli` role with new logic for alias creation and old OS packages and `apt` repo removal.

That should be the last piece for full Rocky/Alma/Fedora compatibility (with few nit's which can be fixed later).

mnaser:
I like this approach a lot because it also eliminates a lot of the headaches we see with system python versions not matching.

PR #965, opened 2024-02-14
feat: containerized openstack cli

fitbeard:
Related: https://github.com/vexxhost/atmosphere/issues/964


images/openstack-cli/Earthfile
```diff
@@ -0,0 +1,29 @@
...
+    python3-manilaclient \
+    python3-neutronclient \
+    python3-novaclient \
+    python3-octaviaclient \
+    python3-openstackclient \
+    python3-osc-placement \
+    python3-senlinclient \
+    python3-swiftclient"
+#  ARG REGISTRY=ghcr.io/vexxhost/atmosphere
+  ARG REGISTRY=10.13.55.223:5000
```
mnaser, line 21: FYI, in your testing, you can keep this as is but run `earthly ./images/openstack-cli+platform-image --REGISTRY=10.13.55.223:5000`

images/openstack-cli/Earthfile
```diff
@@ -0,0 +1,29 @@
...
+  SAVE IMAGE --push \
+    ${REGISTRY}/openstack-cli:${RELEASE}
+
+# image:
+#   BUILD --platform linux/amd64 --platform linux/arm64 +platform-image
+
```
mnaser, line 27: FWIW, I think we can skip multiarch for this since I don't think we support an `aarch64` control plane anyways

images/openstack-cli/Earthfile
mnaser: (Follow-up thought, what do you think of installing the CLI from source instead of packages?)

mnaser (changes requested):

images/openstack-cli/Earthfile
fitbeard: I like this idea better. Will do.

PR #972, opened 2024-02-15
feat: containerize openstack cli

fitbeard:
Related: https://github.com/vexxhost/atmosphere/issues/964


fitbeard:
PR is ready for review.

OS clients are pinned to highest versions for 2023.2
To pass all tests we need a real tag for `ghcr.io/vexxhost/atmosphere/openstack-cli` image

After this I will start `nerdctl` in `vexxhost.containers`

mnaser:
Thanks, I made a small changes too to simplify the build and use the same base we're using everywhere

ricolin:
This is a really awesome pr!
I think what's missed are the following changes from ansible-collection-kubernetes repo:
1. https://github.com/vexxhost/ansible-collection-kubernetes/pull/104
2. https://github.com/vexxhost/ansible-collection-kubernetes/pull/100
Once these merged (in that order), we can bump ansible-collection-kubernetes in galaxy.yml to 1.12.2 and unblock your pr.
I will work on making those changes in ansible-collection-kubernetes land asap

mnaser:
I merged the changes for the new format of things and this should landable soon, sorry for the delay, I'll get it in this weekend.

mnaser:
It seems the Octavia command relies on running some commands through the CLI which doesn't seem to be caught, I think we'll have to switch to using shell instead of command. I'll make the swap tomorrow if you haven't caught it yet.

mnaser:
It looks like we are using `shell` for that call, I don't know why it's failing when we have other references to it...

images/openstack-venv-builder/Dockerfile
```diff
@@ -19,6 +19,7 @@ ADD https://releases.openstack.org/constraints/upper/master /upper-constraints.t
 RUN <<EOF sh -xe
 sed -i '/glance-store/d' /upper-constraints.txt
 sed -i '/horizon/d' /upper-constraints.txt
+sed -i '/python-openstackclient/d' /upper-constraints.txt
```
mnaser, line 22: I think this is dangerous here because then it will fully uncap `python-openstackclient` which will break other things, it's probably best we just don't install from source for clients (for now).

mnaser (changes requested):
I think at this point we've got to figure out how to deal with services that call the OpenStack CLI since the aliases are not loaded when `command` or `shell` seems to be running.

Also once we figured that out I'll have a follow up to fix the installation of latest version from pip instead of latest version from source (to avoid deleting from requirements like listed arlier).

images/openstack-venv-builder/Dockerfile
fitbeard: @mnaser the only way I was able to get it to work is to source related aliases:

```
  tasks:
    - name: Test
      changed_when: false
      ansible.builtin.shell: |
        source /etc/profile.d/atmosphere.sh 
        openstack port list
      environment:
        OS_CLOUD: atmosphere
```
The newest (2.2.0) openstack.cloud collection like openstacksdk >= 1.0.0 also lacks this feature.

mnaser: @fitbeard Let's source it for now, it's not the prettiest but that whole thing is a hack, and then ideally we can get rid of the shell/command calls later.

fitbeard:
@mnaser 
From bash documentation: https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html#Bash-Startup-Files
```
Invoked with name sh

If Bash is invoked with the name sh, it tries to mimic the startup behavior of historical versions of sh as closely as possible, while conforming to the POSIX standard as well. 
```
In EL-like distros `/bin/sh` points to `/bin/bash` and this magic happens automatically. This is why i'm testing in a hybrid env with multi distros in the same cluster. But this time i missed this part because cli role is targeting only control nodes. And all control nodes are Rocky based in my case.

fitbeard:
Tests failed because of other reasons

fitbeard:
recheck

fitbeard:
recheck

mnaser (dismissed):
all good at this point.  once this passes, I'll merge.

PR #976, opened 2024-02-19
feat(horizon): Switch to PyMemcacheCache

### LINKED ISSUE #924: Switch to `PyMemcacheCache` for Horizon
https://docs.djangoproject.com/en/3.2/topics/cache/#memcached

> The `MemcachedCache` backend is deprecated as `python-memcached` has some problems and seems to be unmaintained. Use `PyMemcacheCache` or `PyLibMCCache` instead.

We should make this switch, it's should be a pretty easy straight-forward swap.

https://opendev.org/openstack/horizon/commit/caa99b2010bd914bbe3cb2779940b7c544fe1104

Horizon swapped this 3 months ago as well, so we know it's tested, we even had issues with `python-memcached` reported in #861

gtirloni:
Resolves #924

### DIFF
--- a/charts/horizon/values.yaml
+++ b/charts/horizon/values.yaml
@@ -385,7 +385,7 @@ conf:
 
         CACHES = {
             'default': {
-                'BACKEND': 'django.core.cache.backends.memcached.MemcachedCache',
+                'BACKEND': 'django.core.cache.backends.memcached.PyMemcacheCache',
                 'LOCATION': '{{ tuple "oslo_cache" "internal" "memcache" . | include "helm-toolkit.endpoints.host_and_port_endpoint_uri_lookup" }}',
             }
         }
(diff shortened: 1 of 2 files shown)


charts/horizon/values.yaml
```diff
@@ -385,7 +385,7 @@ conf:
 
         CACHES = {
             'default': {
-                'BACKEND': 'django.core.cache.backends.memcached.MemcachedCache',
+                'BACKEND': 'django.core.cache.backends.memcached.PyMemcacheCache',
```
ricolin, line 388: Consider long term maintaince, this change charts should go to openstack-helm like [1] 
And update [2]. Also If base version need to be bump, bump [3] before do [2].
After this, you should be able to run `hack/sync-charts.sh` and merge the change in after you clean up other parts.


Another way to do this without these change is to update `roles/horizon/vars/main.yml` and override the values.

[1] https://review.opendev.org/c/openstack/openstack-helm/+/899809
[2] https://github.com/vexxhost/atmosphere/blob/69ca2857cf5dea8863a91c87ba1fb37bdfa23e95/hack/sync-charts.sh#L79
[3] https://github.com/vexxhost/atmosphere/blob/69ca2857cf5dea8863a91c87ba1fb37bdfa23e95/vendir.yml#L146

mnaser:
Replaced by https://github.com/vexxhost/atmosphere/pull/1090

charts/horizon/values.yaml
gtirloni: Thanks, Rico. It seems the initial request was to do the change in our own code but now I understand I should have submitted it upstream.

PR #977, opened 2024-02-19
feat: allow prefix to image names to support proxy/caching without needing to maintain separate inventory override of each and every image

runlevel-six:
We have been maintaining in our inventory an override of each and every Atmosphere container image just so it can be proxied (and thus cached) through our image repository (Harbor). With very few exceptions, we do not need or desire to maintain pinned image tags, and would prefer to stay in-line with upstream. As such, I am introducing a `atmosphere_proxy_cache_prefix` variable into the `defaults` role with a default value of `""` that can be used as an override to prefix the image value with the value of the caching repository.

### DIFF
--- a/roles/defaults/defaults/main.yml
+++ b/roles/defaults/defaults/main.yml
@@ -17,6 +17,7 @@ atmosphere_version: 1.10.4 # x-release-please-version
 # Ingress
 atmosphere_ingress_class_name: atmosphere
 atmosphere_ingress_cluster_issuer: atmosphere
+atmosphere_proxy_cache_prefix: ""
 
 # Network backend
 atmosphere_network_backend: openvswitch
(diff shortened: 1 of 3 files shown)


mnaser:
It seems the unit tests are failing for this because we're actually consuming these.  I think we might need to use `go-jinja` in the unit tests somehow to make this work.

mnaser:
Sorry I've dropped the ball on this but I think we're going to need this more now that we're relying on our mirrors for CI to speed things up and reduce CI.

I'll pick this up.

runlevel-six:
@mnaser is this still an option? A lack of support for a proxy prefix is preventing us from using the default images from Atmosphosphere in our deployments - we've had to always override all of them to ensure our local Harbor cached them.

mnaser:
merged already

Issue #727, opened 2023-11-23
The name of keystone_keycloak_realm_name was not set into Horizon


mpiscaer:
When I change keystone_keycloak_realm_name it never sets it to Horizon. it only uses the undefined value.

When I set 

```YAML
horizon_helm_values:
  conf:
    horizon:
      local_settings:
          auth:
            idp_mapping: "{{ keystone_domains | default([{'name': 'atmosphere', 'label': 'MyOwn SSO'}]) | vexxhost.atmosphere.keystone_domains_to_idp_mappings }}" # noqa: yaml[line-length]
```

I get the correct name.

mpiscaer:
Even when I set keystone_domains this does not seems to work.

PR #992, opened 2024-02-27
fix: correct keystone_domains reference from horizon

ricolin:
fix https://github.com/vexxhost/atmosphere/issues/727


mnaser:
@ricolin please revert this, we don't want to merge things directly without Zuul first and I don't agree with the approach we've taken here.

PR #1001, opened 2024-03-01
chore: tunning cinder encrypted volume create

ricolin:
ref https://review.opendev.org/c/openstack/cinder/+/910700


mnaser:
@ricolin looks good overall but I don't think smoke tests include Barbican. 

Did you validate the patch inside DevStack for example or in a test environment locally?

ricolin:
> @ricolin looks good overall but I don't think smoke tests include Barbican.
> 
> Did you validate the patch inside DevStack for example or in a test environment locally?

Yeah, validated encrypted volume with both devstack and a local atmosphere aio env

ricolin:
relate to https://github.com/vexxhost/atmosphere/issues/1016

mnaser:
We'll backport this for customers and then just wait for the backport to land upstream

PR #1024, opened 2024-03-18
fix: fix CI with ovn-kubernetes images

ricolin:


ricolin:
this is targeted in https://github.com/vexxhost/atmosphere/pull/1028

PR #1028, opened 2024-03-19
fix: fix CI fails

ricolin:
fix https://github.com/vexxhost/atmosphere/issues/1033


okozachenko1203:
@ricolin  i can see you did a fix for ovn image build.
Btw, i think this will happen in the future again because we use head commit of upstream ovn-kubernetes project.
I proposed a permanent fix for this https://github.com/vexxhost/atmosphere/pull/1030

okozachenko1203:
@ricolin  https://github.com/vexxhost/atmosphere/issues/1033
I opened a new ticket for ovn latest change. Could you pin the commit hash?
Thanks.

Jenkinsfile
```diff
@@ -142,15 +142,36 @@ pipeline {
...
 
         stages {
+          stage('keycloak') {
+            when { expression { env.SCENARIO == "keycloak" } }
+            steps {
+              // Checkout code with built/pinned images
+              unstash 'src-with-pinned-images'
+
+              // Install dependencies
+              sh 'sudo apt-get install -y git python3-pip docker.io'
```
ricolin, line 160: not sure if we have better way to introduce docker to this job

ricolin:
TODO: I will take follow up task to move keycloak job out of container like aio, so things can go faster and we can use current fetch logs script directly to get logs archieve

ricolin:
replace with https://github.com/vexxhost/atmosphere/pull/1049

Issue #1038, opened 2024-03-27
add openstack_database_exporter support 


ricolin:
https://github.com/vexxhost/openstack_database_exporter

PR #1039, opened 2024-03-27
feat: Add openstack db exporter

ricolin:
fix https://github.com/vexxhost/atmosphere/issues/1038

Depends-On: https://github.com/vexxhost/atmosphere/pull/1050

### DIFF
--- a/roles/kube_prometheus_stack/files/jsonnet/openstack.libsonnet
+++ b/roles/kube_prometheus_stack/files/jsonnet/openstack.libsonnet
@@ -129,6 +129,18 @@
                 severity: 'warning',
               },
             },
+            {
+              alert: 'NeutronMultipleActiveL3AgentForRouter',
+              annotations: {
+                description: '[{{ $labels.router_id }}] Contains multiple active L3 agents.',
+                summary: '[{{ $labels.router_id }}] Contains multiple active L3 agents.',
+              },
+              expr: 'sum by (router_id) (openstack_neutron_l3_agent_of_router{ha_state="active"}) > 1',
+              'for': '5m',
+              labels: {
+                severity: 'P3',
+              },
+            },
           ],
       },
       {
--- a/roles/kube_prometheus_stack/files/jsonnet/openstack.libsonnet
+++ b/roles/kube_prometheus_stack/files/jsonnet/openstack.libsonnet
@@ -130,10 +130,10 @@
               },
             },
             {
-              alert: 'NeutronMultipleActiveL3AgentForRouter',
+              alert: 'NeutronRouterMultipleActiveL3Agents',
               annotations: {
-                description: '[{{ $labels.router_id }}] Contains multiple active L3 agents.',
-                summary: '[{{ $labels.router_id }}] Contains multiple active L3 agents.',
+                summary: 'Neutron HA router has multiple active L3 agents',
+                description: 'The router with ID {{ $labels.router_id }} has {{ $value }} L3 agents in active state which can cause network resets and traffic drops.',
               },
               expr: 'sum by (router_id) (openstack_neutron_l3_agent_of_router{ha_state="active"}) > 1',
               'for': '5m',
--- a/roles/kube_prometheus_stack/files/jsonnet/openstack.libsonnet
+++ b/roles/kube_prometheus_stack/files/jsonnet/openstack.libsonnet
@@ -241,6 +241,24 @@
           },
         ],
       },
+      {
+        name: 'octavia',
+        rules:
+          [
+            {
+              alert: 'OctaviaLoadBalancerNotActive',
+              annotations: {
+                summary: 'Octavia load balancer not active',
+                description: 'Load balancer with ID {{ $labels.id }} stuck in non-active state for more then 15 minutes.',
+              },
+              expr: 'openstack_loadbalancer_loadbalancer_status{provisioning_status!="ACTIVE"} > 0',
+              'for': '15m',
+              labels: {
+                severity: 'P3',
+              },
+            },
+          ],
+      },
     ],
   },
 }
--- a/roles/kube_prometheus_stack/vars/main.yml
+++ b/roles/kube_prometheus_stack/vars/main.yml
@@ -439,6 +439,18 @@ _kube_prometheus_stack_helm_values:
           - interval: 60s
             port: metrics
             relabelings: *relabelings_instance_to_node_name
+      - name: openstack-database-exporter
+        jobLabel: job
+        namespaceSelector:
+          matchNames:
+            - openstack
+        selector:
+          matchLabels:
+            application: openstack-database-exporter
+        podMetricsEndpoints:
+          - interval: 60s
+            port: metrics
+            relabelings: *relabelings_instance_to_pod_name
       - name: percona-xtradb-pxc
         jobLabel: app.kubernetes.io/component
         namespaceSelector:
--- a/roles/openstack_exporter/meta/main.yml
+++ b/roles/openstack_exporter/meta/main.yml
@@ -30,5 +30,4 @@ galaxy_info:
 
 dependencies:
   - role: defaults
-  - role: openstack_helm_endpoints
-    openstack_helm_endpoints_list: ["oslo_db"]
+  - role: secretgen_controller
--- a/roles/openstack_exporter/meta/main.yml
+++ b/roles/openstack_exporter/meta/main.yml
@@ -30,4 +30,3 @@ galaxy_info:
 
 dependencies:
   - role: defaults
-  - role: secretgen_controller
--- a/roles/openstack_exporter/meta/main.yml
+++ b/roles/openstack_exporter/meta/main.yml
@@ -30,3 +30,5 @@ galaxy_info:
 
 dependencies:
   - role: defaults
+  - role: openstack_helm_endpoints
+    openstack_helm_endpoints_list: ["oslo_db"]
--- a/roles/openstack_exporter/tasks/main.yml
+++ b/roles/openstack_exporter/tasks/main.yml
@@ -114,3 +114,51 @@
               targetPort: metrics
           selector:
             application: openstack-exporter
+
+- name: Deploy service
+  kubernetes.core.k8s:
+    state: present
+    definition:
+      - apiVersion: apps/v1
+        kind: Deployment
+        metadata:
+          name: openstack-database-exporter
+          namespace: openstack
+          labels:
+            application: openstack-database-exporter
+        spec:
+          replicas: 1
+          selector:
+            matchLabels:
+              application: openstack-database-exporter
+          template:
+            metadata:
+              labels:
+                application: openstack-database-exporter
+            spec:
+              nodeSelector:
+                openstack-control-plane: enabled
+              containers:
+                - name: openstack-database-exporter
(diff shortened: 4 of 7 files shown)


roles/openstack_exporter/tasks/main.yml
```diff
@@ -114,3 +114,104 @@
...
+                          interface: "$(OS_INTERFACE)"
+                          identity_api_version: 3
+                          identity_interface: "$(OS_INTERFACE)"
+                      EOF
+                  envFrom:
+                    - secretRef:
+                        name: keystone-keystone-admin
+                  volumeMounts:
+                    - name: openstack-config
+                      mountPath: /etc/openstack
```
mnaser, line 168: This init container is useless.

roles/openstack_exporter/tasks/main.yml
```diff
@@ -114,3 +114,104 @@
...
+                  image: "{{ atmosphere_images['prometheus_openstack_database_exporter'] | vexxhost.kubernetes.docker_image('ref') }}"
+                  args:
+                    - --neutron.dsn
+                    - "neutron:{{ openstack_helm_endpoints.oslo_db.auth.neutron.password }}@tcp(percona-xtradb-haproxy.openstack.svc.cluster.local:3306)/neutron"
+                  ports:
+                    - name: metrics
+                      containerPort: 9180
+                  env:
+                    - name: OS_COMPUTE_API_VERSION
+                      value: "2.87"
```
mnaser, line 180: useless too

roles/openstack_exporter/tasks/main.yml
```diff
@@ -114,3 +114,104 @@
...
+                  volumeMounts:
+                    - name: openstack-config
+                      mountPath: /etc/openstack
```
mnaser, line 183: useless too

roles/openstack_exporter/tasks/main.yml
```diff
@@ -114,3 +114,104 @@
...
+                    httpGet:
+                      path: /
+                      port: 9180
+                      scheme: HTTP
+                    periodSeconds: 10
+                    successThreshold: 1
+                    timeoutSeconds: 1
+              volumes:
+                - name: openstack-config
+                  emptyDir: {}
```
mnaser, line 201: useless too

roles/openstack_exporter/tasks/main.yml
```diff
@@ -114,3 +114,104 @@
...
+          labels:
+            application: openstack-database-exporter
+        spec:
+          clusterIP: None
+          ports:
+            - name: metrics
+              port: 9180
+              targetPort: metrics
+          selector:
+            application: openstack-database-exporter
```
mnaser, line 217: let's remove this and switch to using a podmonitor

roles/kube_prometheus_stack/vars/main.yml
```diff
@@ -402,6 +402,23 @@ _kube_prometheus_stack_helm_values:
...
+            application: openstack-database-exporter
+        endpoints:
+          - port: metrics
+            interval: 1m
+            relabelings:
+              - action: replace
+                regex: (.*)
+                replacement: default
+                targetLabel: instance
+            scrapeTimeout: 30s
```
mnaser, line 421: Let's use a podmonitor for this to avoid the creating a servic ewe don't need.

mnaser (changes requested):

ricolin:
quary result looks like this from grafana
image

roles/openstack_exporter/tasks/main.yml
```diff
@@ -114,3 +114,51 @@
...
+                application: openstack-database-exporter
+            spec:
+              nodeSelector:
+                openstack-control-plane: enabled
+              containers:
+                - name: openstack-database-exporter
+                  image: "{{ atmosphere_images['prometheus_openstack_database_exporter'] | vexxhost.kubernetes.docker_image('ref') }}"
+                  args:
+                    - --neutron.dsn
+                    - "root:{{ openstack_helm_endpoints.oslo_db.auth.admin.password }}@tcp(percona-xtradb-haproxy.openstack.svc.cluster.local:3306)/neutron"  # yamllint disable-line rule:line-length
```
mnaser, line 146: I don't really like this, the concern I have with this is that this is a security issue because the password is exposed in the command line.

I'll have to rewrite this to rely on environment variables instead.

roles/kube_prometheus_stack/files/jsonnet/openstack.libsonnet
```diff
@@ -129,6 +129,18 @@
                 severity: 'warning',
               },
             },
+            {
+              alert: 'NeutronMultipleActiveL3AgentForRouter',
+              annotations: {
+                description: '[{{ $labels.router_id }}] Contains multiple active L3 agents.',
+                summary: '[{{ $labels.router_id }}] Contains multiple active L3 agents.',
```
mnaser, line 136:
```suggestion
                summary: 'Neutron HA router has multiple active L3 agents',
                description: 'The router with ID {{ $labels.router_id }} has {{ $value }} L3 agents in active state which can cause network resets and traffic drops.',
```

roles/kube_prometheus_stack/files/jsonnet/openstack.libsonnet
```diff
@@ -129,6 +129,18 @@
                 severity: 'warning',
               },
             },
+            {
+              alert: 'NeutronMultipleActiveL3AgentForRouter',
```
mnaser, line 133:
```suggestion
              alert: 'NeutronRouterMultipleActiveL3Agents',
```

mnaser (changes requested):
Let me get the exporter fixes in.

roles/openstack_exporter/tasks/main.yml
mnaser: The new version now can support environment variables to do this, so this should largely be solved.

However, in order to be more neat here, I think what you can do is add the `secretgen-controller` as a dependency to this.

https://github.com/carvel-dev/secretgen-controller/blob/develop/docs/secret-template.md

We can leverage the existing secrets to build the secret so we don't have to run the openstack helm endpoints role, we can just rely on the secrets which are stored in Kubernetes.  It'll be more smart in that sense (and more dynamic).

Once we have that secret, we can just use `envFrom` to get it from that secret.

mnaser (changes requested):
The one last thing to keep in mind that I forgot was that now with both the `openstack-exporter` and `openstack_database_exporter` running side by side, the data will be duplicated inside Prometheus.

We can disable certain metrics from being logged with the openstack-exporter, so can you make sure you disable the ones that we are scraping in the database exporter (so far 2)... and we'll keep in mind to keep removing more as we go :)

roles/kube_prometheus_stack/files/jsonnet/openstack.libsonnet
```diff
@@ -129,6 +129,18 @@
...
+                description: '[{{ $labels.router_id }}] Contains multiple active L3 agents.',
+                summary: '[{{ $labels.router_id }}] Contains multiple active L3 agents.',
+              },
+              expr: 'sum by (router_id) (openstack_neutron_l3_agent_of_router{ha_state="active"}) > 1',
+              'for': '5m',
+              labels: {
+                severity: 'P3',
+              },
+            },
           ],
```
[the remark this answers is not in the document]
mnaser: FYI, I'm thinking anything where the provisioning_status of the load balancer != ACTIVE for more than 15 minutes is probably something we should alarm on.

roles/openstack_exporter/tasks/main.yml
ricolin: I been searching for a while, but can't find a way to reform secret from 
` mysql+pymysql://neutron:07sj1sB0gjaV4Je5mBIMBpMAewRL4IMT@percona-xtradb-haproxy.openstack.svc.cluster.local:3306/neutron `
to
` neutron:07sj1sB0gjaV4Je5mBIMBpMAewRL4IMT@tcp(percona-xtradb-haproxy.openstack.svc.cluster.local:3306)/neutron `
with SecretTemplate.
I currently using k8s_info to fetch secret, reform it and store back as another secret.

roles/kube_prometheus_stack/files/jsonnet/openstack.libsonnet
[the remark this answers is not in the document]
ricolin: I added one with P3 in this case

roles/openstack_exporter/meta/main.yml
```diff
@@ -30,3 +30,4 @@ galaxy_info:
 
 dependencies:
   - role: defaults
+  - role: secretgen_controller
```
mnaser, line 33: This is unused

roles/openstack_exporter/tasks/main.yml
```diff
@@ -114,3 +116,93 @@
...
+- name: Fetch Octavia DB secret
+  run_once: true
+  no_log: true
+  kubernetes.core.k8s_info:
+    kind: Secret
+    namespace: openstack
+    name: octavia-db-user
+    wait: true
+    wait_sleep: 1
+    wait_timeout: 600
```
mnaser, line 129: This is not necessary.

roles/openstack_exporter/tasks/main.yml
```diff
@@ -114,3 +116,93 @@
...
+- name: Fetch Neutron DB secret
+  run_once: true
+  no_log: true
+  kubernetes.core.k8s_info:
+    kind: Secret
+    namespace: openstack
+    name: neutron-db-user
+    wait: true
+    wait_sleep: 1
+    wait_timeout: 600
```
mnaser, line 141: We can drop this too.

roles/openstack_exporter/tasks/main.yml
```diff
@@ -114,3 +116,93 @@
...
+
+- name: Create "openstack-database-exporter-dsn" secret
+  run_once: true
+  no_log: true
+  kubernetes.core.k8s:
+    state: present
+    definition:
+      apiVersion: v1
+      kind: Secret
+      type: kubernetes.io/rbd
```
mnaser, line 152: Definitely not rbd type secret :)

roles/openstack_exporter/tasks/main.yml
```diff
@@ -114,3 +116,93 @@
...
+        name: openstack-database-exporter-dsn
+        namespace: openstack
+        labels:
+          application: openstack-database-exporter
+      stringData:
+        NEUTRON_DSN: "{{ _neutron_db_user.resources.0.data.DB_CONNECTION | b64decode |  regex_replace('^.*//', '') | regex_replace('@(.*)/', '@tcp(\\1)/') }}"
+        OCTAVIA_DSN: "{{ _octavia_db_user.resources.0.data.DB_CONNECTION | b64decode |  regex_replace('^.*//', '') | regex_replace('@(.*)/', '@tcp(\\1)/') }}"
+
+- name: Deploy service
+  kubernetes.core.k8s:
```
mnaser, line 163: You can probably run this task once too.

mnaser (changes requested):

mnaser:
recheck

mnaser:
recheck

mnaser (approved):

Issue #1040, opened 2024-03-28
OVN NB and SB pods are not using the PVC for their databases


thywyn:
#OVN NB sts config:
$ kubectl get sts -n openstack ovn-ovsdb-nb -o yaml

``` yaml
...
spec:
...
  template:
...
    containers:
      - command:
        - /tmp/ovsdb-server.sh
        - start
        env:
        - name: OVS_DATABASE
          value: nb
        - name: OVS_PORT
          value: "6641"
...
        volumeMounts:
        - mountPath: /tmp/ovsdb-server.sh
          name: ovn-bin
          readOnly: true
          subPath: ovsdb-server.sh
        - mountPath: /run/openvswitch
          name: run-openvswitch
        - mountPath: /var/lib/ovn
          name: data
...
  volumeClaimTemplates:
  - apiVersion: v1
    kind: PersistentVolumeClaim
    metadata:
      creationTimestamp: null
      name: data
    spec:
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 20Gi
      storageClassName: csi-rbd-sc
      volumeMode: Filesystem
...
```
#No Database files
$ kubectl exec -it -n openstack ovn-ovsdb-nb-1 -- ls -al /var/lib/ovn
Defaulted container "ovsdb" out of: ovsdb, init (init)
total 24
drwxr-xr-x 3 root root  4096 Mar 25 23:51 .
drwxr-xr-x 1 root root  4096 Mar 27 14:39 ..
drwx------ 2 root root 16384 Mar 25 23:51 lost+found

$ kubectl exec -it -n openstack ovn-ovsdb-nb-1 -- df -k /var/lib/ovn
Defaulted container "ovsdb" out of: ovsdb, init (init)
Filesystem     1K-blocks  Used Available Use% Mounted on
/dev/rbd7       20466256    24  20449848   1% /var/lib/ovn

#Database files seem to be in the default location
$ kubectl exec -it -n openstack ovn-ovsdb-nb-1 -- ls -al /etc/ovn
Defaulted container "ovsdb" out of: ovsdb, init (init)
total 136
drwxr-xr-x 2 root root   4096 Mar 28 14:40 .
drwxr-xr-x 1 root root   4096 Mar 27 14:39 ..
-rw------- 1 root root      0 Mar 27 14:39 .ovnnb_db.db.~lock~
-rw-r----- 1 root root 127050 Mar 28 17:33 ovnnb_db.db

[...]

PR #1044, opened 2024-04-01
fix: fix ovn db mount path

okozachenko1203:
fix https://github.com/vexxhost/atmosphere/issues/1040

https://review.opendev.org/c/openstack/openstack-helm-infra/+/914807

Depends-On: https://github.com/vexxhost/atmosphere/pull/1043


mnaser:
recheck

mnaser:
@okozachenko1203 seems like this passes the normal OVN test (but then the existing setup always did).

Did you get a chance to confirm if this does fix it?

Issue #1046, opened 2024-04-01
Issues when creating images from volumes


guilhermesteinmuller:
Right now, when users tries to create images from volumes, we get the following issue:

```
2024-04-01 20:16:41.948 7 ERROR glance_store._drivers.rbd [None req-01251fe4-97c9-4772-a5bf-d47bbd0a647f 5e7c9e9fa20c4d7e9e1395eb8f3808e0 d8f09b18703f47eb8199a17b85ce0ce9 - - default default] Failed to store image d0c5f648-bcdd-47ad-971a-2888584d11f5 Store Exception unable to receive chunked part: OSError: unable to receive chunked part
2024-04-01 20:16:42.298 7 ERROR glance.api.v2.image_data [None req-01251fe4-97c9-4772-a5bf-d47bbd0a647f 5e7c9e9fa20c4d7e9e1395eb8f3808e0 d8f09b18703f47eb8199a17b85ce0ce9 - - default default] Failed to upload image data due to internal error: OSError: unable to receive chunked part
2024-04-01 20:16:42.323 7 ERROR glance.common.wsgi [None req-01251fe4-97c9-4772-a5bf-d47bbd0a647f 5e7c9e9fa20c4d7e9e1395eb8f3808e0 d8f09b18703f47eb8199a17b85ce0ce9 - - default default] Caught error: unable to receive chunked part: OSError: unable to receive chunked part
2024-04-01 20:16:42.323 7 ERROR glance.common.wsgi Traceback (most recent call last):
2024-04-01 20:16:42.323 7 ERROR glance.common.wsgi   File "/var/lib/openstack/lib/python3.10/site-packages/glance/common/wsgi.py", line 1295, in __call__
2024-04-01 20:16:42.323 7 ERROR glance.common.wsgi     action_result = self.dispatch(self.controller, action,
2024-04-01 20:16:42.323 7 ERROR glance.common.wsgi   File "/var/lib/openstack/lib/python3.10/site-packages/glance/common/wsgi.py", line 1338, in dispatch
2024-04-01 20:16:42.323 7 ERROR glance.common.wsgi     return method(*args, **kwargs)
2024-04-01 20:16:42.323 7 ERROR glance.common.wsgi   File "/var/lib/openstack/lib/python3.10/site-packages/glance/common/utils.py", line 415, in wrapped
2024-04-01 20:16:42.323 7 ERROR glance.common.wsgi     return func(self, req, *args, **kwargs)
[...]

PR #1047, opened 2024-04-02
fix(uwsgi): tune settings

mnaser:
The current uWSGI settings that we ship are missing a few important
ones, specifically `http-raw-body` which is used by Glance in order
to enable chunked uploads.

This is the reason why a direct upload to Glance works just fine
because it goes through the ingress (which does the chunking for
us), however, it fails when creating an image from volume due to
the fact that Cinder talks directly to Glance without the ingress
being in the path.

Closes: #1046

### DIFF
--- a/charts/cinder/values.yaml
+++ b/charts/cinder/values.yaml
@@ -1015,15 +1015,20 @@ conf:
     uwsgi:
       add-header: "Connection: close"
       buffer-size: 65535
+      chunked-input-limit: 4096000
       die-on-term: true
       enable-threads: true
       exit-on-reload: false
       hook-master-start: unix_signal:15 gracefully_kill_them_all
+      http-auto-chunked: true
+      http-raw-body: true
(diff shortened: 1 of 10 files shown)


mnaser:
recheck

guilhermesteinmuller:
It looks like the changes above don't take any effect. Even adding parameters exactly like[ here](https://raw.githubusercontent.com/openstack/glance/master/httpd/glance-api-uwsgi.ini) we still experience the `OSError: unable to receive chunked part`

gtirloni:
I was able to create the image from a volume after setting `glance_api_uwsgi.uwsgi.chunked-input-limit` to `4096000` in `charts/glance/values.yaml` (default: 1M).

`volume-cinder` sends the contents of the image over to `glance-api` and I suspect the RBD driver's `rbd_store_chunk_size` being set to `4` (megabytes) might have something to do with this.

PR #1049, opened 2024-04-02
Fix image ci

ricolin:
fix https://github.com/vexxhost/atmosphere/issues/1033


mnaser:
@ricolin Can you refresh this to the latest way of doing things for the Docker images?

PR #1050, opened 2024-04-02
ci(ovn): check if ovndb is modified for 5 minutes

mnaser:
We would make sure that the last time the database was modified was
a minute ago but maybe in a quiet cloud that's too much, so we bump
it up to 5 minutes.


mnaser:
recheck

mnaser:
recheck

PR #1052, opened 2024-04-03
fix(defaults): Update sandbox image to pause:3.9

gtirloni:
Fix for:

```
W0403 00:49:34.034468   70448 checks.go:835] detected that the sandbox image "registry.k8s.io/pause:3.8@sha256:f5944f2d1daf66463768a1503d0c8c5e8dde7c1674d3f85abc70cef9c7e32e95" of the container runtime is inconsistent with that used by kubeadm. It is recommended that using "registry.k8s.io/pause:3.9" as the CRI sandbox image.
error execution phase wait-control-plane: couldn't initialize a Kubernetes cluster
To see the stack trace of this error execute with --v=5 or higher
```


mnaser:
recheck

gtirloni:
@mnaser I think I only saw this warning due to another unrelated issue in kubeadm while bootstrapping a new cluster (which made stderr get shown). It doesn't seem to cause a failure since it's just a warning and Atmosphere pins the images anyway.

According to [this](https://github.com/kubernetes/kubeadm/issues/2680), kubeadm doesn't support custom digests and thus the only "fix" would be to remove the sha256 and only use "pause:3.9", which I think is not desirable here.

mnaser:
@gtirloni I still think it would be good to use the newer 3.9 image since that's what k8s recommends!

gtirloni:
@mnaser makes sense! undrafted.

PR #1061, opened 2024-04-03
fix(glance_image): introduce retries

mnaser:
The internet is not reliable.  We instead introduce 3 retries for
downloading images, with a random delay of at least 3 seconds to
make sure we can retry safely.

Depends-On: https://github.com/vexxhost/atmosphere/pull/1063


mnaser:
recheck

PR #1071, opened 2024-04-08
feat: Upgrade monitoring stack

gtirloni:
Loki 2.7.3 wasn't playing well with some queries from Grafana 10.3 and updating it to 2.9.6 fixed the issues.

While at it, I've updated Vector and Grafana/Prometheus as well.

Tested it in a live environment and also AIO. No new errors noticed in the logs. 

Vector is a bit more noisy about issues but it doesn't prevent logs from being forwarded (especially, it complains when a log event is missing expected labels.. whereas before it would just ignore it).


mnaser:
This is awesome, but I'd like this to be a test on how we do cherry-picks, since I'd like to have this in `stable/zed` too since monitoring stack updates are non-user affecting.

mnaser:
regate

PR #1090, opened 2024-04-12
Build "master" images in "main"

mnaser:


mnaser:
It seems like there's an issue with the Horizon image, @okozachenko1203 digging in.

mnaser:
recheck

Issue #42, opened 2022-09-07
Switch all services to use UWSGI


mnaser:
At the moment, the services are running the (very) inefficient version of their eventlet API endpoint.

We should flip over to uWSGI (or something using `mod_wsgi`) in order to be able to have a proper HTTP server that can scale to the demand.

mnaser:
This has become more of an issue, with more recent and larger clouds hitting issues.

- Nova has [deprecated services running under eventlet](https://opendev.org/openstack/nova/commit/b53d81b03cad73fac7f558d287db0354f0a46ec1)
- DevStack has long moved towards uWSGI

In terms of comparing other deployments tools:

- OpenStack-Ansible: uWSGI
- Kolla: mod_wsgi
- TripleO: mod_wsgi

So the favour seems to be towards mod_wsgi, another neat thing about mod_wsgi is that Horizon and Keystone technically use it (and Keystone requires it because of the SSO things)... so I think we have to move towards `mod_wsgi` I believe...

mnaser:
nova:

```
[uwsgi]
chmod-socket = 666
socket = /var/run/uwsgi/nova-api-wsgi.socket
lazy-apps = true
add-header = Connection: close
buffer-size = 65535
hook-master-start = unix_signal:15 gracefully_kill_them_all
thunder-lock = true
plugins = http,python3
enable-threads = true
worker-reload-mercy = 80
exit-on-reload = false
die-on-term = true
master = true
processes = 8
wsgi-file = /opt/stack/data/venv/bin/nova-api-wsgi
```

mnaser:
I started a very simple layout here:

https://github.com/vexxhost/atmosphere/pull/596

If this works, I'll merge it for now since we actually need to land support for Nova due to some issues, but we should do this for all other services:

- Build latest images so uWSGI with `pcre` support is included
- Pin latest images with PCRE support
- Follow same format as other PR to add support for other services

It should be pretty straight forward and just a process of going over all of them one by one.

mnaser:
I believe Neutron does ship with a WSGI.  Also, I think we missed Barbican.

ricolin:
tracking community patch: https://review.opendev.org/q/topic:%22use-uwsgi%22

ricolin:
a relate issue regarding neutron-server+uwsgi https://bugs.launchpad.net/neutron/+bug/2062009

ricolin:
[...]

PR #1119, opened 2024-04-16
fix: Barbican use same uwsgi config as other services.

ricolin:
This renew Barbican uwsgi config and make sure it uses same uwsgi config as other services.

relate to https://github.com/vexxhost/atmosphere/issues/42


mnaser:
@ricolin 

Two issues with this change

1. I have no idea how the CI didn't catch that there wasn't a patch included.. it should have failed because there is a missing patch.
2. We either need to push this upstream and add it as a patch into `.charts.yml` or add it as a local patch.

ricolin:
> @ricolin
> 
> Two issues with this change
> 
> 1. I have no idea how the CI didn't catch that there wasn't a patch included.. it should have failed because there is a missing patch.
> 2. We either need to push this upstream and add it as a patch into .charts.yml or add it as a local patch.

No problem, I was targeting of shift it to upstream once we approve it, but let me link from path now

charts/barbican/charts/helm-toolkit/Chart.yaml
```diff
@@ -9,4 +9,4 @@ name: helm-toolkit
 sources:
 - https://opendev.org/openstack/openstack-helm-infra
 - https://opendev.org/openstack/openstack-helm
-version: 0.2.55
```
mnaser, line 12: Please pin helm-toolkit to 0.2.55 as we've done in the other charts in order to avoid a this big commit change and make it a much smaller change.

mnaser (changes requested):

mnaser (changes requested):
@ricolin :

Can you make a follow-up patch similar to these?

https://github.com/vexxhost/atmosphere/blob/main/charts/patches/cinder/0001-tune-uwsgi-config.patch

mnaser:
@ricolin Since we are actually using backports now and have stable branches, we're trying to avoid bumps on master branch for changes that would be backported (such as this one).

We can have another seperate PR for the bump that we would not backport

mnaser:
recheck

ricolin:
> @ricolin Since we are actually using backports now and have stable branches, we're trying to avoid bumps on master branch for changes that would be backported (such as this one).
> 
> We can have another seperate PR for the bump that we would not backport

I just put them to here now https://github.com/vexxhost/atmosphere/pull/1136

mnaser:
regate

mnaser:
regate

mnaser:
regate

mnaser (approved):

mnaser:
regate

PR #1126, opened 2024-04-18
Change Loki log_level from info to warn

gtirloni:
The default log_level is `info` which causes every query to generate dozens of new log lines with the query's contents which then show up in future results if the query is execute again making it harder to find what you were looking for in the middle of extra noise that has to be filtered out when doing a global search.


mnaser:
regate

mnaser:
regate

mnaser:
regate

PR #1127, opened 2024-04-18
Increase rook-ceph-operator CPU limit to 1

gtirloni:
~~[Default](https://github.com/vexxhost/atmosphere/blob/main/charts/rook-ceph/values.yaml#L29) is 100m which causes a lot of alert flapping even in a tiny idle cluster.~~

Default CPU limit is 500m which causes alert flapping. Increasing to 1 to accommodate peaks.

Tweaking it in the role so it's easier to sync the chart's values.yml with upstream.


mnaser (approved):

mnaser:
regate

mnaser:
regate

mnaser:
regate

PR #1129, opened 2024-04-18
Add monitoring for stuck VMs

mnaser:
If RabbitMQ decides to go for a nap, some messages may be stuck
and it may cause certain VMs to be stuck in certain states such
as deleting or building.

This adds monitoring in order to catch if any of those operations
are stuck in the cloud.


mnaser:
recheck

mnaser:
regate

Issue #1004, opened 2024-03-04
OVN 23.x not supported anymore


gtirloni:
https://mail.openvswitch.org/pipermail/ovs-announce/2024-March/000344.html

```
OVN v24.03.0 is the newest Long Term Support (LTS) version of OVN.

OVN v22.03.6 is the final regularly-scheduled release of the previous 
LTS version of OVN. It now is entering security-maintenance mode, 
meaning that it will only receive security fixes and other 
urgently-required fixes.

With this release, OVN v23.03.0 is now no longer supported and will 
receive no planned future releases
```

https://github.com/vexxhost/atmosphere/blob/main/images/ovn/Earthfile#L4

PR #1147, opened 2024-04-21
Prepare for Antelope branch

mnaser:
This change contains a big bump of versions for many components
but some are not fully bumped to the latest which are:

- Keycloak (in order to go to 23.X first, then B will go to 24.X)
- RabbitMQ (In order to go to 3.11 first, then B will go to 3.12)
- OVN (24.03 LTS packages don't exist yet)

Once we merge this, we should run some upgrade tests from the last
branch (in this case, `stable/zed`) to `main` before we branch
this out to `stable/2023.1`.

Closes: #1004


roles/magnum/vars/main.yml
```diff
@@ -81,7 +81,6 @@ _magnum_helm_values:
 _magnum_registry_ingress_annotations:
   # NOTE(mnaser): We only want to allow GET/HEAD requests to the registry
   #               to make sure it's read-only.
-  nginx.ingress.kubernetes.io/configuration-snippet: |
-    if ($request_method !~* "^(GET|HEAD)$") {
-      return 403;
-    }
```
mnaser, line 87: This isn't the same thing because this is for CORS only, however, we have actually enabled the registry itself to be read only with another change, so you can remove this whole part because in the environment variables you can see how we set things to be read on

roles/loki/tasks/main.yml
```diff
@@ -20,4 +20,4 @@
     release_namespace: "{{ loki_helm_release_namespace }}"
     create_namespace: true
     kubeconfig: /etc/kubernetes/admin.conf
-    values: "{{ _loki_helm_values | combine(loki_helm_values, recursive=True) }}"
+    values: "{{ _loki_helm_values | combine(loki_helm_values, recursive=True) | to_json }}"
```
mnaser, line 23: What is the reason behind the `to_json` here since it seems all other roles don't need it?

mnaser (commented):
@ricolin few comments

roles/loki/tasks/main.yml
ricolin: Add this because loki raise error when deploy helm chart:
``` error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go value of type map[string]interface {} ```

roles/magnum/vars/main.yml
ricolin: okay, removed

ricolin:
recheck

Issue #1016, opened 2024-03-13
Encrypted volume creation tunings


ricolin:
This regards tunning encrypted volume creation for both:

1. empty volume
2. create volume from image

ricolin:
works for `empty volume` task
- https://review.opendev.org/c/openstack/cinder/+/910700
- https://github.com/vexxhost/atmosphere/pull/1001

PR #1149, opened 2024-04-22
[stable/zed] fix: add missing qemu rbd package

ricolin:
relate to https://github.com/vexxhost/atmosphere/issues/1016


ricolin:
recheck

mnaser:
@ricolin we need this in master first then we'll backport to stable/zed.

PR #1160, opened 2024-04-24
feat: Neutron OVS+uwsgi support

ricolin:
relate to https://github.com/vexxhost/atmosphere/issues/42

Note: ovn can't properly use uWSGI for now due to: https://bugs.launchpad.net/neutron/+bug/1912359


ricolin:
ovn+wsgi in neutron seems still not there https://bugs.launchpad.net/neutron/+bug/1912359

ricolin:
reference openstack-helm patch: https://review.opendev.org/c/openstack/openstack-helm/+/916862

ricolin:
recheck

mnaser (changes requested):
I would like us to keep the API to use `neutron-server` instead.  Without this, there will be an outage on the API until the Ingress resource is created which isn't ideal.

ricolin:
> I would like us to keep the API to use `neutron-server` instead. Without this, there will be an outage on the API until the Ingress resource is created which isn't ideal.

refactored!

mnaser (approved):

mnaser:
@ricolin Do you think we should back port this to `stable/zed` or it's okay as is in there?  We've been running it for a long time with no issues so I feel like it might not be super necessary to do that.

ricolin:
> @ricolin Do you think we should back port this to `stable/zed` or it's okay as is in there? We've been running it for a long time with no issues so I feel like it might not be super necessary to do that.

agree, as consider ovn+uwsgi not working, it might not be save to backport such a feature to zed.

PR #1167, opened 2024-04-29
Add Portworx CSI support

mnaser:


mnaser:
recheck

roles/portworx/templates/storage_cluster.yml
```diff
@@ -0,0 +1,53 @@
...
+    enabled: true
+  monitoring:
+    telemetry:
+      enabled: true
+    prometheus:
+      enabled: false
+      exportMetrics: true
+  env:
+  - name: PURE_FLASHARRAY_SAN_TYPE
+    value: "{{ portworx_pure_flasharray_san_type }}"
```
gtirloni, line 44: I think we could make this `portworx_cluster_env` or something along these lines so we don't have to keep track of all the variables that Portworx will ever support. wdyt?

roles/multipathd/tasks/main.yml
```diff
@@ -12,6 +12,18 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Add backports PPA
+  ansible.builtin.apt_repository:
+    repo: ppa:vexxhost/backports
```
gtirloni, line 17: Minor warning from apt:

```
W: http://ppa.launchpad.net/vexxhost/backports/ubuntu/dists/jammy/InRelease: Key is stored in legacy trusted.gpg keyring (/etc/apt/trusted.gpg), see the DEPRECATION section in apt-key(8) for details.
```

I think the workaround is to grab the key directly.. something along these lines:

```
- name: Add backports PPA key
  ansible.builtin.get_url:
    url: http://ppa.launchpad.net/vexxhost/backports/ubuntu/dists/jammy/Release.gpg
    dest: /etc/apt/keyrings/ppa_vexxhost_backports.asc
```

roles/portworx/templates/storage_cluster.yml
mnaser: I feel like the SAN type is a required value, it _has_ to be one or the other, I'm usually in favour of keeping very specific minimal set of configurable variables to make things supportable.

gtirloni: Ah, agreed. If there is no option around it making it required prevents foot shooting.

PR #1175, opened 2024-05-02
Bump nova octavia

ricolin:


mnaser:
@ricolin Can you please add details to the PR about why this change is done?

PR #1177, opened 2024-05-02
fix: use kubelet_hostname instead of inventory_hostname_short

fitbeard:
In `vexxhost.kubernetes.kubelet` `kubelete_hostname` default value is `inventory_hostname_short`.  But it also can be changed to something else. Let's use proper variable here.


mnaser:
regate

Issue #1185, opened 2024-05-04
Openstack CLI with private CA


mpiscaer:
In our dev setup we use Atmosphere with multi nodes. For this we use a ACME cert that uses a private CA.

In the old manner the public key got imported in the keystore via /usr/local/share/ca-certificates/

But with the new Openstack CLI inside a container this CA is not imported into the used keystore.

What is the best way to get this done?

Currently we get the error:
```
root@ctl1:~# openstack network list
SSL exception connecting to https://identity.infra.cluster6.DOMAIN.TLD/v3/auth/tokens: HTTPSConnectionPool(host='identity.infra.cluster6.DOMAIN.TLD', port=443): Max retries exceeded with url: /v3/auth/tokens (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1007)')))
```

mpiscaer:
I see that [alias](https://github.com/vexxhost/atmosphere/blob/main/roles/openstack_cli/templates/atmosphere.sh.j2#L4 ) imports the ca when it is cluster_issuer_type self-signed or ca.

mnaser:
@fitbeard curious if you have any thoughts about this?

fitbeard:
@mpiscaer @fitbeard https://github.com/vexxhost/atmosphere/pull/1197 that should be the missing piece.

@mpiscaer can you test this change please? If this fix the issue with private CA then next step would be fix for https://github.com/vexxhost/atmosphere/issues/1186 using bits from your https://github.com/vexxhost/atmosphere/pull/1188 + mount changes from my PR.

mpiscaer:
@fitbeard The problem with #1197 is that openstack-cli image uses certifi and certifi does not accept private signed ca's. Certifi uses its own keystore.

I try'd to manual uninstall the pip certifi package, but then openstack application complains about missing certifi module.

fitbeard:
> @fitbeard The problem with #1197 is that openstack-cli image uses certifi and certifi does not accept private signed ca's. Certifi uses its own keystore.
> 
> I try'd to manual uninstall the pip certifi package, but then openstack application complains about missing certifi module.

[...]

PR #1188, opened 2024-05-06
fix: set for the Prometheus oauth2-proxy to skip validation

mpiscaer:


fitbeard:
Let's discuss here: https://github.com/vexxhost/atmosphere/issues/1185

roles/cluster_issuer/defaults/main.yml
```diff
@@ -15,6 +15,8 @@
 cluster_issuer_name: "{{ atmosphere_ingress_cluster_issuer }}"
 cluster_issuer_type: acme
 
+cluster_issuer_acme_private_ca: false
```
mnaser, line 18: I think if we end up defaulting this to the way we discussed in #1185 then we'll need to change this.

mnaser (changes requested):

Issue #1196, opened 2024-05-08
Keystone ServiceAccount already exists


ricolin:
Currently running `poetry run molecule converge` wiil hit Keystone ServiceAccount `keystone-api` already exists issue

I think we need some solution like adding annotations for resource already exists issue
https://handbook.giantswarm.io/docs/dev-and-releng/app-developer-processes/how-to-resolve-resources-exists-in-helm/

PR #1198, opened 2024-05-09
fix: service account resource conflicts and pxc wait timeout

ricolin:
fix https://github.com/vexxhost/atmosphere/issues/1196


ricolin:
recheck

ricolin:
all issues are detected with `poetry run molecule converge`

mnaser:
I think this issue is probably an inconsistency on your side, we run idempotency checks here.

PR #773, opened 2023-12-27
feat: neutron changes for OVN and dpdk init container

thywyn:
This adds another INIT container to initialize DPDK interfaces for OVN.  I prefered this in the OVN charts but really didn't want another place for all the DPDK values that already existed in the Neturon chart. Essentially, just copied the OVS agent init container manifest that does this for an OVS without OVN config.

I changed the br-phys to br-ex bridge name because that is a default in both the OVN and neutron charts to use.  Otherwise you'd have an orphaned br-ex doing nothing in the config.

TODO:  Needs to be tested against a non-DPDK environment..  haven't had the ability to do this yet.

PR #1203, opened 2024-05-09
feat: NIC initialization moved from neutron to openvswitch charts

thywyn:
This PR is to move the NIC initialization from neutron init scripts to openvswitch.  My PR https://github.com/vexxhost/atmosphere/pull/773 introduced a race condition between neutron-ovn-metadata-agent and ovn-controller daemonsets, where ovn-controller was waiting on metadata to initialize the bridge interface br-ex and neutron-ovn-metadata was waiting on ovn-controller.. 

moved the NIC/BOND initialization stuff from neutron to the openvswitch helm charts.  This is important for environments using DPDK, OVN/OpenVswitch deployments.

### DIFF
--- a/charts/patches/openvswitch/0002-copy-nic-init-to-openvswitch-chart.patch
+++ b/charts/patches/openvswitch/0002-copy-nic-init-to-openvswitch-chart.patch
@@ -1,0 +1,849 @@
+diff --git a/openvswitch/templates/bin/_openvswitch-nic-init.sh.tpl b/openvswitch/templates/bin/_openvswitch-nic-init.sh.tpl
+new file mode 100644
+index 0000000..84647ff
+--- /dev/null
++++ b/openvswitch/templates/bin/_openvswitch-nic-init.sh.tpl
+@@ -0,0 +1,508 @@
++#!/bin/bash
++
++{{/*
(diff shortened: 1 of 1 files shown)


thywyn:
Just as an FYI this has been tested lightly in the environment describe above.. Not in a non-DPDK environment, or using just openvswitch... hoping to test that in the future.

thywyn:
We had issues with needing more than one socket_memory config for our environment so I've added the overrides capability similar to neutron and nova

thywyn:
@mnaser I needed to add jq to the image for this stuff to work.. It seems the CI pipeline is not using an image with JQ on there... Anything I need to do more than update the dockerfile?

thywyn:
@mnaser rewrote this as a patch file for the openvswitch chart.  Orignally I removed the neutron-ovs-agent-init.sh from Neutron but technically that or this nic init script will lie dormant until used.

thywyn:
removed debug package from image change

thywyn:
Going through and applying this change after the non-root user patch broke things.. Revamping the patch to work after the non-root user patch

thywyn:
This has been merged upstream https://review.opendev.org/c/openstack/openstack-helm/+/95022

PR #1211, opened 2024-05-14
fix: perconas haproxy default cfg

fitbeard:
Since we are now using version 1.14.0 of the operator and default haproxy tag from it is also 1.14.0 we need to update haproxy config to `new` version. Without kubernetes resolver startup of haproxy pods takes much longer. And yes, it has nothing to do with crVersion: "1.13.0" from spec itself:)

### DIFF
--- a/roles/percona_xtradb_cluster/vars/main.yml
+++ b/roles/percona_xtradb_cluster/vars/main.yml
@@ -43,14 +43,21 @@ _percona_xtradb_cluster_spec:
     image: "{{ atmosphere_images['percona_xtradb_cluster_haproxy'] | vexxhost.kubernetes.docker_image('ref') }}"
     nodeSelector:
       openstack-control-plane: enabled
+    # NOTE(fitbeard): The actual default configuration file can be found here:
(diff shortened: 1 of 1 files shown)


mnaser:
ugh, I used to actually have a golang unit test that isn't running between the GHA/Jenkins/Zuul dance that would have caught this.

Do you have time to port that test to Python instead or would you like me to if you don't have spare time @fitbeard ?

fitbeard:
@mnaser I need a working example to understand whole logic. Maybe you can create tests this time? When I have source for reference i'm very productive copy-cat'er:)

okozachenko1203:
@fitbeard there are `var_test.go` files in roles and have a task in earthfile https://github.com/vexxhost/atmosphere/blob/main/Earthfile#L3-L14
This is an example test file for pxc. https://github.com/vexxhost/atmosphere/blob/main/roles/percona_xtradb_cluster/vars_test.go

btw, @mnaser i updated the current go unit test code to avoid version mismatching between image ref and cluster spec.CRVersion https://github.com/vexxhost/atmosphere/pull/1977/commits/bf5ddfac318c2b37592a54709616a238cc7d25ea
And also added go unit test in the zuul CI. https://github.com/vexxhost/atmosphere/pull/1983

okozachenko1203 (approved):

Issue #738, opened 2023-12-01
Enable extending monitoring rules for Atmosphere


mnaser:
At this point, there's no easy way of enabling users to extend the existing monitoring rules in Prometheus.  We should add a simple way to be able to do that so new rules can be added ad-hoc without overriding the whole thing.

PR #1215, opened 2024-05-15
[ATMOSPHERE-48]fix: allow prometheus extend rule

ricolin:
fix: https://github.com/vexxhost/atmosphere/issues/738

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ molecule/*/host_vars/*
 ## Misc
 *.tar.gz
 __pycache__
+tests/.coverage
 tests/output
 *.orig
 *.rej
--- a/doc/source/admin/monitoring.rst
+++ b/doc/source/admin/monitoring.rst
@@ -270,10 +270,10 @@ Here is an example:
 .. code-block:: yaml
 
    kube_prometheus_stack_extend_rules:
-     ipmi-exporter-extends: {
+     ipmi-exporter: {
          groups: [
              {
-                 name: "rules",
+                 name: "new_rules",
                  rules: [
                      {
                          alert: "IpmiCollectorDown",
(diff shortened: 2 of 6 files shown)


mnaser (changes requested):
I see a few issues with this current approach.

This means that there's no way of actually editing the existing rules, this is just extending/adding to the additional existing rules.  This means that this can simply be replaced with `kube_prometheus_stack_helm_values.additionalPrometheusRulesMap` instead which becomes just a documentation change.

I'd rather we do that than introduce a new variable for this.

mpiscaer:
Is don't know if i understand this correctly but when you use at the combine filter use 'list_merge=append' it will extend the string and then you can use the feature from jsonnet to overide/extend the existing settings.

ricolin:
> I see a few issues with this current approach.
> 
> This means that there's no way of actually editing the existing rules, this is just extending/adding to the additional existing rules. This means that this can simply be replaced with `kube_prometheus_stack_helm_values.additionalPrometheusRulesMap` instead which becomes just a documentation change.
> 
> I'd rather we do that than introduce a new variable for this.

If we need override existing, another way we I can try is to write an ansible plugin so we can handle combine on real recursive ( and avoid constraint that brought by existing combine function https://github.com/vexxhost/atmosphere/pull/1215/files#diff-2162201c39f35aed971e9b5b2fc33cb1860cfa756c8cf776683d15fe686b9cfaR291-R296 )

ricolin:
> @mnaser Is don't know if i understand this correctly but when you use at the combine filter use 'list_merge=append' it will extend the string and then you can use the feature from jsonnet to overide/extend the existing settings.

The issue with directly using list_merge=append for combine, is that will give us two rule group with same name and makes error for charts like 
```
{ 
    "ipmi-exporter": {
        "groups": [
            {
                "name": "rules",
                "rules": [ { "alert": "IpmiCollectorDown", "expr": "ipmi_up == 0", "for": "10m", "labels": {"severity": "P3"}, } ],
            },
            {
                "name": "rules",
                "rules": [ { "alert": "IpmiCollectorDown", "expr": "ipmi_up == 0", "for": "16m", "labels": {"severity": "P2"}, } ],
            },
        ]
    }   
}
```
I just added a plugin to handle that error

ricolin:
recheck

PR #1216, opened 2024-05-15
feat: trust private acme CA

fitbeard:
Related: https://github.com/vexxhost/atmosphere/issues/1185

We disscused with @mpiscaer on Slack about the reason why this happens and found that such a configuration was not foreseen.

Mounting whole directory like here: https://github.com/vexxhost/ansible-collection-kubernetes/blob/main/roles/cert_manager/vars/main.yml#L21 is a bad idea because on EL distros this host path doesn't exist.

### DIFF
--- a/roles/cluster_issuer/defaults/main.yml
+++ b/roles/cluster_issuer/defaults/main.yml
@@ -15,6 +15,8 @@
 cluster_issuer_name: "{{ atmosphere_ingress_cluster_issuer }}"
 cluster_issuer_type: acme
 
+cluster_issuer_acme_private_ca: false
+
 cluster_issuer_acme_server: https://acme-v02.api.letsencrypt.org/directory
 # cluster_issuer_acme_email:
 cluster_issuer_acme_private_key_secret_name: cert-manager-issuer-account-key
--- a/roles/openstack_cli/templates/atmosphere.sh.j2
+++ b/roles/openstack_cli/templates/atmosphere.sh.j2
@@ -1,9 +1,13 @@
 alias osc='nerdctl run --rm --network host \
       --volume $PWD:/opt --volume /tmp:/tmp \
       --volume /etc/openstack:/etc/openstack:ro \
-{% if (cluster_issuer_type is defined and cluster_issuer_type in ('self-signed', 'ca')) or (cluster_issuer_acme_private_ca is defined and cluster_issuer_acme_private_ca | bool) %}
+{% if cluster_issuer_type is defined and cluster_issuer_type in ('self-signed', 'ca') %}
       --volume {{ '/usr/local/share/ca-certificates/atmosphere.crt:/usr/local/share/ca-certificates/atmosphere.crt:ro' if ansible_facts['os_family']
-      in ['Debian'] else '/etc/pki/ca-trust/source/anchors/atmosphere.crt:/etc/pki/ca-trust/source/anchors/atmosphere.crt:ro' }} \
+      in ['Debian'] else '/etc/pki/ca-trust/source/anchors/atmosphere.crt:/usr/local/share/ca-certificates/atmosphere.crt:ro' }} \
+{% endif %}
+{% if cluster_issuer_acme_private_ca is defined and cluster_issuer_acme_private_ca | bool %}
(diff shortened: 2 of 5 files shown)


roles/cluster_issuer/defaults/main.yml
```diff
@@ -15,6 +15,8 @@
 cluster_issuer_name: "{{ atmosphere_ingress_cluster_issuer }}"
 cluster_issuer_type: acme
 
+cluster_issuer_acme_private_ca: false
```
mnaser, line 18: Could we technically default this to `{{ cluster_issuer_acme_server != 'https://acme-v02.api.letsencrypt.org/directory' }}` ?

roles/openstack_cli/templates/atmosphere.sh.j2
```diff
@@ -3,7 +3,11 @@ alias osc='nerdctl run --rm --network host \
       --volume /etc/openstack:/etc/openstack:ro \
 {% if cluster_issuer_type is defined and cluster_issuer_type in ('self-signed', 'ca') %}
       --volume {{ '/usr/local/share/ca-certificates/atmosphere.crt:/usr/local/share/ca-certificates/atmosphere.crt:ro' if ansible_facts['os_family']
-      in ['Debian'] else '/etc/pki/ca-trust/source/anchors/atmosphere.crt:/etc/pki/ca-trust/source/anchors/atmosphere.crt:ro' }} \
+      in ['Debian'] else '/etc/pki/ca-trust/source/anchors/atmosphere.crt:/usr/local/share/ca-certificates/atmosphere.crt:ro' }} \
+{% endif %}
+{% if cluster_issuer_acme_private_ca is defined and cluster_issuer_acme_private_ca | bool %}
```
mnaser, line 8: any reason we can't make this `elif`?

roles/openstack_cli/templates/openrc.j2
```diff
@@ -12,5 +12,8 @@ export OS_PROJECT_DOMAIN_NAME=Default
 export OS_PROJECT_NAME=admin
 
 {% if cluster_issuer_type is defined and cluster_issuer_type in ('self-signed', 'ca') %}
-export OS_CACERT={{ '/usr/local/share/ca-certificates' if ansible_facts['os_family'] in ['Debian'] else '/etc/pki/ca-trust/source/anchors' }}/atmosphere.crt
+export OS_CACERT=/usr/local/share/ca-certificates/atmosphere.crt
+{% endif %}
+{% if cluster_issuer_acme_private_ca is defined and cluster_issuer_acme_private_ca | bool %}
```
mnaser, line 17: same here

roles/openstacksdk/templates/clouds.yaml.j2
```diff
@@ -9,5 +9,8 @@ clouds:
       project_domain_name: Default
     region_name: "{{ openstack_helm_endpoints_keystone_region_name }}"
 {% if cluster_issuer_type is defined and cluster_issuer_type in ('self-signed', 'ca') %}
-    cacert: "{{ '/usr/local/share/ca-certificates' if ansible_facts['os_family'] in ['Debian'] else '/etc/pki/ca-trust/source/anchors' }}/atmosphere.crt"
+    cacert: "/usr/local/share/ca-certificates/atmosphere.crt"
+{% endif %}
+{% if cluster_issuer_acme_private_ca is defined and cluster_issuer_acme_private_ca | bool %}
```
mnaser, line 14: and here too

mnaser (changes requested):

roles/cluster_issuer/defaults/main.yml
mpiscaer: I think setting 'cluster_issuer_acme_server' does not always means a private ca is used.

fitbeard: I know at least two other publicly trusted SSL certificate providers with ACME support:

- zerossl
- buypass

mnaser: TIL!

mnaser (approved):

PR #1218, opened 2024-05-16
fix: images: horizon: fix swift container URL

ricolin:
ref: https://review.opendev.org/c/openstack/horizon/+/886891


ricolin:
okay, we use 2024 horizon, and the fix is there. need to directly patch to zed

PR #1219, opened 2024-05-16
fix: images: horizon: fix swift container URL

ricolin:
ref: https://review.opendev.org/c/openstack/horizon/+/886891


ricolin:
we use 2024.1 horizon, and the fix is there and can't be merged. So need to directly patch to zed

PR #1241, opened 2024-05-24
[ATMOSPHERE-362] generate_workspace: Set mgr/cephadm/warn_on_stray_daemons to false

gtirloni:
When rook-ceph deploys daemon outside of cephadm purview, cephadm will trigger a warning about stray daemons, putting the cluster in HEALTH_WARN state.

https://docs.ceph.com/en/latest/cephadm/operations/#cephadm-stray-daemon

### DIFF
--- a/playbooks/ceph.yml
+++ b/playbooks/ceph.yml
@@ -26,3 +26,8 @@
 - import_playbook: vexxhost.ceph.site
   vars:
     containerd_pause_image: "{{ atmosphere_images['pause'] }}"
+    ceph_conf_overrides:
+      # Avoid HEALTH_WARN state due to daemons deployed by rook-ceph outside of cephadm (e.g. rgw)
+      - section: mgr
+        option: mgr cephadm warn_on_stray_daemons
+        value: false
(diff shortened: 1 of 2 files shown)


mnaser (changes requested):
This is awesome.  This fix that Rico implemented was not so clean but I didn't know this existed.

https://github.com/vexxhost/atmosphere/blob/abb177586fda668aea747e54231d5d1630bfb57e/roles/rook_ceph_cluster/tasks/main.yml#L15-L22

I would push to make this an option in `ansible-collection-ceph` and then when Atmosphere calls it, we make it disable this option.  What do you think?

ricolin:
I think we should also backport this to 2023.1 and zed too

gtirloni:
@ricolin @mnaser  so the ceph collection accepts `ceph_conf_overrides` already and it's just an option in an ini file. Since it's inherent to the way atmosphere deploys Ceph, I've moved it from generate_workspace (which is optional, AFAIK) to the ceph playbook here so it's always applied. Does that look okay? If you think it's better to have a variable in some role in the ceph collection to disable/enable this, I can do that as well.

ricolin:
recheck

gtirloni:
@ricolin any tips about how to extract the state of the pod that's stuck in Pending in this CI build? i can't reproduce locally but my setup might be different.

gtirloni:
recheck

gtirloni:
recheck

mnaser (approved):

Issue #1245, opened 2024-05-28
bug: allow ssh login for nova user for nova-compute-ssh image


okozachenko1203:
## Context
Instance migraiton is failing with `protocol error: mtime.sec not present` error.

## Reason
We use `/usr/sbin/nologin` for users created during image to not allow ssh login.
https://github.com/vexxhost/atmosphere/blob/abb177586fda668aea747e54231d5d1630bfb57e/images/openstack-runtime/Dockerfile#L20C13-L23

it should be allowed for nova-compute-ssh.

mnaser:
Good catch.  Hopefully this doesn't bite us again, would be nice if there was some 'unit tests' for images that exist, but oh well :)

PR #1246, opened 2024-05-28
fix: allow ssh login for nova user in nova-ssh image

okozachenko1203:
nologin avoids ssh connection between noova-computes which is required for live migrations.

fix https://github.com/vexxhost/atmosphere/issues/1245


mnaser (approved):

mnaser:
recheck

PR #1255, opened 2024-05-29
Mount CA into Keystone + Magnum

mnaser:


fitbeard:
@mnaser this is valid only with Debian family:/ Like here: https://github.com/vexxhost/ansible-collection-kubernetes/blob/896d67092706d01b6fb9d7f4856750819f8898cf/roles/cert_manager/vars/main.yml#L25 this will break things on rhel. In cert-manager in rockylinux i have to override this to pass empty list. Otherwise it breaks cert-manager entirely.

fitbeard:
Aka empty mount.

mnaser:
> Wouldn't it be better to use a secret where the CA is stored or an ansible variable used in various places?

I think this is an easier first approach to "get it done".  I think long term we should look into a solution where we capture all the little nooks and crannies and mount a secret with the CA everywhere a private CA is used.. but this kinda "unblocks" us enough in combination with the cert-manager CA installation.

Also, good catch @fitbeard -- let me fix it.

mnaser:
@fitbeard @mpiscaer let me know what you think of this revision.  This makes it easy for new to avoid having too much logic for private CA vs not private CA and defers it to the host instead.

I do think long term having it mounted directly from a secret would be the nicest.

fitbeard:
@mnaser from rhel:
```
[root@controller1 ~]# ls -al /etc/pki/tls/certs/
total 0
drwxr-xr-x. 2 root root  54 Jan 23 02:46 .
drwxr-xr-x. 5 root root 126 Apr 30 15:19 ..
lrwxrwxrwx. 1 root root  49 Sep 13  2023 ca-bundle.crt -> /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
lrwxrwxrwx. 1 root root  55 Sep 13  2023 ca-bundle.trust.crt -> /etc/pki/ca-trust/extracted/openssl/ca-bundle.trust.crt
```
...so the real path on host is `/etc/pki/ca-trust/extracted/openssl` and file name is `ca-bundle.trust.crt` instead of `ca-certificates.crt`

Example: https://github.com/vexxhost/atmosphere/pull/1238/files#diff-129ada3e23e707220637b1ff6840a8d2ee2c3b06c988af91787c0ea6b3b7df50R9

PS. someday i'll fix mounts for cert-manager role

mnaser:
@fitbeard Since I'm mounting a folder, do you think it'll actually matter if I point the exact file.. like would it be better to mount the whole folder instead of the file?

fitbeard:
I think Ubuntu expects ca-certificates.crt under this path and trusts only content from there.

mnaser:
@fitbeard I think this won't break your use case I believe.

fitbeard:
@mnaser nice!

PR #1257, opened 2024-05-30
Align horizon chart

KeplerSysAdmin:
- Update pymysql version info. Horizon requiers newer mysql client version than we had previously
- Update scripts to use sysconfig.get_path instead of distutils.sysconfig.get_python_lib which is deprecated

reference: https://review.opendev.org/c/openstack/openstack-helm/+/919778

### DIFF
--- a/.charts.yml
+++ b/.charts.yml
@@ -70,7 +70,7 @@ charts:
         review.opendev.org:
           - 899931
   - name: horizon
-    version: 0.3.15
+    version: 0.3.22
     repository: *openstack_helm_repository
     dependencies: *openstack_helm_dependencies
     patches:
--- a/charts/horizon/Chart.yaml
+++ b/charts/horizon/Chart.yaml
@@ -9,4 +9,4 @@ name: horizon
 sources:
 - https://opendev.org/openstack/horizon
 - https://opendev.org/openstack/openstack-helm
-version: 0.3.15
+version: 0.3.22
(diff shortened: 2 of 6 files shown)


mnaser:
Can you please look into making those changes in the `.charts.yml` file, either as a patch or simply as a bump and then running `too -e sync-charts` in order to see things in sync?

Thanks for the contribution!

KeplerSysAdmin:
> Can you please look into making those changes in the `.charts.yml` file, either as a patch or simply as a bump and then running `too -e sync-charts` in order to see things in sync?
> 
> Thanks for the contribution!

Align horizon chart version on file .chart.yml

mnaser:
Did you make sure to run the sync after updating?  I doubt that there was no updates from .15 to .22 :)

KeplerSysAdmin:
> Did you make sure to run the sync after updating? I doubt that there was no updates from .15 to .22 :)

I'm not sure what you mean.

mnaser:
You will need to run `tox -e sync-charts` after making the change in charts which will resynchronize all the charts.

KeplerSysAdmin:
Those are a lot unexpected changes:
Screenshot from 2024-05-30 23-26-40

mnaser:
There shouldn't be that much changes if you run it, perhaps the process fails, do you have patch diff missing perhaps?

KeplerSysAdmin:
Hi @mnaser 
Please close this ticket, it is aligned on tag v1.11.2 so this work is unnecessary

PR #1264, opened 2024-05-30
Build requirements from pinned Git hash

mnaser:
Depends-On: https://github.com/vexxhost/atmosphere/pull/1266

PR #1269, opened 2024-05-30
Add rebuild time for images

mnaser:
If we want to rebuild all images for security reasons or when
branching a new version of OpenStack, we need to be able to
rebuild all images.  Doing this change in the job definition
is not enough to cause the promotion to happen.

This adds a tool that can modify the file by passing a Dockerfile
or many to it.

Depends-On: https://github.com/vexxhost/atmosphere/pull/1264


mnaser:
Blocked by #1281

mnaser:
recheck

mnaser:
recheck

Issue #1278, opened 2024-05-31
Fail on ingress template


ricolin:
current ingress annotation in roles/ingress/templates/ingress.yml.j2 got changes to become form like below if both ingress_annotations and ingress_global_annotations happen to keep in default dict( `{}`).

```
  annotations:
    cert-manager.io/common-name: service_name.domain_name.atmosphere.dev
    {} 
```

PR #1279, opened 2024-05-31
fix: fix ingress annotations format

ricolin:
fix https://github.com/vexxhost/atmosphere/issues/1278

### DIFF
--- a/roles/ingress/templates/ingress.yml.j2
+++ b/roles/ingress/templates/ingress.yml.j2
@@ -5,7 +5,9 @@ metadata:
   namespace: {{ ingress_namespace | default('default') }}
   annotations:
     cert-manager.io/common-name: {{ ingress_host }}
+{% if (ingress_annotations|length) > 0 or (ingress_global_annotations|length) > 0 %}
     {{ ingress_annotations | combine(ingress_global_annotations, recursive=True) | to_nice_yaml | indent(4) }}
+{% endif %}
 spec:
   ingressClassName: {{ ingress_class_name }}
   rules:


roles/ingress/defaults/main.yml
```diff
@@ -38,5 +38,9 @@ ingress_class_name: "{{ atmosphere_ingress_class_name }}"
 #               annotations to all Ingress resources.
 ingress_annotations: {}
 
+# Seperate cert manager annotations so ingress_annotations can be override.
+ingress_cert_annotations:
+  cert-manager.io/common-name: "{{ ingress_host }}"
```
mnaser, line 43: @ricolin is there a reason we can't keep the existing template and not introduce a new variable to handle this?

mnaser (changes requested):

roles/ingress/defaults/main.yml
ricolin: depends on how we like the override to works here.
We can totally write a hard code dict in template like ansible.builtin.combine({cert-manager.io/common-name: "{{ ingress_host }}"}
or put this cert-manager stuff into `ingress_annotations`

ricolin: like this?
```
  annotations:
      {{ { cert-manager.io/common-name: {{ ingress_host }} } | combine(ingress_annotations, recursive=True) | combine(ingress_global_annotations, recursive=True) | to_nice_yaml | indent(4) }
```

ricolin: updated, please check :)

mnaser: Is there no way of keeping the annotation hard coded the way it is and then adding the extra annotation on top of it?

ricolin: Another way is to check keys before parse to nice yaml function so {} won't be generated.
Just updated again. Please check. Thanks

ricolin:
recheck

ricolin:
recheck

ricolin:
recheck

mnaser (approved):

PR #1306, opened 2024-06-04
fix: nova capacity alert

mpiscaer:
the NovaCapacity it does not take ram_allocation_ratio into account I have hardcoded this at 90%. But U think we can do better


mnaser:
regate

Issue #1302, opened 2024-06-04
rook ceph exporter init failed with secret "rook-ceph-exporter-keyring" not found


ricolin:
```
root@instance:/home/ubuntu# kubectl -n openstack describe pod rook-ceph-exporter-instance-54df8dfc9b-2v996
Events:
  Type     Reason       Age                  From     Message
  ----     ------       ----                 ----     -------
  Warning  FailedMount  42s (x869 over 29h)  kubelet  MountVolume.SetUp failed for volume "rook-ceph-exporter-keyring" : secret "rook-ceph-exporter-keyring" not found
```

This is due to missing secret like reported on https://github.com/rook/rook/issues/14275
This issue affect rook-ceph > 1.13.0 as https://github.com/rook/rook/pull/13239 introduced separate keyring for exporter but somehow didn't actually create exporter keyrings out

parth-gr:
Seems like you havent create the external cephcluster CR https://github.com/rook/rook/blob/master/deploy/examples/external/cluster-external.yaml

mnaser:
https://github.com/vexxhost/atmosphere/pull/1896

Issue #1282, opened 2024-05-31
improve upgrade from zed to 2023.1


ricolin:

mnaser:
@ricolin the missing images are now promoted/uploaded and this shouldn't happen again!  :)

mnaser:
closing in favour of the #1068

PR #1309, opened 2024-06-05
fix: downgrade rook ceph

ricolin:
fix https://github.com/vexxhost/atmosphere/issues/1302
relate to https://github.com/vexxhost/atmosphere/issues/1282


ricolin:
recheck

mnaser (changes requested):
@ricolin I need more context on this change, as I'm not able to review, I don't know why the downgrade is happening or anything else.

PR #1308, opened 2024-06-05
chore: add upgrade job from zed to 2023.1

ricolin:
fix https://github.com/vexxhost/atmosphere/issues/1282

PR #1319, opened 2024-06-05
fix: reinstall prometheus_pushgateway when <2.0.0

ricolin:
Reinstall prometheus_pushgateway helm when existing release <2.0.0

Related-To: #1282
Related-To: #1308

### DIFF
--- a/roles/prometheus_pushgateway/tasks/main.yml
+++ b/roles/prometheus_pushgateway/tasks/main.yml
@@ -12,6 +12,22 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Gather information prometheus_pushgateway
+  run_once: true
+  kubernetes.core.helm_info:
+    name: "{{ prometheus_pushgateway_helm_release_name }}"
+    release_namespace: "{{ prometheus_pushgateway_helm_release_namespace }}"
+  register: result
+
+- name: Uninstall Helm chart
+  run_once: true
+  kubernetes.core.helm:
+    name: "{{ prometheus_pushgateway_helm_release_name }}"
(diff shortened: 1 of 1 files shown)


roles/prometheus_pushgateway/tasks/main.yml
```diff
@@ -12,6 +12,22 @@
...
+  register: result
+
+- name: Uninstall Helm chart
+  run_once: true
+  kubernetes.core.helm:
+    name: "{{ prometheus_pushgateway_helm_release_name }}"
+    release_namespace: "{{ prometheus_pushgateway_helm_release_namespace }}"
+    state: absent
+    wait: true
+  when: "'status' in result and result.status.status is match('deployed') and (result.status.chart | replace(prometheus_pushgateway_helm_release_name + '-','')) is version('2.0.0', operator='lt', strict=True)"  # noqa: yaml[line-length]
```
mnaser, line 29:
```suggestion
  when:
    - "'status' in result"
    - "result.status.status is match('deployed')"
    - "(result.status.chart | replace(prometheus_pushgateway_helm_release_name + '-','')) is version('2.0.0', operator='lt', strict=True)"  # noqa: yaml[line-length]
```

It does feel a little clunky though, let's see how it goes on the CI.

mnaser (changes requested):

mnaser:
@ricolin I have fixed changed this to be a `stable/2023.1` change only since I don't think this will apply for anything else.

mnaser:
recheck

mnaser:
regate

mnaser (approved):

PR #1328, opened 2024-06-07
fix(neutron): Add missing parameters to create networks task

gtirloni:
Add:
* admin_state_up
* dns_domain
* project


mnaser:
recheck

mnaser:
recheck

PR #1330, opened 2024-06-07
Add cinder-tmp to c-vol

guilhermesteinmuller:
When using other backends like purestorage and users wants to create volume that are not raw, we need to have cinder-tmp dir inside the c-vol container to be able to perform the image copy and conversion.

### DIFF
--- a/charts/cinder/templates/deployment-volume.yaml
+++ b/charts/cinder/templates/deployment-volume.yaml
@@ -157,8 +157,6 @@ spec:
 {{- end }}
           terminationMessagePath: /var/log/termination-log
           volumeMounts:
-            - name: cinder-tmp
-              mountPath: /var/lib/cinder/tmp
             - name: pod-tmp
               mountPath: /tmp
             - name: cinder-bin
@@ -288,8 +286,6 @@ spec:
 {{- dict "enabled" $envAll.Values.manifests.certificates "name" $envAll.Values.endpoints.oslo_messaging.auth.admin.secret.tls.internal "path" "/etc/rabbitmq/certs" | include "helm-toolkit.snippets.tls_volume_mount" | indent 12 }}
 {{ if $mounts_cinder_volume.volumeMounts }}{{ toYaml $mounts_cinder_volume.volumeMounts | indent 12 }}{{ end }}
       volumes:
-        - name: cinder-tmp
-          emptyDir: {}
         - name: pod-tmp
           emptyDir: {}
         - name: cinder-bin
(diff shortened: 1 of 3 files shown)


mnaser (changes requested):
let's leverage `volumeMounts` and `volumes` inside the `pod` section inside the values.

charts/cinder/templates/deployment-volume.yaml
```diff
@@ -286,6 +288,8 @@ spec:
 {{- dict "enabled" $envAll.Values.manifests.certificates "name" $envAll.Values.endpoints.oslo_messaging.auth.admin.secret.tls.internal "path" "/etc/rabbitmq/certs" | include "helm-toolkit.snippets.tls_volume_mount" | indent 12 }}
 {{ if $mounts_cinder_volume.volumeMounts }}{{ toYaml $mounts_cinder_volume.volumeMounts | indent 12 }}{{ end }}
       volumes:
+        - name: cinder-tmp
+          emptyDir: {}
```
gtirloni, line 292: considering the possibility for non-malicious abuse here, it would be nice to set a [sizeLimit](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#emptydirvolumesource-v1-core) to prevent a single user from completely filling up the disk area where this is allocated from.

guilhermesteinmuller: huuum that could be a good approach, but im thinking of the limitations that could cause. 

In one way, I completely agree that having this conversion locally on a controller and having no control over the sizes is a dangerous thing.... thats WHY we are disabling this upstream. 

On the other hand, it's a limitation for users to be always using a raw images. Thats difficult to control unless we explicitly tell them the risks and create a note for ALL the users to be only using RAW images

charts/cinder/values.yaml
```diff
@@ -204,7 +204,11 @@ pod:
       init_container: null
       cinder_volume:
         volumeMounts:
+          - name: cinder-tmp
+            mountPath: /var/lib/cinder/tmp
         volumes:
+          - name: cinder-tmp
+           emptyDir: {}
```
mnaser, line 211: Also there is a missing space here too.

mnaser (changes requested):

guilhermesteinmuller:
> Please move those changes into `roles/cinder/vars/main.yml` so that they can be used from there instead.

@mnaser  I though I have removed in the 3rd commit here

roles/cinder/vars/main.yml
```diff
@@ -20,6 +20,16 @@ __cinder_helm_values:
     replicas:
       api: 3
       scheduler: 3
+    mounts:
+      cinder_volume:
+        init_container: null
```
mnaser, line 25: This can be removed, as it's not necessary.

mnaser (changes requested):

mnaser (approved):

PR #1338, opened 2024-06-11
Add secretsgen-controller chart (v0.16.0)

gtirloni:
To migrate to the Helm chart:

```
kubectl delete ns secretgen-controller
ansible-playbook -t secretgen-controller playbooks/site.yml
```

logs should show it detected all the certificates and other resources it manages.


mnaser (changes requested):
I think we need to handle upgrade impact, because I think Helm will give errors if the resource already exists.

gtirloni:
> I think we need to handle upgrade impact, because I think Helm will give errors if the resource already exists.

would it be okay to have a task to delete the secretgen-operator namespace in this case? that's what i did in my tests and it works fine

mnaser:
Sure, but then it would delete it and redeploy on every single run so we just have to accommodate for that :)

gtirloni:
> Sure, but then it would delete it and redeploy on every single run so we just have to accommodate for that :)

I added a task that ensures the secretgen-controller namespace is absent. I think that should do it.

I actually moved the controller into the kube-system together with other cluster-wide services. I think that's what you might mean with with things getting redeployed (that is, if they were in the original namespace)?

I think there's no practical requirement to have a separate namespace for this, that's why I moved it.

gtirloni:
Related #1368

PR #1342, opened 2024-06-12
fix: add more retries to mysql_ready check

ricolin:
relate to https://github.com/vexxhost/atmosphere/issues/1282
relate to https://github.com/vexxhost/atmosphere/pull/1308


ricolin:
recheck

PR #1343, opened 2024-06-12
ipmi-exporter: Add ability to specify imagePullSecrets

gtirloni:
Quick fix to add imagePullSecrets. Proper fix is to turn this into a helm chart.


gtirloni:
Related #1368

PR #1350, opened 2024-06-14
feat: build arm64 images for computes

fitbeard:


mnaser:
recheck

fitbeard:
The question is why tests did not failed without base image layer for arm64

mnaser:
> The question is why tests did not failed without base image layer for arm64

My guess is that since that layer is pretty empty then somehow Docker dedups it to the upstream version?

mnaser:
@fitbeard how far back do you want to back port this btw?

fitbeard:
@mnaser would be nice to have arm starting from 2023.1

mnaser:
I just realized we need them for zed too :)

mnaser:
it seems that has uncovered a missing dependency where the ovn jobs aren't actually depending on the ovs jobs

mnaser:
regate

mnaser:
https://github.com/vexxhost/atmosphere/pull/1408

That's the missing piece I believe.

mnaser:
regate

Issue #1068, opened 2024-04-05
Introduce major releases


mnaser:
At the moment, `main` simply points to a single version of OpenStack.  However, we've noticed that customers do sometimes want to run things for longer.  It also makes it tougher for upgrade paths, therefore, we're going to plan to adding major releases that correspond to OpenStack releases, with the plan being the following:

- [x] `main`: move all images to point to `master` branches of OpenStack
- [x] `stable/zed`: branch from `main`, update all images to point to `stable/zed` (https://github.com/vexxhost/atmosphere/pull/1069)
- [x] `stable/zed`: release as `1.x`
- [x] `stable/2023.1`: branch from `main`, update all images to point to `stable/2023.1` (https://github.com/vexxhost/atmosphere/pull/1227)
- [x] `stable/2023.1`: introduce job to deploy `stable/zed` and upgrade to `stable/2023.1` (https://github.com/vexxhost/atmosphere/issues/1282)
- [x] `stable/2023.1`: release as `2.x`
- [x] `stable/2023.2`: branch from `main`, update all images to point to `stable/2023.2`
- [x] `stable/2023.2`: introduce jobs to deploy from `stable/2023.1` and upgrade to `stable/2023.2`
- [x] `stable/2023.2`: release as `3.x`
- [x] `stable/2024.1`: branch from `main`, update all images to point to `stable/2024.1`
- [x] `stable/2024.1`: introduce jobs to deploy from `stable/2023.2` and upgrade to `stable/2024.1`
- [x] `stable/2024.1`: release as `4.x`
- [x] https://github.com/vexxhost/atmosphere/issues/1267

For upstream, we'll support releases on a best effort basis and deal with direct customers on individual basis if they are really behind.  It'll allow users to spend more time on their releases.

For upgrades, the ideal path would be major => major without jumping, and for those who feel comfortable can jump releases.

mpiscaer:
This sounds very help full and good.

mnaser:
Alright, the next step here is to update `main` to point at `master` OpenStack, since we've released 1.11.0 for those users who want to continue to carry on using Zed.

From there, we'll branch out the other releases.

mnaser:
The PR https://github.com/vexxhost/atmosphere/pull/1090 brings the `main` branch in line with what we're doing in `stable/zed` so that backports can happen cleaner over there.

okozachenko1203:
[...]

PR #1384, opened 2024-06-18
chore: add upgrade job

ricolin:
Relate-to: https://github.com/vexxhost/atmosphere/issues/1068
Relate-to: https://github.com/vexxhost/atmosphere/issues/1282

Depends-On: https://github.com/vexxhost/atmosphere/pull/1408
Depends-On: https://github.com/vexxhost/atmosphere/pull/1409

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -84,3 +84,24 @@
     voting: false
     vars:
       tox_envlist: molecule-aio-ovn
+
+- job:
+    name: atmosphere-molecule-aio-upgrade-base
+    run: zuul.d/playbooks/upgrade/run-upgrade.yml
+    abstract: true
+    timeout: 10800
+    parent: atmosphere-molecule-aio-full
+    vars:
+      from_branch: "stable/2024.1"
+
+- job:
+    name: atmosphere-molecule-aio-upgrade-openvswitch
+    parent: atmosphere-molecule-aio-upgrade-base
+    vars:
+      network_backend: "openvswitch"
+
+- job:
+    name: atmosphere-molecule-aio-upgrade-ovn
+    parent: atmosphere-molecule-aio-upgrade-base
+    vars:
+      network_backend: "ovn"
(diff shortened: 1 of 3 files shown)


mnaser:
@ricolin does it make sense to branch 2024.1 so we can make upgrade job from 2024.1 to main since thats a more supported upgrade route?

The other thing I wanted to bring up is what do you think of introducing a new upgrade pipeline, because if we run upgrade jobs on every commit, we doubled or CI time potentially (which is already kinda long tbh)...

ricolin:
> @ricolin does it make sense to branch 2024.1 so we can make upgrade job from 2024.1 to main since thats a more supported upgrade route?
I trust your experience with upgrade:) and avoid cross multi-version upgrade sounds like a good plan too.

> 
> The other thing I wanted to bring up is what do you think of introducing a new upgrade pipeline, because if we run upgrade jobs on every commit, we doubled or CI time potentially (which is already kinda long tbh)...

Agree, +1 on a upgrade pipeline with comment_filter so we can do verify when we plan to before release or make it a perodic job.

roles/keycloak/tasks/main.yml
```diff
@@ -34,7 +34,7 @@
       - version
   register: mysql_ready
   until: mysql_ready is not failed
-  retries: 60
```
mnaser, line 37: can you make this a separate change please?

mnaser (changes requested):
I'm going to spend a bit of time thinking how we can accomplish this in a way where we have a separate pipeline probably or a periodic job.

roles/keycloak/tasks/main.yml
ricolin: https://github.com/vexxhost/atmosphere/pull/1409

zuul.d/jobs.yaml
```diff
@@ -84,3 +84,24 @@
...
+    name: atmosphere-molecule-aio-upgrade-base
+    run: zuul.d/playbooks/upgrade/run-upgrade.yml
+    abstract: true
+    timeout: 10800
+    parent: atmosphere-molecule-aio-full
+    vars:
+      from_branch: "stable/2023.2"
+
+- job:
+    name: atmosphere-molecule-aio-upgrade-ovs
```
mnaser, line 98: For consistency, please rename the job to use `openvswitch` instead of `ovs`.

mnaser (changes requested):

zuul.d/jobs.yaml
ricolin: done

ricolin:
recheck

mnaser (approved):

PR #1391, opened 2024-06-18
feat: add full Rook support

fitbeard:
This change introduces full Rook support also preserving old functionality. This means that without setting `atmosphere_rook_ceph_enabled: true` everything should be like before - ceph with cephadm  and "thin" Rook in `openstack` namespace for RadosGateway only. There are few places with code dublication also to preserve maximal original functionality. Before starting writing Molecule tests I need your opinion Yay or Nay:)

btw: tested on real hw.


mnaser:
not a review of the code, but yay.  I would like this option to be available until I feel like I trust it a bit more than cephadm :)

fitbeard:
@mnaser this change is ready for initial review. I just temporarily disabled other AIO tests to not waste resources.

PR #1428, opened 2024-06-21
feat: add neutron_policy_server support

ricolin:
Depends-On https://github.com/vexxhost/neutron-policy-server/pull/1


mnaser (changes requested):
please run this via uwsgi :)

mnaser (changes requested):
@ricolin small revision, I think it will make life a lot easier, I think instead of building a new image, we should instead have the existing image install neutron-policy-server in the same image, therefore, it will allow us to co-install and share dependencies and re-use the same image.

I think we will need to finalize that release first though, let me get that through.

roles/neutron/defaults/main.yml
```diff
@@ -27,3 +27,4 @@ neutron_ingress_annotations: {}
 
 # Enable dns integration
 neutron_designate_integration_enabled: false
+neutron_policy_server_integration_enabled: false
```
mnaser, line 30: Honestly, I see the usefulness of this for everyone, so let's enable it by default, and add a comment above it and an extra space.

mnaser (changes requested):

ricolin:
I think the last uwsgi/image change introduce the failed on CI. Debuging now

mnaser:
regate

PR #1456, opened 2024-06-26
Start using Tini for Open vSwitch

mnaser:
- **Add tini to Open vSwitch**
- **Bump Open vSwitch chart to support Tini**


mnaser:
regate

PR #1464, opened 2024-06-27
[stable/2023.2] chore: use git ref for neutron-vpnaas

ricolin:
relate to https://github.com/vexxhost/atmosphere/pull/1458


mnaser:
regate

Issue #352, opened 2023-03-04
Prometheus using emptydir instead of persistence 


guilhermesteinmuller:
We should set prometheus to use persistence so it stores data inside the storage cluster and doest take local storage from controllers

mnaser:
This is starting to become more of an issue as we are scraping more and more metrics in the monitoring.

1. We are running out of space on contollers
2. The solution involves losing data, so not a good time.

We need to explore either #524 but before that, we need to understand why we have so much data in the first place.

mnaser:
For `ethtool`, we are seeing around an average of 1000 records per system with some systems having almost 10K records per host... leading to envs with almost 50K records...

That's a lot of data for us to only need this...

https://github.com/vexxhost/atmosphere/blob/c8785e0def34d628e9e4d53fb835a38246f9189f/roles/kube_prometheus_stack/files/jsonnet/legacy.libsonnet#L10

mnaser:
A workaround for environments where https://github.com/vexxhost/atmosphere/commit/3a0cc7582ae830efd2469f26132f2141417b0955 is not deployed yet is to run this:

```bash
kubectl -n monitoring delete podmonitor/ethtool-exporter
kubectl -n monitoring delete ds/ethtool-exporter
```

of course, in the next deployment, it will go boom, but something to keep in mind.

PR #1490, opened 2024-07-03
fix: use pvc for prometheus

ricolin:
fix https://github.com/vexxhost/atmosphere/issues/352


mnaser:
Duplicate of #1652 and we can stick to 100G in the other PR, that's more space.

PR #1493, opened 2024-07-03
Fix issue Live migration broken on 2023.1

yaguangtang:
Fixes CLOUDOPS-233-live-migration-broken-on-2023-1


mnaser (changes requested):
Can you move the comment above it and do like this

```
# NOTE(mnaser): ....
```

Instead use your username :)

mnaser (approved):

PR #1518, opened 2024-07-05
Remove Senlin project code as it retires after 2024.1

### LINKED ISSUE #1267: Remove Senlin role in main after branching `stable/2024.1`
Senlin did not have a release for `stable/2024.1` and got retired.  We already worked on landing the code for it's removal in #1266 but we should remove that in `main` after we branch to not have dead code floating around.

yaguangtang:
fix #1267


charts/memcached/values_overrides/netpol.yaml
```diff
@@ -44,9 +44,6 @@ network_policy:
         - podSelector:
             matchLabels:
               application: neutron
-        - podSelector:
-            matchLabels:
-              application: senlin
```
mnaser, line 49: Can you please remove undo this change?

mnaser (changes requested):

charts/memcached/values_overrides/netpol.yaml
yaguangtang: done with a new commit

mnaser (approved):

PR #1538, opened 2024-07-08
[stable/2024.1] OSSA-2024-001

yaguangtang:
This commit updates  2024.1 stable commit id of Nova,Glance,Cinder Dockerfile to fix OSSA-2024-001 security issue.


mnaser:
recheck

PR #1549, opened 2024-07-09
Fix issue failed to allocate directory watch

### LINKED ISSUE #1492: Failed to allocate directory watch: Too many open files
Since the running on Jammy I sometimes see the error `Failed to allocate directory watch: Too many open files` when doing `apt update` in our "development" clusters.

To solve this we set the values:
```
root@ctl1:~# sysctl fs.inotify
fs.inotify.max_queued_events = 1048576
fs.inotify.max_user_instances = 8192
fs.inotify.max_user_watches = 1048576
```

yaguangtang:
This fixes #1492 by add inotify settings in sysctl

### DIFF
--- a/molecule/default/prepare.yml
+++ b/molecule/default/prepare.yml
@@ -20,6 +20,18 @@
       become: yes
       command: apt-get update
 
+    - name: Update sysctl settings for kubernetes
+      become: yes
+      ansible.builtin.sysctl:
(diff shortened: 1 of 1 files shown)


mnaser (changes requested):
I think this is better suited to be inside vexxhost/ansible-collection-kubernetes where we have a role that sets all the sysctls

PR #1575, opened 2024-07-12
Fix CLI issues

mnaser:
This fixes the ability to be able to run the CLI anywhere
other than /tmp and also exports the SHELL variable so that
the Magnum clients works happily.


fitbeard:
Nice!

PR #1576, opened 2024-07-12
Fix multipathd ppa install failure in non Ubuntu Jammy release

### LINKED ISSUE #1574: multipathd ppa only for Jammy
The current [PPA](https://github.com/vexxhost/atmosphere/blob/main/roles/multipathd/defaults/main.yml#L23) is only for Jammy.

Our current Computes are on Focal. Would be create if you also add the Jammy packages.

yaguangtang:
Closes #1574


mpiscaer:
LGTM

mnaser:
regate

PR #1601, opened 2024-07-16
[ATMOSPHERE-364] cert-manager: Add support for Azure DNS

gtirloni:


gtirloni:
@guilhermesteinmuller could you help test this?

guilhermesteinmuller (commented):
I am thinking now if we should support that. 

I had issues with DNSSEC: NSEC Missing: validation failure while attempting to issue the cert and it looks like azure dns doesnt support that...

mnaser:
recheck

mnaser:
regate

PR #1629, opened 2024-07-18
feat: update loki version

okozachenko1203:
apply a patch to support ruler sidecar in singleBinary mode

https://github.com/grafana/loki/pull/13572


mnaser:
@okozachenko1203 is this still necessary?

Issue #1630, opened 2024-07-18
bug: some logs are missing in loki


okozachenko1203:
typically, logs of `oauth2-proxy` container of alertmanager pod and `haproxy` of `percona-xtradb-haproxy` replicas are missing.

loki outputs this kinda error.
```sh
level=error ts=2024-07-18T13:54:15.969797889Z caller=manager.go:49 component=distributor path=write msg="write operation failed" details="entry for stream '{container_image=\"docker.io/percona/percona-xtradb-cluster-operator:1.13.0-haproxy\", container_name=\"haproxy\", node=\"instance\", pod_ip=\"10.0.0.111\", pod_label_app_kubernetes_io_component=\"haproxy\", pod_label_app_kubernetes_io_instance=\"percona-xtradb\", pod_label_app_kubernetes_io_managed_by=\"percona-xtradb-cluster-operator\", pod_label_app_kubernetes_io_name=\"percona-xtradb-cluster\", pod_label_app_kubernetes_io_part_of=\"percona-xtradb-cluster\", pod_label_apps_kubernetes_io_pod_index=\"0\", pod_label_controller_revision_hash=\"percona-xtradb-haproxy-78b9d6c4ff\", pod_label_statefulset_kubernetes_io_pod_name=\"percona-xtradb-haproxy-0\", pod_name=\"percona-xtradb-haproxy-0\", pod_namespace=\"openstack\", pod_uid=\"7dddbb4b-ec70-464c-a3ce-ab89276748f4\", source_type=\"kubernetes_logs\"}' has 16 label names; limit 15" org_id=fake
```

```sh
[...]

PR #1650, opened 2024-07-19
chore: limit pod labels in vector

okozachenko1203:
exclusive https://github.com/vexxhost/atmosphere/pull/1651

closes https://github.com/vexxhost/atmosphere/issues/1630

### DIFF
--- a/roles/vector/vars/main.yml
+++ b/roles/vector/vars/main.yml
@@ -30,7 +30,7 @@ _vector_helm_values:
       loki_kubernetes_logs:
         type: loki
         endpoint: http://loki-gateway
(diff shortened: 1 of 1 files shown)


gtirloni:
It seems like a lot of labels, why are they unncessary?

okozachenko1203:
> It seems like a lot of labels, why are they unnecessary?
> 
> nitpick: "chore" is for non-production changes, this will impact deployments.

@gtirloni i added the target issue and you can find the context

gtirloni:
Yes, I understand the issue this is trying to fix. I was hoping to understand why each of them is unnecessary though.

LGTM.

PR #1652, opened 2024-07-19
Change promethues to use pvc for data store

### LINKED ISSUE #352: Prometheus using emptydir instead of persistence

### LINKED ISSUE #1648: Prometheus data isn't using pv to store
Prometheus data can't suffer a host reboot or pod recreate without using PV, it's necessary to keep the monitoring history data for performance analysis and auditing.

yaguangtang:
Closes #352
Closes #1648


yaguangtang:
recheck

PR #1640, opened 2024-07-18
Add Goldpinger + node-exporter-full

mnaser:
Closes #1494 

cc @fnpanic

PR #1663, opened 2024-07-22
[stable/zed] Add Goldpinger + node-exporter-full (#1640)

okozachenko1203:
Closes https://github.com/vexxhost/atmosphere/issues/1647


roles/defaults/vars/main.yml
```diff
@@ -61,7 +61,8 @@ _atmosphere_images:
   glance_registry: "registry.atmosphere.dev/library/glance:{{ atmosphere_release }}"
   glance_storage_init: "registry.atmosphere.dev/library/glance:{{ atmosphere_release }}"
   grafana_sidecar: quay.io/kiwigrid/k8s-sidecar:1.26.1
-  grafana: docker.io/grafana/grafana:10.4.0
+  grafana: docker.io/grafana/grafana:11.0.0
```
mnaser, line 64: Upgrade here that shouldn't happen.

mnaser (changes requested):

PR #1675, opened 2024-07-23
fix(horizon): apply a patch to fix resource provider panel

okozachenko1203:
upstream patch: https://review.opendev.org/c/openstack/horizon/+/924729


mnaser (changes requested):
Curious why this was changed?

okozachenko1203:
> Curious why this was changed?

Not much infor in this question, but i guess you are curious why horizon git ref changed, right?

That is because i made the upstream commit onto master branch. To decrease any unexpected issues with new changes, i chose the last commit of the changed files instead of master head commit 
image
image

mnaser (approved):

PR #1676, opened 2024-07-24
[ATMOSPHERE-302] fix: set variables for cluster issuer name for keycloak and kube-prom-stack

okozachenko1203:
…-stack

### DIFF
--- a/roles/ingress/defaults/main.yml
+++ b/roles/ingress/defaults/main.yml
@@ -37,4 +37,4 @@ ingress_class_name: "{{ atmosphere_ingress_class_name }}"
 ingress_annotations: {}
 
 # List of annotations to apply to all Ingress resources as default
-ingress_default_annotations: "{{ atmosphere_ingress_annotations }}"
+ingress_default_annotations: "{{ ingress_global_annotations | default(atmosphere_ingress_annotations) }}"
--- a/roles/ingress/defaults/main.yml
+++ b/roles/ingress/defaults/main.yml
@@ -32,9 +32,9 @@ ingress_class_name: "{{ atmosphere_ingress_class_name }}"
 #               the ingress role to pass annotations to the
 #               Ingress resource.  You should use role-specific
 #               variable to pass annotations or use the
-#               ingress_global_annotations variable to apply
+#               ingress_default_annotations variable to apply
 #               annotations to all Ingress resources.
 ingress_annotations: {}
 
-# List of annotations to apply to all Ingress resources
-ingress_global_annotations: {}
+# List of annotations to apply to all Ingress resources as default
+ingress_default_annotations: "{{ atmosphere_ingress_annotations }}"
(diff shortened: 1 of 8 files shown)


gtirloni (dismissed):

mnaser (changes requested):
Can you find a way to integrate this variable into it?

https://github.com/vexxhost/atmosphere/blob/main/roles/ingress/defaults/main.yml#L39-L40

This will be needed to issue certificates successfully.

okozachenko1203:
> Can you find a way to integrate this variable into it?
> 
> https://github.com/vexxhost/atmosphere/blob/main/roles/ingress/defaults/main.yml#L39-L40
> 
> This will be needed to issue certificates successfully.

ok, updated

roles/kube_prometheus_stack/vars/main.yml
```diff
@@ -75,8 +75,7 @@ _kube_prometheus_stack_helm_values:
       enabled: true
       servicePort: 8081
       ingressClassName: "{{ kube_prometheus_stack_ingress_class_name }}"
-      annotations:
-        cert-manager.io/cluster-issuer: atmosphere
+      annotations: "{{ kube_prometheus_stack_ingress_annotations | combine(ingress_global_annotations, recursive=True) }}"
```
mnaser, line 78: This variable will not exist, I think you need to maybe move some of these into `vars` and then also make sure there is some form of defaults since it may not be defined.

mnaser (changes requested):

roles/kube_prometheus_stack/vars/main.yml
okozachenko1203: @mnaser ok, i see
btw, i think we can rename `ingress_global_annotations` as  `ingress_default annotations` which is a role-level variable.
And set its default value with a new global variable `atmosphere_ingress_annotations`.\

what do you think about this?

mnaser: If we do that, we have users that already depend on it, so we need to make sure that we allow users to use the old variable.

roles/ingress/defaults/main.yml
```diff
@@ -32,9 +32,9 @@ ingress_class_name: "{{ atmosphere_ingress_class_name }}"
...
 #               variable to pass annotations or use the
-#               ingress_global_annotations variable to apply
+#               ingress_default_annotations variable to apply
 #               annotations to all Ingress resources.
 ingress_annotations: {}
 
-# List of annotations to apply to all Ingress resources
-ingress_global_annotations: {}
+# List of annotations to apply to all Ingress resources as default
+ingress_default_annotations: "{{ atmosphere_ingress_annotations }}"
```
mnaser, line 40: We should maybe do this:

```
ingress_default_annotations: "{{ ingress_global_annotations | default(atmosphere_ingress_annotations) }}"
```

or maybe it's the other way around?

mnaser (changes requested):

roles/ingress/defaults/main.yml
okozachenko1203: i totally agree with you.

mnaser (approved):

PR #1689, opened 2024-07-25
[ATMOSPHERE-253] docs: Add how to access the cloud

gtirloni:


mnaser (changes requested):
This is not something that can only be used for the quick start, but it can be used for any way of "accessing" the cloud.

Could you maybe find a place in the documentation on how to access the cloud, and you can perhaps move the docs to access the cloud after standing up a test environment to point to it, and that ultimately becomes the "how to use cloud"? :)

Thanks

mnaser:
recheck

mnaser (approved):

PR #1716, opened 2024-07-30
[ATMOSPHERE-321] Fix run multipathd role failure

yaguangtang:
Enable gather facts for the multipathd role

### DIFF
--- a/playbooks/openstack.yml
+++ b/playbooks/openstack.yml
@@ -66,7 +66,6 @@
 - name: Configure operating system
   hosts: controllers:computes
   become: true
-  gather_facts: false
   roles:
     - role: lpfc
       tags:


playbooks/openstack.yml
```diff
@@ -66,7 +66,7 @@
 - name: Configure operating system
   hosts: controllers:computes
   become: true
-  gather_facts: false
+  gather_facts: true
```
mnaser, line 69: This is the default, so let's just remove the line :)

mnaser (changes requested):

playbooks/openstack.yml
yaguangtang: thanks

mnaser (approved):

mnaser (approved):

mnaser:
regate

PR #1717, opened 2024-07-30
[ATMOSPHERE-291] feat: Add support for OVS exporter

okozachenko1203:


mnaser (changes requested):
Can you please do multi-stage image builds of the source code, so use ADD ... with the git repo, and then build the Golang binary in one image, and then copy it to the other?

That will make it easier to pin to specific commits, and also add patches if needed.

mnaser (changes requested):
I think we should run the ovs exporter only on the ovs pods. The ovn exporter can come later to get ovn specific data.

okozachenko1203:
ovn passed.
ovs failed because of net conn issue
image

okozachenko1203:
recheck

okozachenko1203:
https://review.vexxhost.dev/c/atmosphere/+/203

PR #1745, opened 2024-08-07
[ATMOSPHERE-374] Enable Nova automatic host discovery at a set interval

yaguangtang:


yaguangtang:
recheck

yaguangtang:
recheck

roles/nova/vars/main.yml
```diff
@@ -107,6 +107,7 @@ _nova_helm_values:
         ovsdb_connection: unix:/run/openvswitch/db.sock
       scheduler:
         workers: 8
+        discover_hosts_in_cells_interval: 300
```
mnaser, line 110: Let's switch this to 30, it makes more sense in our environments and it will appear faster.

mnaser (changes requested):

roles/nova/vars/main.yml
yaguangtang: OK

mnaser (approved):

mnaser:
recheck

mnaser:
regate

mnaser:
regate

PR #1798, opened 2024-08-19
[ATMOSPHERE-394] 2023 2 bobcat allow for smoother ingress nginx upgrades

yaguangtang:
Set the wait up from 300s to 10s for the drain of ingress connections. Set nginx worker-shutdown-timeout to 5s from default 240s.


yaguangtang:
regate

yaguangtang:
regate

PR #1805, opened 2024-08-21
[ATMOSPHERE-405] fix: wait until keycloak ready

okozachenko1203:
At the moment, it seems that we’re not waiting for Keycloak to finish rolling out before we move forward, so the migration can be still be running.


mnaser:
regate

mnaser:
regate

Issue #1804, opened 2024-08-20
Issue with encrypted volumes be create from encrypted volume types


thywyn:
So, the volume type gets created as the script sets it up.  The volumes created are failing to create.  Here's the code block the helm template creates:
20240820_08:42:40 $ kubectl get  cm -n openstack cinder-bin -o jsonpath={'.data.bootstrap\.sh'}
#!/bin/bash

set -ex
export HOME=/tmp
      # Create a volume type if it doesn't exist.
      # Assumption: the volume type name is unique.
      openstack volume type show rbd1-encrypted || \
      openstack volume type create \
        --public \
      rbd1-encrypted
      
      is_in_use=$(openstack volume list --long --all-projects -c Type -f value | grep -E "^rbd1-encrypted\s*$" || true)
      if [[ -z ${is_in_use} ]]; then
        openstack volume type set --property encryption-cipher=aes-xts-plain64 rbd1-encrypted
        openstack volume type set --property encryption-control-location=front-end rbd1-encrypted
        openstack volume type set --property encryption-key-size=256 rbd1-encrypted
        openstack volume type set --property encryption-provider=luks rbd1-encrypted
        openstack volume type set --property volume_backend_name=rbd1 rbd1-encrypted
      fi
        openstack volume type show rbd1 || \
        openstack volume type create \
        --public \
        --property volume_backend_name=rbd1 \
        rbd1
openstack volume type list --long
openstack volume qos list

exit 0
______

I run this experiment based on the script:

openstack volume type create rbd1-way-1
openstack volume type set --property encryption-cipher=aes-xts-plain64 rbd1-way-1
openstack volume type set --property encryption-control-location=front-end rbd1-way-1
openstack volume type set --property encryption-key-size=256 rbd1-way-1
openstack volume type set --property encryption-provider=luks rbd1-way-1
openstack volume type set --property volume_backend_name=rbd1 rbd1-way-1

openstack volume type create rbd1-way-2
openstack volume type set --encryption-provider=luks rbd1-way-2
openstack volume type set --encryption-cipher=aes-xts-plain64 rbd1-way-2
openstack volume type set --encryption-control-location=front-end rbd1-way-2
openstack volume type set --encryption-key-size=256 rbd1-way-2
openstack volume type set --property volume_backend_name=rbd1 rbd1-way-2

openstack volume create --type rbd1-way-1 --size 1 volume-way-1
openstack volume create --type rbd1-way-2 --size 1 volume-way-2

[...]

PR #1807, opened 2024-08-21
[CLOUDOPS-344] [stable/2024.1] fix: bump cinder chart and add bootstra…

ricolin:
relate to https://github.com/vexxhost/atmosphere/issues/1804


mnaser:
@ricolin the linters will fail here if the patch is not included.

ricolin:
> @ricolin the linters will fail here if the patch is not included.

Yeah, was leave it back at my local repo :)

PR #1809, opened 2024-08-21
[ATMOSPHERE-410] chore(deps): bump ceph collection to  fix ceph metrics issue

yaguangtang:


galaxy.yml
```diff
@@ -16,7 +16,7 @@ dependencies:
   community.mysql: 3.6.0
   kubernetes.core: 2.4.0
   openstack.cloud: 1.7.0
-  vexxhost.ceph: 3.0.1
+  vexxhost.ceph: 3.0.2
```
mnaser, line 19: Let's move towards `>=3.0.2` instead, we want to move away from pinned since it is causing issues when trying to update things.

mnaser (changes requested):

yaguangtang:
recheck

PR #1822, opened 2024-08-22
feat: specify api microversion for nova client in magnum conf

okozachenko1203:
2.15 is minimum version to support soft-* server group policy rule


mnaser (approved):
2.12 is the max which is liberty.  2.25 is mitaka, safe to say, we're good.

https://docs.openstack.org/nova/latest/reference/api-microversion-history.html

PR #1845, opened 2024-08-30
[ATMOSPHERE-432] Fix goldpinger grafana dashboard threshold for nodes…

yaguangtang:
Remove threshold for nodes in grafana dashboard which may cause end user confuseing


yaguangtang:
recheck

PR #1871, opened 2024-09-10
Add optional kubeconfig path to roles

austintalbot:


fitbeard:
I see potential here for the future if we need access to remote clusters. For example for multi region "infrastructure" deployment.

roles/barbican/tasks/main.yml
```diff
@@ -19,7 +19,7 @@
     chart_ref: "{{ barbican_helm_chart_ref }}"
     release_namespace: "{{ barbican_helm_release_namespace }}"
     create_namespace: true
-    kubeconfig: /etc/kubernetes/admin.conf
+    kubeconfig: "{{ kubeconfig_path | default('/etc/kubernetes/admin.conf') }}"
```
mnaser, line 22: `ansible-lint` won't like this in the long term, because it's not using a role variable.

Would you be able to add something to the `defaults/main.yml` for each role (barbican in this example):

```yaml
barbican_helm_kubeconfig: "{{ kubeconfig_path | default('/etc/kubernetes/admin.conf') }}"
```

and then...

```suggestion
    kubeconfig: "{{ barbican_helm_kubeconfig }}"
```

This would allow users that would for some reason want to use different `kubeconfig` for different roles (e..g. multi-region as @fitbeard described) possible.

mnaser (changes requested):
Small revision!

roles/barbican/tasks/main.yml
austintalbot: Updated all of the defaults/main.yaml and files respectively.

mnaser:
recheck

roles/coredns/tasks/main.yml
```diff
@@ -20,5 +20,5 @@
     chart_ref: "{{ coredns_helm_chart_ref }}"
     release_namespace: "{{ coredns_helm_release_namespace }}"
     create_namespace: true
-    kubeconfig: /etc/kubernetes/admin.conf
+    kubeconfig: "{{ cordens_kubeconfig }}"
```
mnaser, line 23: I think this is a typo and might be causing the failure.

mnaser (changes requested):
small nit!

mnaser:
@austintalbot sorry, I was out of the office because of a personal emergency, in order to avoid a bunch of back and forth, I've made a small amendment in order to be able to fit better the existing variable naming convention.

mnaser (approved):

PR #1872, opened 2024-09-11
Update quick start AIO instructions

jizaymes:
Update minimum requirements for AIO, and update quick start instructions to imply root user (#)

### DIFF
--- a/doc/source/quick-start.rst
+++ b/doc/source/quick-start.rst
@@ -35,13 +35,6 @@ and all the OpenStack services inside a single machine.
 In order to get started, you'll need a **Ubuntu 22.04** system with the
 following minimum system requirements:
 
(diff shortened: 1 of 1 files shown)


doc/source/quick-start.rst
```diff
@@ -53,35 +46,34 @@ is always better!):
     otherwise the performance of the VMs will be un-usable.
 
 You'll need to start by installing all of the necessary dependencies first,
-**you also need to make sure you run all of these commands as ``root``**:
+**you also need to make sure you run all of these commands as root**:
 
 .. code-block:: console
 
-    $ sudo -i
```
ricolin, line 60: I think we should keep this to make sure we have environment vars loaded.

PR #1873, opened 2024-09-12
[ATMOSPHERE-543] [stable/zed] log warning when failed to load pkcs12 cert 

okozachenko1203:
https://opendev.org/openstack/octavia/commit/0ad935c7aa679f764d11d692c613aee939a508bc this commit was not backported to zed in upstream because it is unmaintained


mnaser:
regate

PR #1877, opened 2024-09-16
feat: support disabling sso

okozachenko1203:
by setting keystone_domains as emtpy

fix https://github.com/vexxhost/atmosphere/issues/1878


mnaser (changes requested):
I don't think we really want people to use native SSO out of the box, Keystone is a headache to deal with when it comes to this, so if someone wants to override this, they can using Helm configuration values (and I don't really want to document this so it doesn't become some form of "supported" option).

PR #1932, opened 2024-10-01
[stable/zed] [CLOUDOPS-405] fix: revert cinder volume direct image copy

yaguangtang:
This revert two patches from cinder image as they limit the use case to narrow scope. Revert until we have more proper approach for directly copy image and encrypt volume in RBD.


ricolin:
fix https://github.com/vexxhost/atmosphere/issues/1924

PR #1943, opened 2024-10-02
[ATMOSPHERE-466] Support loki log rotation

### LINKED ISSUE #525: Add log rotation for Loki
We need to setup a log rotation in place to setup a tail like 365d logging or so.

yaguangtang:
Enable loki log rotation by default, the default retention period is 30 days.

Fixes #525


mnaser:
recheck

PR #1948, opened 2024-10-02
[ATMOSPHERE-455] fix: disable ceph exporter managed by rook operator (#1896)

yaguangtang:
Currently we are using external ceph cluster, so ceph exporter service managed by rook operator is useless to collect k8s ceph metrics.

Reviewed-by: Mohammed Naser <mnaser@vexxhost.com>


mnaser:
recheck

yaguangtang:
recheck

PR #1977, opened 2024-10-11
[ATMOSPHERE-488] fix: sync pxc crd version with image ref

okozachenko1203:


fitbeard:
Would be reasonable combine this together: https://github.com/vexxhost/atmosphere/pull/1211

PR #1979, opened 2024-10-11
[ATMOSPHERE-485] update magnum images to  1.29.6 1.30.2 1.31.1

yaguangtang:
also update molecule magnum to use latest 1.31.1 k8s image

### DIFF
--- a/molecule/aio/group_vars/all/molecule.yml
+++ b/molecule/aio/group_vars/all/molecule.yml
@@ -214,10 +214,7 @@ magnum_helm_values:
       api: 1
       conductor: 1
 magnum_image_disk_format: qcow2
-magnum_images:
-  - name: ubuntu-2204-kube-v1.27.8s
-    url: https://object-storage.public.mtl1.vexxhost.net/swift/v1/a91f106f55e64246babde7402c21b87a/magnum-capi/ubuntu-2204-kube-v1.27.8.qcow2
-    distro: ubuntu
+magnum_images: "[ {{ _magnum_images[-1] }} ]"
 
 manila_helm_values:
   conf:
--- a/roles/magnum/defaults/main.yml
+++ b/roles/magnum/defaults/main.yml
@@ -43,16 +43,7 @@ magnum_clusterctl_config:
 # List of images to load into OpenStack for Magnum
 magnum_image_container_format: bare
 magnum_image_disk_format: raw
-magnum_images:
-  - name: ubuntu-2204-kube-v1.25.11
-    url: https://object-storage.public.mtl1.vexxhost.net/swift/v1/a91f106f55e64246babde7402c21b87a/magnum-capi/ubuntu-2204-kube-v1.25.11.qcow2
-    distro: ubuntu
-  - name: ubuntu-2204-kube-v1.26.6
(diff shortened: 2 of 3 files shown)


roles/magnum/defaults/main.yml
```diff
@@ -44,14 +44,14 @@ magnum_clusterctl_config:
...
-  - name: ubuntu-2204-kube-v1.26.6
-    url: https://object-storage.public.mtl1.vexxhost.net/swift/v1/a91f106f55e64246babde7402c21b87a/magnum-capi/ubuntu-2204-kube-v1.26.6.qcow2
+  - name: ubuntu-2204-kube-v1.29.6
+    url: https://static.atmosphere.dev/artifacts/magnum-cluster-api/ubuntu-jammy-kubernetes-1-29-6-1720107687.qcow2
     distro: ubuntu
-  - name: ubuntu-2204-kube-v1.27.3
-    url: https://object-storage.public.mtl1.vexxhost.net/swift/v1/a91f106f55e64246babde7402c21b87a/magnum-capi/ubuntu-2204-kube-v1.27.3.qcow2
+  - name: ubuntu-2204-kube-v1.28.11
+    url: https://static.atmosphere.dev/artifacts/magnum-cluster-api/ubuntu-jammy-kubernetes-1-28-11-1720107685.qcow2
     distro: ubuntu
```
mnaser, line 55: Can you actually create in vars.yml a variable called `_magnum_images` with those values, and then set the default of `magnum_images` to `_magnum_images`.

Then, inside the Molecule configuration, instead of having to sync the `magnum_images` all the time, perhaps, you can just make it set to `_magnum_images | last`

Also would be nice in that case to just order them by old to new :)

molecule/aio/group_vars/all/molecule.yml
```diff
@@ -214,10 +214,7 @@ magnum_helm_values:
       api: 1
       conductor: 1
 magnum_image_disk_format: qcow2
-magnum_images:
-  - name: ubuntu-2204-kube-v1.27.8s
-    url: https://object-storage.public.mtl1.vexxhost.net/swift/v1/a91f106f55e64246babde7402c21b87a/magnum-capi/ubuntu-2204-kube-v1.27.8.qcow2
-    distro: ubuntu
+magnum_images: "{{ _magnum_images | last }}"
```
mnaser, line 217: This needs to be a list and then you simply have the single item in there be the the last one.

mnaser (dismissed):

molecule/aio/group_vars/all/molecule.yml
yaguangtang: in the magnum tasks, we upload images , need it to be an list of dict. so I update to _magnum_images[-1]
- name: Upload images
  ansible.builtin.include_role:
    name: glance_image
  loop: "{{ magnum_images }}"
  vars:
    glance_image_name: "{{ item.name }}"
    glance_image_url: "{{ item.url }}"
    glance_image_container_format: "{{ magnum_image_container_format }}"
    glance_image_disk_format: "{{ magnum_image_disk_format }}"
    glance_image_properties:
      os_distro: "{{ item.distro }}"

yaguangtang:
regate

PR #1983, opened 2024-10-16
[ATMOSPHERE-533] ci: update go unit tests

okozachenko1203:
- remove 1.22 from `TestKubeconform` test (`node-feature-discovery` uses grpc in probe which is available since 1.22)
- add a workaround values in `TestKubeconform` for loki chart
- skip `CephFilesystemSubVolumeGroup` kind for ceph chart
- fix `oslo_db` test expected values
- update `TestImageExist` test to support image tags without digest
- fix pxc test codes
- supplement placement chart value for database max_retries param

depends-on: https://github.com/vexxhost/atmosphere/pull/1977 
depends-on: https://github.com/vexxhost/atmosphere/pull/1211
depends-on: https://github.com/vexxhost/atmosphere/pull/1987 https://github.com/vexxhost/atmosphere/pull/1211 #1977

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -1,4 +1,11 @@
 - job:
+    name: atmosphere-unit-test
+    parent: golang-go
+    vars:
+      go_version: 1.21.13
+      go_command: test ./... -v
+
+- job:
     name: atmosphere-chart-vendor
     parent: chart-vendor
 
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -1,8 +1,14 @@
 - job:
-    name: atmosphere-unit-test
+    name: atmosphere-golang-go
     parent: golang-go
+    abstract: true
     vars:
       go_version: 1.21.13
+
+- job:
+    name: atmosphere-golang-go-test
+    parent: atmosphere-golang-go
+    vars:
       go_command: test ./... -v
 
 - job:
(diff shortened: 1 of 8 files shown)


mnaser:
recheck

zuul.d/jobs.yaml
```diff
@@ -1,3 +1,10 @@
+- job:
+    name: atmosphere-unit-test
+    parent: golang-go
+    vars:
+      go_version: 1.21.13
+      go_command: test ./... -v
```
mnaser, line 6: Can you split this into two jobs?

One called `atmosphere-golang-go` with the `go_version` variable in there.  Then, another one called `atmosphere-golang-go-test` which the parent of the is `atmosphere-golang-go` which has the test there.  :)

mnaser (changes requested):

zuul.d/jobs.yaml
okozachenko1203: well, it means the first one should be an abstract?

mnaser: That's correct. :)

mnaser (approved):

okozachenko1203:
cherry-pick stable/2024.1

okozachenko1203:
cherry-pick stable/2023.2

okozachenko1203:
cherry-pick stable/2023.1

okozachenko1203:
cherry-pick stable/zed

PR #1986, opened 2024-10-18
doc: add doc how to send logs to syslog server in vector

okozachenko1203:


mnaser (changes requested):
If you can add this into the `docs` section since that's part of the Sphinx published docs.

PR #2010, opened 2024-10-23
[ATMOSPHERE-493] chart: bump ovn chart version and add liveness check for northd

okozachenko1203:
because required the upstream patch rebased and also new versions include some fixes, i updated ovn chart version first and added liveness check in the patch for introducing kube-ovn


charts/ovn/templates/daemonset-controller.yaml
```diff
@@ -55,19 +81,36 @@ spec:
...
+              kubectl get node ${NODENAME} -o jsonpath='{.metadata.labels.l3-agent}' > /tmp/gw-enabled/gw-enabled
+          env:
+            - name: NODENAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+          volumeMounts:
+            - name: gw-enabled
+              mountPath: /tmp/gw-enabled
+              readOnly: false
```
mnaser, line 106: This breaks our current process, since I don't think we use `l3-agent` label to decide what's gateway and what's not..

mnaser (changes requested):

charts/ovn/templates/daemonset-controller-gw.yaml
```diff
@@ -66,8 +66,6 @@ spec:
           command:
             - /tmp/ovn-controller-init.sh
           env:
-            - name: OVN_CMS_OPTIONS
-              value: {{ .Values.conf.gw_ovn_cms_options | quote }}
```
mnaser: I think we need to restore this.

charts/ovn/templates/bin/_ovn-controller-init.sh.tpl
```diff
@@ -144,13 +144,20 @@ ovs-vsctl set open . external-ids:rundir="/var/run/openvswitch"
...
 ovs-vsctl set open . external-ids:ovn-bridge="{{ .Values.conf.ovn_bridge }}"
 ovs-vsctl set open . external-ids:ovn-bridge-mappings="{{ .Values.conf.ovn_bridge_mappings }}"
-ovs-vsctl set open . external-ids:ovn-cms-options="${OVN_CMS_OPTIONS}"
+
+GW_ENABLED=$(cat /tmp/gw-enabled/gw-enabled)
+if [[ ${GW_ENABLED} == enabled ]]; then
+  ovs-vsctl set open . external-ids:ovn-cms-options={{ .Values.conf.ovn_cms_options_gw_enabled }}
+else
+  ovs-vsctl set open . external-ids:ovn-cms-options={{ .Values.conf.ovn_cms_options }}
+fi
```
mnaser: This should be gone now since we have it directly in the DS, so we should revert back to how it was before.

charts/ovn/values.yaml
```diff
@@ -37,16 +38,16 @@ images:
...
   ovn_northd:
-    node_selector_key: openstack-compute-node
+    node_selector_key: openstack-network-node
     node_selector_value: enabled
   ovn_controller:
-    node_selector_key: openstack-compute-node
+    node_selector_key: openvswitch
     node_selector_value: enabled
   ovn_controller_gw:
     node_selector_key: openstack-control-plane
```
mnaser, line 53: This will break existing deployments I believe.  We don't have `openstack-network-node` label on these systems.   Also, technically, ovsdb is just db and northd is just a control plane service, so i hink they should live on the control plane

also, this will now run ovn gw and normal ovn on control plane because both ds match, so we should keep labels as is.

charts/ovn/templates/daemonset-controller.yaml
```diff
@@ -85,12 +85,12 @@ spec:
...
-          command:
-            - /root/ovnkube.sh
-            - ovn-controller
 {{ tuple $envAll "ovn_controller" | include "helm-toolkit.snippets.image" | indent 10 }}
 {{ tuple $envAll $envAll.Values.pod.resources.server | include "helm-toolkit.snippets.kubernetes_resources" | indent 10 }}
 {{ dict "envAll" $envAll "application" "ovn_controller" "container" "controller" | include "helm-toolkit.snippets.kubernetes_container_security_context" | indent 10 }}
+          command:
+            - /root/ovnkube.sh
+            - ovn-controller
 {{ dict "envAll" . "component" "ovn_controller" "container" "controller" "type" "readiness" "probeTemplate" (include "controllerReadinessProbeTemplate" . | fromYaml) | include "helm-toolkit.snippets.kubernetes_probe" | indent 10 }}
```
mnaser, line 94: Can we avoid this change to minimize the delta nad make backporting easier?

mnaser (changes requested):
Let's try to get this in shape with those changes please :)

charts/ovn/templates/bin/_ovn-controller-init.sh.tpl
okozachenko1203: this is already fixed. i think during your review, i made a new commit

charts/ovn/templates/daemonset-controller-gw.yaml
okozachenko1203: this is already fixed. i think during your review, i made a new commit

charts/ovn/values.yaml
okozachenko1203: it should not affect us. I set them in ovn roles variable because upstream doesn't accept this.

charts/ovn/templates/daemonset-controller.yaml
okozachenko1203: well, the upstream chart keeps command in that place. So i thought this change is more near to the upstream

mnaser (changes requested):

mnaser (dismissed):

PR #2019, opened 2024-10-24
[ATMOSPHERE-503] fix: remove softnet squeeze rules in kube-prometheus-stack

okozachenko1203:


mnaser (changes requested):
I think I'm not too happy with the existing system that replies on recording rules because it used the count instead of the actual value. 

Could you drop the recording rules and simply make it alert on the actual metric instead in the alert?

mnaser (approved):
It'll do for now.

mnaser:
regate

mnaser:
regate

mnaser:
regate

PR #2024, opened 2024-10-25
[ATMOSPHERE-498]add user and group to ovn container image to rebuild for dpdk file ac…

### LINKED ISSUE #1982: ovnkube.sh resets all permissions in /run/openvswitch to root:root by default
We are running into issues where the /run/openvswitch/ directory is  getting reset to root:root after ovn-controller pods runs  in the ovnkube.sh startup script.  neuton-ovn-metadata-agent init scripts only resets the control and socket files back to 42424:42424.. So, VMs building with DPDK interfaces fail with errors like the following "libvirt.libvirtError: internal error: process exited while connecting to monitor: 2024-10-14T14:30:53.518766Z qemu-system-x86_64: -chardev socket,id=charnet0,path=/var/run/openvswitch/vhua820f710-65,server=on: Failed to bind socket to /var/run/openvswitch/vhua820f710-65: Permission denied"

ovn-controller pods do have an environment variable that  could be used "OVS_USER_ID=...."  but the 42424 user is not defined on the container.  Without rebuilding the container with that user, the only other way I can think of is to expand the scope of the chown that neuton-ovn-metadata-agent init scripts does.

yaguangtang:
fix #1982

### DIFF
--- a/images/ovn/Dockerfile
+++ b/images/ovn/Dockerfile
@@ -34,3 +34,11 @@ EOF
 COPY --from=ovn-kubernetes --link /src/dist/images/ovndb-raft-functions.sh /root/ovndb-raft-functions.sh
 COPY --from=ovn-kubernetes --link /src/dist/images/ovnkube.sh /root/ovnkube.sh
 COPY --from=ovn-kubernetes --link /usr/bin/ovn-kube-util /usr/bin/ovn-kube-util
+
+ARG PROJECT=ovn
+ENV OVS_USER_ID=42424
+RUN \
+    groupadd -g 42424 ${PROJECT} && \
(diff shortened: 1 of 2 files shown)


images/ovn/Dockerfile
```diff
@@ -34,3 +34,10 @@ EOF
 COPY --from=ovn-kubernetes --link /src/dist/images/ovndb-raft-functions.sh /root/ovndb-raft-functions.sh
 COPY --from=ovn-kubernetes --link /src/dist/images/ovnkube.sh /root/ovnkube.sh
 COPY --from=ovn-kubernetes --link /usr/bin/ovn-kube-util /usr/bin/ovn-kube-util
+
+ENV OVS_USER_ID=42424
+ONBUILD RUN \
```
mnaser, line 39: I think this should be `RUN`.  It doesn't seem like the CI ran it.

mnaser (changes requested):

mnaser (approved):

PR #2042, opened 2024-10-28
[ATMOSPHERE-514] fix ovn for backward compatibility

okozachenko1203:
we used to keep `type: hv|gw` labels in ovn-controller and ovn-controller-gw daemonsents to distinguish them.
in the PR #2023, we removed ovn-controller-gw daemonset and keep only one DS to sync with upstream as much as possible. so `type: xx` label is unnecessary. But we cannot remove it for running clouds because `spec.selector` field is immutable.

this pr reverts that label for backward compatibility.

### DIFF
--- a/roles/ovn/tasks/main.yml
+++ b/roles/ovn/tasks/main.yml
@@ -27,7 +27,7 @@
       failed_when: false
 
     - name: Check if ovn controller has type label in spec.selector
-      set_fact:
+      ansible.builtin.set_fact:
         _ovn_controller_replace: "{{ 'type' in _ovn_controller_ds_info.resources[0].spec.selector.matchLabels }}"
       when: _ovn_controller_ds_info.resources | length > 0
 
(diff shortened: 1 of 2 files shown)


okozachenko1203:
recheck

roles/ovn/tasks/main.yml
```diff
@@ -12,6 +12,35 @@
...
+
+    - name: Delete existing ovn controller DaemonSet if type label is found
+      kubernetes.core.k8s:
+        api_version: apps/v1
+        kind: DaemonSet
+        name: ovn-controller
+        namespace: "{{ ovn_helm_release_namespace }}"
+        state: absent
+        kubeconfig: "{{ ovn_helm_kubeconfig }}"
+      when: _ovn_controller_replace | default(false)
```
mnaser, line 42:
```suggestion
      when:
        - _ovn_controller_ds_info.resources | length > 0
        - 'type' in _ovn_controller_ds_info.resources[0].spec.selector.matchLabels
```

What do you think of doing this and removing the set_fact?

mnaser (approved):

PR #2044, opened 2024-10-28
add non root ovs-user-id to ovn container image to rebuild 

### LINKED ISSUE #1982: ovnkube.sh resets all permissions in /run/openvswitch to root:root by default

yaguangtang:
update ovn-controler script to chown ovs bridge socket files so that libvirt pod can read

fix #1982

### DIFF
--- a/charts/ovn/templates/bin/_ovn-controller-init.sh.tpl
+++ b/charts/ovn/templates/bin/_ovn-controller-init.sh.tpl
@@ -175,5 +175,8 @@ do
   then
     ovs-vsctl --may-exist add-port $bridge $iface
     migrate_ip_from_nic $iface $bridge
+  # update bridge socket file to non root owner 42424
+  chown 42424:42424 /var/run/openvswitch/*.mgmt
+  chown 42424:42424 /var/run/openvswitch/*.snoop
   fi
 done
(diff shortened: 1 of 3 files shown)


mnaser (changes requested):
@yaguangtang I think we need to use this instead:

https://github.com/apconole/ovn-kubernetes/blob/master/dist/images/ovnkube.sh#L95-L96

We should add environment variable `OVS_USER_ID` and set to `42424` with a new image, so we don't need to make a chart change.

yaguangtang:
> @yaguangtang I think we need to use this instead:
> 
> https://github.com/apconole/ovn-kubernetes/blob/master/dist/images/ovnkube.sh#L95-L96
> 
> We should add environment variable `OVS_USER_ID` and set to `42424` with a new image, so we don't need to make a chart change.

It's added to ovn image

PR #2087, opened 2024-11-01
[ATMOSPHERE-558] chart/nova: update to the latest version

okozachenko1203:
`0.3.27` -> `0.3.46`
- includes a bug fix for dealing with null value of `log_config_append`
- includes several features, not used for now
- remove local patch to role vars


charts/nova/templates/deployment-novncproxy.yaml
```diff
@@ -144,9 +145,14 @@ spec:
               readOnly: true
             - name: pod-shared
               mountPath: /tmp/pod-shared
+            {{- if $vencrypt_enabled }}
+            - name: {{ .Values.secrets.tls.compute_novnc_proxy.vencrypt.internal }}
+              mountPath: /etc/pki/nova-novncproxy
+              readOnly: true
+            {{- end }}
+
```
mnaser, line 153: This seems to have changed, I don't think this works for us or does it?

charts/nova/templates/deployment-novncproxy.yaml
```diff
@@ -144,9 +145,14 @@ spec:
...
 {{- dict "enabled" .Values.manifests.certificates "name" .Values.endpoints.oslo_db.auth.admin.secret.tls.internal "path" "/etc/mysql/certs" | include "helm-toolkit.snippets.tls_volume_mount" | indent 12 }}
 {{- dict "enabled" .Values.manifests.certificates "name" .Values.secrets.tls.compute_novnc_proxy.novncproxy.internal "path" "/etc/nova/certs" | include "helm-toolkit.snippets.tls_volume_mount" | indent 12 }}
-{{- dict "enabled" $vencrypt_enabled "name" .Values.secrets.tls.compute_novnc_proxy.vencrypt.internal "path" "/etc/pki/nova-novncproxy" | include "helm-toolkit.snippets.tls_volume_mount" | indent 12 }}
```
mnaser, line 149: Seems like this is what it was replaced from.

mnaser (changes requested):
https://github.com/vexxhost/atmosphere/blob/f0e5466870c0f7b0ca6eb4bea0c425bc5f2948d3/roles/neutron/vars/main.yml#L72

I think we can drop that then too.

charts/nova/templates/deployment-novncproxy.yaml
okozachenko1203: i think it will work and be better because it doesn't use subpath

charts/nova/templates/deployment-novncproxy.yaml
okozachenko1203: i think it will work and be better because it doesn't use subpath

mnaser:
regate

mnaser:
recheck

mnaser (approved):

yaguangtang:
recheck

PR #2089, opened 2024-11-04
[CLOUDOPS-563]  Enable Emulated-TPM

larainema:
update the nova image include emulated-tpm packages 
update the nova compute conf to enable emulated-tpm


larainema:
recheck

PR #2103, opened 2024-11-08
feat: tempest generate subunit reports

mpiscaer:
generates subunit and junit reports from the tempest tests


mpiscaer:
I still need to figure out how to best export the reports from the pod

PR #2110, opened 2024-11-12
[ATMOSPHERE-567] allow failover on ERROR Amphora

ricolin:
add health_manager.failover_on_error config to allow running failover when an Amphora is expired with heartbeats and in ERROR status. ref: https://review.opendev.org/c/openstack/octavia/+/934638


ricolin:
let's wait for https://review.opendev.org/c/openstack/octavia/+/934638 review for now

PR #2114, opened 2024-11-12
[ATMOSPHERE-570][stable/zed] allow failover on ERROR Amphora

ricolin:
add health_manager.failover_on_error config to allow running failover when an Amphora is expired with heartbeats and in ERROR status. ref: https://review.opendev.org/c/openstack/octavia/+/934638


ricolin:
Let's wait for review finish on https://review.opendev.org/c/openstack/octavia/+/934638 and https://github.com/vexxhost/atmosphere/pull/2110 before new PR update here

PR #2120, opened 2024-11-13
[ATMOSPHERE-585] support SQLAlchemy 2.0 and fix the number of max active fernet keys in Keystone

okozachenko1203:
Depends-On: https://github.com/vexxhost/atmosphere/pull/2133


yaguangtang:
regate

PR #2150, opened 2024-11-26
upgrade mcapi, k8s collection and capo version

okozachenko1203:
- mcapi 0.22.1 -> 0.24.0 https://github.com/vexxhost/magnum-cluster-api/releases/tag/v0.24.0
- k8s collection 0.14.3 -> 0.15.0 https://github.com/vexxhost/ansible-collection-kubernetes/releases/tag/v1.15.0


mnaser:
recheck

mnaser:
@okozachenko1203 i think we need to probably change the image tags

okozachenko1203:
image
keycloak is unhealthy in CI

mnaser:
It seems the CAPI controller is still 1.6.0 and it seems we're not even using the image we build for CAPO.

@okozachenko1203 I think maybe we should remove our build and go upstream only, so we can remove the extra job that's a bit useless :p

PR #2151, opened 2024-11-27
[ATMOSPHERE-367] Add the NodeTimeSkewDetected alert

larainema:


roles/kube_prometheus_stack/files/jsonnet/vendor/github.com/prometheus/node_exporter/docs/node-mixin/alerts/alerts.libsonnet
mnaser, line 1: https://www.robustperception.io/time-metric-from-the-node-exporter/

Based on this, I think we need to maybe use take the advice there?

mnaser (changes requested):

mnaser:
recheck

yaguangtang:
recheck

mnaser:
recheck

mnaser:
recheck

mnaser (approved):

roles/kube_prometheus_stack/files/jsonnet/tests.yml
```diff
@@ -72,3 +72,23 @@ tests:
...
+      - eval_time: 5m
+        alertname: NodeTimeSkewDetected
+        exp_alerts:
+          - exp_labels:
+              severity: P3
+              instance: instance3
+              job: node
+            exp_annotations:
+              summary: "Node instance3 has a time difference."
+              description: "Node instance3 has a time difference 2."
```
mnaser, line 94: Actually, can you make a test here also for if time is sync'd to make sure it doesnt trigger if time is syncd?

mnaser (changes requested):

mnaser (approved):

PR #2153, opened 2024-11-28
set clusterctl version for capi explicitly

okozachenko1203:


mnaser (changes requested):
Will need a galaxy.yml requirements bump as well to the next major release.

PR #2182, opened 2024-12-06
[ATMOSPHERE-165] fix manila share mount use device name issue

### LINKED ISSUE #1065: Manila relies on device name from Nova (which is not reliable)
For some reason, if there are volumes that fail to attach properly, you will end up in a scenario where the volumes will fail to attach properly because the expected `/dev/vdX` does not match the correct one.

We should make Manila try and detect `/dev/disk/by-id` and rely on the `virtio` first, at least initially, then fallback to the Nova path:

```
manila@ubuntu:~$ ls -al /dev/disk/by-id
total 0
drwxr-xr-x 2 root root  60 Apr  2 20:58 .
drwxr-xr-x 7 root root 140 Apr  2 20:57 ..
lrwxrwxrwx 1 root root   9 Apr  2 20:58 virtio-a9f53ad5-f359-46b9-8 -> ../../vdc
```

yaguangtang:
fix #1065 

Upstream work https://review.opendev.org/c/openstack/manila/+/937190


yaguangtang:
recheck

mnaser:
this is already merged/fixed.

PR #2184, opened 2024-12-06
[ATMOSPHERE-606] replace value of dnsmasq_dns_servers to 1.1.1.1

adrianreza:
The bug was detected when customer env use OVN and they got 10.96.0.20 for nameserves from DHCP Agent. We can unset in on [atmosphere/roles/neutron/vars/main.yml at main · vexxhost/atmosphere](https://github.com/vexxhost/atmosphere/blob/main/roles/neutron/vars/main.yml) line 63 or replace that value with public DNS, e.g. 1.1.1.1.

### DIFF
--- a/roles/neutron/vars/main.yml
+++ b/roles/neutron/vars/main.yml
@@ -58,7 +58,6 @@ __neutron_helm_values:
         driver: noop
       service_providers:
         service_provider: VPN:strongswan:neutron_vpnaas.services.vpn.service_drivers.ipsec.IPsecVPNDriver:default
-      neutron_default_dns_servers: 10.96.0.20
     dhcp_agent:
       DEFAULT:
         dnsmasq_dns_servers: "{{ neutron_coredns_cluster_ip | default('10.96.0.20') }}"
@@ -94,10 +93,6 @@ __neutron_ovn_helm_values:
         ovn_emit_need_to_frag: true
       service_providers:
         service_provider: VPN:strongswan:neutron_vpnaas.services.vpn.service_drivers.ovn_ipsec.IPsecOvnVPNDriver:default
-    dhcp_agent:
-      DEFAULT:
-        dnsmasq_dns_servers: "{{ neutron_coredns_cluster_ip | default('1.1.1.1') }}"
-        enable_isolated_metadata: true
     ovn_metadata_agent:
       DEFAULT:
         metadata_proxy_shared_secret: "{{ openstack_helm_endpoints['compute_metadata']['secret'] }}"
--- a/roles/neutron/vars/main.yml
+++ b/roles/neutron/vars/main.yml
@@ -58,9 +58,10 @@ __neutron_helm_values:
         driver: noop
       service_providers:
         service_provider: VPN:strongswan:neutron_vpnaas.services.vpn.service_drivers.ipsec.IPsecVPNDriver:default
+      neutron_default_dns_servers: 10.96.0.20
     dhcp_agent:
       DEFAULT:
-        dnsmasq_dns_servers: "{{ neutron_coredns_cluster_ip | default('1.1.1.1') }}"
+        dnsmasq_dns_servers: "{{ neutron_coredns_cluster_ip | default('10.96.0.20') }}"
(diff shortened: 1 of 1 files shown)


roles/neutron/vars/main.yml
```diff
@@ -60,7 +60,7 @@ __neutron_helm_values:
         service_provider: VPN:strongswan:neutron_vpnaas.services.vpn.service_drivers.ipsec.IPsecVPNDriver:default
     dhcp_agent:
       DEFAULT:
-        dnsmasq_dns_servers: "{{ neutron_coredns_cluster_ip | default('10.96.0.20') }}"
+        dnsmasq_dns_servers: "{{ neutron_coredns_cluster_ip | default('1.1.1.1') }}"
```
mnaser, line 63: When using ML2/OVS, the `10.96.0.20` works just fine, but when using ML2/OVN we need to use `1.1.1.1`.

roles/neutron/vars/main.yml
mnaser, line 1: We can simply add this:

```
dnsmasq_dns_servers: "{{ neutron_coredns_cluster_ip | default('1.1.1.1') }}"
```

Into the OVN section, since that gets merged into the OVN part.

Also, perhaps I think we should add a variable called `neutron_default_dns_servers` which gets set to `10.96.0.20` if its not using ovn, and by default, do not set it, so the user has to set it if its using ovn.

mnaser (changes requested):

roles/neutron/vars/main.yml
adrianreza: I added dnsmasq_dns_servers to OVN part. And for neutron_default_dns_servers I put it on neutron conf for ML2/OVS part. I'm not sure if it was correct or not. Need your feedback @mnaser

roles/neutron/vars/main.yml
adrianreza: I reverted back for ML2/OVS to 10.96.0.20

roles/neutron/vars/main.yml
```diff
@@ -58,6 +58,7 @@ __neutron_helm_values:
         driver: noop
       service_providers:
         service_provider: VPN:strongswan:neutron_vpnaas.services.vpn.service_drivers.ipsec.IPsecVPNDriver:default
+      neutron_default_dns_servers: 10.96.0.20
```
mnaser, line 61: There is no such option, so you have to remove this.

roles/neutron/vars/main.yml
```diff
@@ -93,6 +94,10 @@ __neutron_ovn_helm_values:
         ovn_emit_need_to_frag: true
       service_providers:
         service_provider: VPN:strongswan:neutron_vpnaas.services.vpn.service_drivers.ovn_ipsec.IPsecOvnVPNDriver:default
+    dhcp_agent:
+      DEFAULT:
+        dnsmasq_dns_servers: "{{ neutron_coredns_cluster_ip | default('1.1.1.1') }}"
+        enable_isolated_metadata: true
```
mnaser, line 100: Let's remove this option, since it gets merged into the first one anyways.

mnaser (changes requested):
can you make those revisions?

roles/neutron/vars/main.yml
mnaser: Please remove this option, it doesn't exist..

mnaser (changes requested):

roles/neutron/vars/main.yml
```diff
@@ -61,7 +61,6 @@ __neutron_helm_values:
     dhcp_agent:
       DEFAULT:
         dnsmasq_dns_servers: "{{ neutron_coredns_cluster_ip | default('10.96.0.20') }}"
-        enable_isolated_metadata: true
```
mnaser, line 64: Don't remove this, please keep this,.

mnaser (changes requested):

mnaser:
please move the change to gerrit @adrianreza

adrianreza:
> please move the change to gerrit @adrianreza 

It is done @mnaser

PR #2185, opened 2024-12-11
Stop using CDN for CI

mnaser:
- **Switch image builds to use Harbor**
- **Switch Zuul to avoid CDN**
- **Small more fixes**
- **Configure CI bits to not use CDN**


mnaser:
recheck

PR #2186, opened 2024-12-11
Enhance `MySQLDown` alert

JustHumanz:
Due to MySQLDown being very noisy even the mysql cluster still running above minimal replication, so we need to enhance it 

this pr will replace the `MySQLDown` with `MysqlClusterDown`, `MysqlClusterDown` have two severity P3(warning) and P5(info). i.e when we have 5 percona-xtradb and only 1 replication down it only trigger `MysqlClusterDown` with P5 because the mysql cluster still can operational normally but if we lost 3 replication(which means there only 2 replication) it will trigger `MysqlClusterDown` with P3 since mysql cluster unable to operate

### DIFF
--- a/roles/kube_prometheus_stack/files/jsonnet/mixins.libsonnet
+++ b/roles/kube_prometheus_stack/files/jsonnet/mixins.libsonnet
@@ -147,25 +147,32 @@ local mixins = {
             },
             {
               alert: 'MysqlClusterDown',
-              'for': '1m',
-              expr: 'count(mysql_up==0) != count(mysql_up)',
-              labels: {
-                severity: 'info',
-              },
+              'for': '5m',
+              expr: 'mysql_up == 0',
+              labels: { severity: 'info' },
               annotations: {
-                summary: '{{ $value }} percona-xtradb replication down',
+                summary: 'Percona XtraDB Cluster replica is down',
+                description: "{{ $labels.instance }} replica is down.",
+              },
+            },
+            {
+              alert: 'MysqlClusterDown',
+              'for': '5m',
+              expr: 'round(count(mysql_up==1) / count(mysql_up) * 100) <= 50',
+              labels: { severity: 'warning' },
+              annotations: {
+                summary: 'Percona XtraDB Cluster replicas are down',
+                description: "{{ $value }}% of replicas are online.",
               },
             },
             {
               alert: 'MysqlClusterDown',
               'for': '1m',
-              expr: 'round(count(mysql_up==1)/count(mysql_up) * 100) <= 50',
-              labels: {
-                severity: 'warning',
-              },
+              expr: 'count(mysql_up==0) == count(mysql_up)',
+              labels: { severity: 'critical' },
               annotations: {
-                summary: 'Only {{ $value }}% percona-xtradb cluster are online',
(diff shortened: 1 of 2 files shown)


roles/kube_prometheus_stack/files/jsonnet/tests.yml
```diff
@@ -105,3 +105,18 @@ tests:
...
+  - interval: 1m
+    input_series:
+      - series: 'mysql_up'
+        values: '1'
+    input_series:
+      - series: 'mysql_up'
+        values: '1'
+    input_series:
+      - series: 'mysql_up'
+        values: '0'
```
mnaser, line 118: You're using the same key here 3 times, you need to have multiple series lines with perhaps different labels for the different instances.

You should have a test for no alerts if they're all up, and then alerts for the different ranges.

roles/kube_prometheus_stack/files/jsonnet/mixins.libsonnet
```diff
@@ -141,6 +145,29 @@ local mixins = {
...
+                severity: 'info',
+              },
+              annotations: {
+                summary: '{{ $value }} percona-xtradb replication down',
+              },
+            },
+            {
+              alert: 'MysqlClusterDown',
+              'for': '1m',
+              expr: 'count(mysql_up==1) < 3',
```
mnaser, line 162: This is hard coded, we should operate on percentage based instead.

mnaser (changes requested):

roles/kube_prometheus_stack/files/jsonnet/tests.yml
JustHumanz: already update the mysql_down unit test

roles/kube_prometheus_stack/files/jsonnet/tests.yml
```diff
@@ -105,3 +105,18 @@ tests:
...
+  - interval: 1m
+    input_series:
+      - series: 'mysql_up{instance="percona-xtradb-pxc-0", job="pxc"}'
+        values: '1'
+    input_series:
+      - series: 'mysql_up{instance="percona-xtradb-pxc-1", job="pxc"}'
+        values: '1'
+    input_series:
+      - series: 'mysql_up{instance="percona-xtradb-pxc-3", job="pxc"}'
+        values: '0'
```
mnaser, line 118:
```suggestion
    input_series:
      - series: 'mysql_up{instance="percona-xtradb-pxc-0", job="pxc"}'
        values: '1'
      - series: 'mysql_up{instance="percona-xtradb-pxc-1", job="pxc"}'
        values: '1'
      - series: 'mysql_up{instance="percona-xtradb-pxc-3", job="pxc"}'
        values: '0'                
```

You need to make this way otherwise it only processes the last value :)

Also we need to have tests for:

- all instances are up = no alerts
- one replica down = low alert
- multiple replica down = higher alert

etc :)

mnaser (changes requested):

roles/kube_prometheus_stack/files/jsonnet/tests.yml
JustHumanz: sorry for being so silly, already fix it

mnaser (approved):

mnaser:
regate

mnaser:
regate

PR #2217, opened 2024-12-13
Fix Ingress when user costume the rook cluster name

JustHumanz:
variable `rook-ceph-rgw-ceph` cannot be static like this since user can naming the rook cluster name through `rook_ceph_cluster_name` variable


mnaser:
https://github.com/vexxhost/atmosphere/pull/2208/files

PR #2231, opened 2024-12-17
chore: remove registry secret from zuul jobs 

yaguangtang:


zuul.d/container-images/base.yaml
```diff
@@ -47,21 +47,11 @@
...
     name: atmosphere-promote-container-image
     parent: ci-promote-container-image
     irrelevant-files:
       - ^doc/
-    secrets:
-      name: container_registry_credentials
-      secret: atmosphere-harbor-credentials
-      pass-to-parent: true
-    nodeset:
-      nodes: []
```
mnaser, line 66: this needs to stay.

mnaser (changes requested):

zuul.d/container-images/base.yaml
yaguangtang: updated

mnaser (approved):

PR #2249, opened 2025-01-01
[CLOUDOPS-706] rebuild libvirt image

larainema:


mnaser:
sorry @larainema -- i think the issue here is that the image might be missing from the nova image, not the libvirt?

https://github.com/vexxhost/atmosphere/pull/2248 could be the fix but just needs tob eb ackported, sorry.

PR #2262, opened 2025-01-03
update non/bfv instances

donnyao:


doc/source/admin/maintenance.rst
```diff
@@ -39,10 +39,34 @@ the following command:
...
-In order to evacuate a compute node, you will need to start by disabling the
-OpenStack compute service on the node.  This will prevent new workloads from
-being scheduled on the node.  To disable the OpenStack compute service, run
-the following command against the node you want to evacuate:
+Before initiating VM evacuation, it is important to understand the implications for
+different types of instances:
+
+- Boot from Volume Instances:
+
+These instances have their root disks stored on an external volume (e.g., Ceph).
```
guilhermesteinmuller, line 47: instead of saying "an external volume", please use distributed storage system

doc/source/admin/maintenance.rst
```diff
@@ -39,10 +39,34 @@ the following command:
...
+Evacuating these instances is safe and ensures that data integrity is maintained
+since the root volume is detached from the failing hypervisor and reattached to a
+new one.
+
+- Non-Boot from Volume Instances:
+
+These instances have their root disks stored locally on the hypervisor's ephemeral
+storage. Evacuating these instances will result in the loss of all data stored
+in the root disk, as ephemeral storage is not persistent and cannot be transferred
```
guilhermesteinmuller, line 56: I think the part that says "ephemeral is not persistent and cannot be transferred to another hypervisor" is not accurate. 

ephemeral storage is persistent, however it doesnt have the same level of redundancy as a volume sitting on a distributed storage system. In the case of a hypervisor failure, if the failure is not reversible, the vm will need to be rebuilt onto another hypervisor which will cause data loss on the root partition.

doc/source/admin/maintenance.rst
```diff
@@ -39,10 +39,34 @@ the following command:
...
+    is acceptable or has been mitigated through backups or other mechanisms.
+
+The evacuation process can be divided into two scenarios based on the state of
+the hypervisor. In order to evacuate a compute node with the hypervisor running,
+you will need to start by disabling the OpenStack compute service on the node.
+This will prevent new workloads from being scheduled on the node.  To disable
+the OpenStack compute service, run the following command against the node you
+want to evacuate:
 
 .. code-block:: console
```
guilhermesteinmuller, line 71: overral, i think we should format this as a warning or note here, so the text doesnt sound confusing

guilhermesteinmuller (changes requested):

doc/source/admin/maintenance.rst
donnyao: changed

doc/source/admin/maintenance.rst
donnyao: i think its enough in loss of all data stored in the root disk ya. i will remove it

doc/source/admin/maintenance.rst
donnyao: i just formatted the note into warning

guilhermesteinmuller (changes requested):
I think its getting better. 

however, after the warning, you are mentioning two scenarios, but just explaining the one which the hypervisor is running, which is bascially live migrating all the resources. 

The other is the incident we got today, which is hte hypervisor down, which consists in the same evacuate command buti with no "live" on it. This is the destructive one and we should highlight as well. 

I would also make a heading for "Boot from Volume Instances:" and "Non-Boot from Volume Instances" for things to be more cohesive to visualize.

PR #2280, opened 2025-01-08
fix: Use more standard configs for DBs

ricolin:
Improve DB configs for Barbican and Manila.

Ref from https://github.com/vexxhost/atmosphere/pull/2263
Also push to  https://review.opendev.org/c/openstack/openstack-helm/+/938638


mnaser:
regate

mnaser:
regate

mnaser:
recheck

PR #2313, opened 2025-01-14
add orphan-delete role

okozachenko1203:
To handle scenarios where a Helm upgrade introduces new labels, we rely on maual process to decide whether to perform orphan deletion. We can enable this by setting the `atmosphere_orphan_delete` boolean variable at the release level.

Each role also has its own specific `xxxx_orphan_delete` boolean variable, which defaults to the value of atmosphere_orphan_delete. This allows users to apply orphan deletion differently for each role as needed.

This approach provides flexibility while ensuring consistency across roles, enabling users to customize behavior for their specific use case.

### DIFF
--- a/roles/orphan_delete/tasks/main.yml
+++ b/roles/orphan_delete/tasks/main.yml
@@ -12,12 +12,23 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-memcached_helm_release_name: memcached
-memcached_helm_chart_path: "../../charts/memcached/"
-memcached_helm_chart_ref: /usr/local/src/memcached
 
-memcached_helm_release_namespace: openstack
-memcached_helm_kubeconfig: "{{ kubeconfig_path | default('/etc/kubernetes/admin.conf') }}"
-memcached_helm_values: {}
+- name: Delete deployments
+  kubernetes.core.k8s:
+    name: testing
+    api_version: apps/v1
+    kind: Deployment
+    state: absent
+    delete_options:
+      propagationPolicy: Orphan
+  loop: {{ orphan_delete_deployments}}
 
-memcached_orphan_delete: "{{ atmosphere_orphan_delete }}"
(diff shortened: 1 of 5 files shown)


okozachenko1203:
this is just a poc PR with memcached. Once the approach is approved, we can replicate it to all roles.

roles/memcached/defaults/main.yml
```diff
@@ -19,3 +19,5 @@ memcached_helm_chart_ref: /usr/local/src/memcached
 memcached_helm_release_namespace: openstack
 memcached_helm_kubeconfig: "{{ kubeconfig_path | default('/etc/kubernetes/admin.conf') }}"
 memcached_helm_values: {}
+
+memcached_orphan_delete: "{{ atmosphere_orphan_delete }}"
```
mnaser, line 23: This is not a good approach IMHO, because with CI/CD in place, it's going to be impossibel for users to know when to use what, it should auto-detect and automatically enable as needed.

roles/orphan_delete/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
...
+
+- name: Delete statefulsets
+  kubernetes.core.k8s:
+    name: testing
+    api_version: apps/v1
+    kind: StatefulSet
+    state: absent
+    delete_options:
+      propagationPolicy: Orphan
+  loop: {{ orphan_delete_statefulsets}}
```
mnaser, line 34: This should all be auto-detected.

mnaser (changes requested):

roles/orphan_delete/tasks/main.yml
okozachenko1203: @mnaser i made some progress for auto detection
let me make a patch in the gerrit

okozachenko1203:
https://review.vexxhost.dev/c/atmosphere/+/313

Issue #2492, opened 2025-04-17
barbican-db-sync fails on upgrade from Bobcat to Caracal


legit-ninja:
Barbican-db-sync container crashloopbackoff while attempting an upgrade from Atmosphere v3.3.1 (Bobcat 2023.2) to v4.4.1 (Caracal 2024.1)

[Original bug](https://docs.sqlalchemy.org/en/14/changelog/changelog_14.html#change-fc71152a7598dd55bf730bbadfc508f0)

[Changelog for 2.0.0b1](https://docs.sqlalchemy.org/en/20/changelog/changelog_20.html#change-2.0.0b1)

[Relevant Topic](https://lists.openstack.org/archives/list/openstack-discuss@lists.openstack.org/thread/A7PWH2IOORVUVRD2IGD55V2K6X3GGBWW/)

[OpenStack Requirements](https://review.opendev.org/c/openstack/requirements/+/879743)

```
$kubectl -n openstack logs -f barbican-db-sync-k7g54
Defaulted container "barbican-db-sync" out of: barbican-db-sync, init (init)
+ barbican-db-manage upgrade
2025-04-16 15:27:43.434 7 WARNING oslo_db.sqlalchemy.engines [-] MySQL SQL mode is 'ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION', consider enabling TRADITIONAL or STRICT_ALL_TABLES
2025-04-16 15:27:43.439 7 INFO alembic.runtime.migration [-] Context impl MySQLImpl.
2025-04-16 15:27:43.439 7 INFO alembic.runtime.migration [-] Will assume non-transactional DDL.
+ set +x
Ensuring that project KEKs are wrapped with the target global KEK
MySQL SQL mode is 'ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION', consider enabling TRADITIONAL or STRICT_ALL_TABLES
Traceback (most recent call last):
  File "/tmp/simple_crypto_kek_rewrap.py", line 157, in <module>
    main()
Retrieving all available projects
Retrieving KEKs for Project 887f141cb3704636bf795456825a0219
[...]

PR #2524, opened 2025-04-22
Check transaction before begin

ricolin:
Allow using nested transaction if db session already starts root transaction. Check if DB transaction already starts in barbican kek rewrap. And use nested transaction if DB session already starts it's root transaction.

related: https://github.com/vexxhost/atmosphere/issues/2492

Change-Id: I829698ad55011c1a429920d85eb29f7de9e30e0b

### DIFF
--- a/.charts.yml
+++ b/.charts.yml
@@ -18,6 +18,7 @@ charts:
         review.opendev.org:
           - 916034
           - 944975
+          - 947782
   - name: ceph-csi-rbd
     version: 3.11.0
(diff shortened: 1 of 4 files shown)


mnaser:
@legit-ninja do you think you can test this out at your side and see if it also solves your issue?

legit-ninja:
This directly addresses the issue I was having and seems to be a better solution.  You can close my PR.  Thanks!

Issue #2377, opened 2025-04-07
[463] LVM volumes of guests are exposed to the host


okozachenko1203:
When using a storage backend that uses iSCSI or Fiber channel, the device is exposed directly to the physical host and then passed through to the virtual machine.

It seems that there are edge cases where LVM tries to load the logical volumes that live inside of the volumes of the customer.  We should make sure that LVM doesn’t load anything that is exposed over the SAN.

The workaround for this right now is to run the following on the compute host:
```
vgchange -an
```
This will disable all the volume groups on the system, but you must be careful if LVM is in use already on the host.

PR #2533, opened 2025-04-25
Setup a lvm role to update lvm configure

larainema:
https://github.com/vexxhost/atmosphere/issues/2377


roles/lvm/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Discover lvm devices
+  ansible.builtin.shell: |
+    set -o pipefail
+    /sbin/pvdisplay | awk '/PV Name/ {print $3}' | sed 's/\/dev\///g'
+  register: lvm_devices
+  changed_when: lvm_devices.rc != 0
+  failed_when: false
```
mnaser, line 21: The issue here is that if we have an environment where this issue has already manifested, Atmosphere won't be able to fix it and instead add those PVs into it, which we don't want.

roles/lvm/templates/lvm.conf.j2
```diff
@@ -0,0 +1,31 @@
...
+            {% endif %}
+        {% endfor %}
+    {% endif %}
+
+# Ansible Discovered LVM Devices {{ lv_devices }}
+{% endif %}
+
+{% if used_lvm_devices|length <= 0 %}
+    {# If there are no LVM devices present, allow all devices to be scanned #}
+    {% if used_lvm_devices.append('"a/.*/"') %}{% endif %}
```
mnaser, line 21: We actually want to not allow anything, because if there is no LVMs, we will allow plugged volumes to be scanned, this is how I fixed it in an environment where the base OS is not using LVM:

```
        filter = [ "r|.*|" ]
```

mnaser (changes requested):

PR #2537, opened 2025-04-28
fix libvirt tls sidecar missing libvirt run directory

yaguangtang:


charts/libvirt/templates/daemonset-libvirt.yaml
```diff
@@ -166,6 +166,8 @@ spec:
               mountPath: /etc/pki/libvirt
             - name: etc-pki-libvirt-vnc
               mountPath: /etc/pki/libvirt-vnc
+            - name: run
+              mountPath: /run
```
mnaser, line 170: Would it make more sense to mount the libvirt run folder instead?

mnaser (changes requested):

charts/libvirt/templates/daemonset-libvirt.yaml
yaguangtang: yeah, that whould be more secure.

yaguangtang:
recheck

adrianreza (approved):

guilhermesteinmuller (approved):

ricolin (approved):

ricolin:
Also would be nice to have this in https://github.com/openstack/openstack-helm too so we don't have to reform stuff when upgrade OpenStack-helm

mnaser:
> Also would be nice to have this in https://github.com/openstack/openstack-helm too so we don't have to reform stuff when upgrade OpenStack-helm

I think in our case here, `libvirt-tls-sidecar` is not something upstream uses.

charts/libvirt/templates/daemonset-libvirt.yaml
```diff
@@ -166,6 +166,8 @@ spec:
               mountPath: /etc/pki/libvirt
             - name: etc-pki-libvirt-vnc
               mountPath: /etc/pki/libvirt-vnc
+            - name: libvirt-run
```
mnaser, line 169: For consistency can we make this run-libvirt?

releasenotes/notes/fix-vnc-tls-renew-16e30a87662190ba.yaml
mnaser, line 1: Can you rewrite this without acronyms such as cert?  AI can be helpful here at writing something using Microsoft technical writing standards.

mnaser (changes requested):

mark-ac-garcia:
Just wondering if there's anything I can do to help get this PR ready for merging quickly?

charts/libvirt/templates/daemonset-libvirt.yaml
mnaser: I think this should still be `run-libvirt` for consistency.

.github/styles/config/vocabularies/Base/accept.txt
mnaser, line 1: You'll need to fix the merge conflict here and make sure this is sorted alphabetically since I added a `pre-commit` check for this.

charts/libvirt/templates/daemonset-libvirt.yaml
```diff
@@ -354,6 +356,9 @@ spec:
         - name: run
           hostPath:
             path: /run
+        - name: libvirt-run
```
mnaser, line 359: Same here

mnaser (changes requested):

yaguangtang:
@mnaser So the order check is based on the container image url with sha256 added, to pass the CI, I just added the sha256 for each image, is that OK?

charts/libvirt/templates/daemonset-libvirt.yaml
yaguangtang: done

mnaser:
> @mnaser So the order check is based on the container image url with sha256 added, to pass the CI, I just added the sha256 for each image, is that OK?

I think you're seeing an unrelated thing for the shasums, this is just because we build the image inside `ansible-lint`.

You need to fix the sorts only.

mnaser:
@yaguangtang  see https://github.com/vexxhost/atmosphere/commit/f1227d1644778fedc42fa39b2c682c1036bfd008 for the fix i put in

mnaser (approved):

PR #2547, opened 2025-04-29
chore: fix vale lint for docs

yaguangtang:


mnaser:
dropped vale for now.. causing a lot of headaches.

PR #2551, opened 2025-04-30
[stable/2023.1]Fix manila multiattach

ricolin:
Bump nova microversion in manila to support multiattach feature. Also use internalURL instead of publicURL to keep manila-nova communication in internal.

Change-Id: I11fdc56abbabdac6b660fe6521ad74eb2027f089


yaguangtang:
recheck

Issue #2523, opened 2025-04-22
[ovn] neutron_pg_drop deleted when using network log


mnaser:
reference: https://bugs.launchpad.net/neutron/+bug/2107925

---------------------

When you create a network log resource to enable OVN logging:

https://docs.openstack.org/neutron/latest/contributor/internals/ovn/ovn_network_logging.html

It seems that will add a an ACL rule similar to this:

```
_uuid : bd6eaad4-939d-4624-b2f9-9701d1ec402e
action : drop
direction : to-lport
external_ids : {}
label : 1753594327
log : true
match : "outport == @neutron_pg_drop && ip"
meter : acl_log_meter
name : neutron-4054c65c-9a05-4bf4-8abe-f31959dbd56f
options : {log-related="true"}
priority : 1001
severity : info
tier : 0
```

However, once that ACL rule comes in, Neutron starts to think that the rule is not supposed to be there, so when a sync runs, it'll "create" new ACLs and "delete" these ones (not from this one above but similar):

```
2025-04-11 00:52:12.959 37 WARNING neutron.plugins.ml2.drivers.ovn.mech_driver.ovsdb.ovn_db_sync [None req-85cacb15-72f4-4c66-b5f4-587c2d558780 - - - - - -] ACLs-to-be-added 7 ACLs-to-be-removed 7
2025-04-11 00:52:12.959 37 WARNING neutron.plugins.ml2.drivers.ovn.mech_driver.ovsdb.ovn_db_sync [None req-85cacb15-72f4-4c66-b5f4-587c2d558780 - - - - - -] ACL found in Neutron but not in OVN DB for port group pg_c40034c8_8393_4a72_9ca9_6d4be2da5db1
2025-04-11 00:52:12.960 37 WARNING neutron.plugins.ml2.drivers.ovn.mech_driver.ovsdb.ovn_db_sync [None req-85cacb15-72f4-4c66-b5f4-587c2d558780 - - - - - -] ACL found in Neutron but not in OVN DB for port group pg_c40034c8_8393_4a72_9ca9_6d4be2da5db1
2025-04-11 00:52:12.960 37 WARNING neutron.plugins.ml2.drivers.ovn.mech_driver.ovsdb.ovn_db_sync [None req-85cacb15-72f4-4c66-b5f4-587c2d558780 - - - - - -] ACL found in Neutron but not in OVN DB for port group pg_c40034c8_8393_4a72_9ca9_6d4be2da5db1
[...]

PR #2568, opened 2025-05-06
Add NeutronOVNACLNotSync alert

ricolin:
relate-to: https://github.com/vexxhost/atmosphere/issues/2523

This dependency is for make sure the sync works properly with network log resources exists.
So it will no trigger false alert.
 needs https://github.com/vexxhost/atmosphere/pull/2670

Change-Id: I86bcccb6607eca98d86aa5008e0c538f30357a44


mnaser (changes requested):
I think we should make this into a CronJob as part of the Neutron Helm chart rather than a standalone manifest, it will make the lifecycle management much easier.

mnaser (changes requested):
It seems this change drops a specific feature for using the interface nickname.

ricolin:
switch to same repo pr https://github.com/vexxhost/atmosphere/pull/2726

PR #2645, opened 2025-05-14
feat: tune PXC Liviness probes

fitbeard:
This PR introduces 2 changes:

Remove sidecar livenessprobe - sidecar should wait for main containers indefinitely and not terminate pxc pod. For example when long SST is running.

Increase liveness timeout and failure rates for pxc from defaults:
`pxc.livenessProbes.timeoutSeconds: 5`
`pxc.livenessProbes.failureThreshold: 3`

In cases with large databases and SST refill defaults ar extremely small.


roles/percona_xtradb_cluster/vars/main.yml
```diff
@@ -27,12 +27,11 @@ _percona_xtradb_cluster_spec:
         ports:
           - name: metrics
             containerPort: 9104
-        livenessProbe:
-          httpGet:
-            port: metrics
-            path: /
```
mnaser, line 33: Does it make sense to switch this to a `readinessProbe` perhaps?

mnaser (changes requested):
Small question, but good change otherwise

roles/percona_xtradb_cluster/vars/main.yml
fitbeard: let me test it once again

fitbeard: Yes. Changed to `readinessProbe`.

PR #2657, opened 2025-05-15
add ovs exporter in ovn chart

okozachenko1203:


releasenotes/notes/ovn-ff7648cdb00ea90d.yaml
```diff
@@ -0,0 +1,3 @@
+---
+features:
+  - add support for ovs exporter
```
ricolin: need to either add this to accept val list or change to 
`OVS (Open vSwitch)`

Issue #2658, opened 2025-05-15
Add documentation for SEV


mnaser:
SEV works out of the box in Atmosphere, however, we need to document how users can enable it and talk about all the different limitations:

https://docs.openstack.org/nova/latest/admin/sev.html

PR #2659, opened 2025-05-16
Add SEV doc

larainema:
https://github.com/vexxhost/atmosphere/issues/2658


mnaser (changes requested):
This seems to be mostly a copy/paste of the upstream docs, could it be rewritten to be adapted to Atmosphere, for example, can you check if we use a new enough libvirt to justify avoiding `num_memory_encrypted_guests` and also the examples of creating instance still have vtpm in there.

PR #2665, opened 2025-05-16
feat: introduce metrics-server

fitbeard:
This is initial introduction of components from `atmosphere.common` collection.

One thing about FluxCD i don't like is change tracking. We are spoiled by `kubernetes.core.helm` and how chart overrides are handled.

With FluxCD after we set:
```
metrics_server_helm_values:
  nodeSelector:
```
the only way to get rid of this value is:
```
metrics_server_helm_values:
  nodeSelector: null
```


fitbeard:
@mnaser this is not a bug but feature https://github.com/fluxcd/kustomize-controller/issues/372
I'll try to workaround this in atmosphere.common with something like `current_helm_release.resources[0].spec.values` -> new values -> diff -> null

mnaser:
@fitbeard I think I've seen similar behaviour with the `kubernetes.core.k8s` module, so I wonder if it's compounding.  I think if you try and create a `ConfigMap` for example, it'll end up being created with 2 keys for example, but if you remove one of them and run the module again, it'll actually not remove it.

fitbeard:
I totally messed here. Recreated new one with only relevant changes: https://github.com/vexxhost/atmosphere/pull/2784

Issue #2317, opened 2025-02-12
keycloak: increase replicaCount


satrik:
Hey there, 

setting the keycloak_helm_values replicaCount to 2 (or more) leads to admin UI stuck in `Loading the Admin UI` and the network tab of the developer console shows s few 404s for some files. 

This is the config we made. 
```
keycloak_helm_values:
  replicaCount: 3
```

We want just to have a HA setup for keycloak so what else is needed to make this work?

fitbeard:
Bitnami chart is not designed to run Keycloak (especially newer versions) in HA mode. Most of the time it's just a version bump without checking/testing deeper what's changed.
This is override snippet from our setup with Keycloak 25.x (keep in mind that on a newer versions Keycloak admin and public interfaces are located on different ports 9090 and 8080). Here Infinispan is properly enabled and sessions survines pod restarts.
For even better HA external Infinispan with different cache stack then 'kubernetes' is recommended.
```
- name: keycloak
  version: 22.2.3
  repository: oci://registry-1.docker.io/bitnamicharts
```

```
keycloak:
  image:
    registry: registry
    repository: images/keycloak
    tag: 25.x.x
  replicaCount: 3
  # Overriding Bitnami default which is 'Parallel'
  # https://github.com/keycloak/keycloak/issues/21108
  podManagementPolicy: OrderedReady
  resources:
    requests:
      cpu: 500m
      memory: 512Mi
  # These workarounds below are needed to allow the Bitnami Helm chart to work with
  # the upstream image.
  enableDefaultInitContainers: false
  containerSecurityContext:
    readOnlyRootFilesystem: false
    runAsUser: 1000
  extraEnvVars:
    - name: JAVA_OPTS_APPEND
      value: "-Djgroups.dns.query=keycloak-headless.keycloak.svc"
    - name: KC_CACHE_STACK
      value: kubernetes
    - name: KC_CACHE
      value: ispn
  command:
    - /opt/keycloak/bin/kc.sh
    - --verbose
    - start
    - --health-enabled=true
    - --http-enabled=true
    - --http-port=8080
    - --hostname-strict=true
    - --proxy-headers=xforwarded
    - --transaction-xa-enabled=false
    - --metrics-enabled=true
  auth:
    adminUser: keycloak-admin
    existingSecret: keycloak-admin-secret
    passwordSecretKey: keycloak_admin_password
  adminRealm: master
  adminIngress:
    enabled: true|false
    ingressClassName: atmosphere
[...]

PR #2669, opened 2025-05-19
feat: HA Keycloak

fitbeard:
Fixes: https://github.com/vexxhost/atmosphere/issues/2317

But i think more important are findings from me trying to upgrade Keycloak's major version: https://github.com/vexxhost/atmosphere/issues/2671


fitbeard:
We can enable scaling for current Keycloak's version we use currently (24.x).

PR #2675, opened 2025-05-22
feat: upgrade pxc components

fitbeard:
Stick with `1.16.1` operator because of kubernetes 1.28 support.

Engine version pin to latest 8.0.x

Add CRD update tasks

### DIFF
--- a/doc/source/admin/database-backups.rst
+++ b/doc/source/admin/database-backups.rst
@@ -118,8 +118,7 @@ Examples:
   metadata:
     finalizers:
       # Finalizer can be set even if you use a persistent volume.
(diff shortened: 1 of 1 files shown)


mnaser:
@fitbeard it seems that it's failing to cleanly backport, do you mind making PRs for the backports here please?

fitbeard:
@mnaser this also? https://github.com/vexxhost/atmosphere/pull/2648

fitbeard:
@mnaser PRs are ready. FYI i skipped `stable/zed` because it's still on PXC CR 1.13.0 and every new version is back compatible with only 3 versions including current (we are bumping to 1.16.1).

PR #2730, opened 2025-06-04
feat: add dynamic-routing and networking-generic-switch support for Neutron

fitbeard:
For Baremetal switching we need `networking-generic-switch`
For bgp-dragent we need `neutron-dynamic-routing`

### DIFF
--- a/.github/styles/config/vocabularies/Base/accept.txt
+++ b/.github/styles/config/vocabularies/Base/accept.txt
@@ -1,3 +1,4 @@
+BGP
 CRD
 Ceph
 Glance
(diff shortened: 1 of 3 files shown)


fitbeard:
```
 #0 2.873   × No solution found when resolving dependencies:
#0 2.873   ╰─▶ Because only oslo-service[threading]<=4.2.0 is available
#0 2.873       and networking-generic-switch==7.6.0.dev11 depends on
#0 2.873       oslo-service[threading]>=4.2.0, we can conclude that
#0 2.873       networking-generic-switch==7.6.0.dev11 depends on oslo-service==4.2.0.
#0 2.873       And because networking-generic-switch==7.6.0.dev11
#0 2.873       depends on oslo-service==4.1.1, we can conclude that
#0 2.873       networking-generic-switch==7.6.0.dev11 cannot be used.
#0 2.873       And because only networking-generic-switch==7.6.0.dev11 is available
#0 2.873       and you require networking-generic-switch, we can conclude that your
#0 2.873       requirements are unsatisfiable.
```

mnaser (changes requested):
I think this deserves a release note to say that the images now include those compoennts.

PR #2735, opened 2025-06-06
[stable/2024.1]add missing cinder.auth_type

ricolin:
[stable/2024.1]add missing cinder.auth_type

Change-Id: Idfa71708a45c1a3db9115601ba580432883cbbec

### DIFF
--- a/releasenotes/notes/Fix-cinder-auth-type-9ea017879a578f2a.yaml
+++ b/releasenotes/notes/Fix-cinder-auth-type-9ea017879a578f2a.yaml
@@ -1,0 +1,7 @@
+---
+fixes:
+  - |
+    The ``[cinder]/auth_type`` configuration value wasn't set resulting in
+    the entire Cinder section not render in the configuration file, it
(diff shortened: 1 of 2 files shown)


ricolin:
> @ricolin how come we're going straight to 2024.1 and before?

The fix for 2024.2 and main has already been introduced when you fix  https://github.com/vexxhost/atmosphere/commit/0a5b2e41b077615eab4870694d2ded84f9d17056#diff-0bddc48df6875866cd94690ab0c8227a1d488861da02514d3556b3a9e6b5785dR65-R115
But on the way to backport to 2024.1 and further, this part got dropped

mnaser:
@ricolin in the backports though we removed the other variables, shouldn't we be doing the same thing here too?

mnaser (changes requested):
@ricolin in the backports though we removed the other variables, shouldn't we be doing the same thing here too?

ricolin:
> @ricolin in the backports though we removed the other variables, shouldn't we be doing the same thing here too?

removed

PR #2739, opened 2025-06-10
fix: grant enough permission to c-vol

okozachenko1203:
volume creation from qcow2 image fails with this error

```sh
raise FailedToDropPrivileges(msg)\n', 'oslo_privsep.daemon.FailedToDropPrivileges: Privsep daemon failed to start\n']
```

### DIFF
--- a/roles/cinder/vars/main.yml
+++ b/roles/cinder/vars/main.yml
@@ -29,6 +29,12 @@ __cinder_helm_values:
           volumes:
             - name: cinder-tmp
               emptyDir: {}
+    security_context:
+      cinder_volume:
+        pod:
+          cinder_volume:
+            privileged: true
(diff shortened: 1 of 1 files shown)


releasenotes/notes/cinder-vol-grant-priviledge-for-privsep-failure-8d4317405cc61f60.yaml
mnaser, line 1: Can you remove the trailing new line? pre-commit doesn't like it.  Also, can you add a bit more context to the change here in the release note, assuming someone reads this without the commit.

mnaser (changes requested):

mnaser (approved):

PR #2765, opened 2025-06-13
chore: fix glance image dependency[stable/2025.1 only]

yaguangtang:


ricolin (dismissed):
the issue seems exist for now in libpcre2-10 and after so libpcre2-8 seems correct to me
Maybe we can leave a comment inline so we can remove this fix once the issue fix in libpcre

mnaser:
@yaguangtang I think the right thing to do here is that we should actually bump `main` branch to use Noble, fix all the issues in that PR, and backport the fixes into 2025.1

PR #2773, opened 2025-06-14
Bump openstack service container image to noble

yaguangtang:


.python-version
mnaser, line 1: Why did we change this?  Technically there is nothing changed on the Ansible or CI side, so why did we flip this to Python 3.12?

I think we should remove this file to be honest.

.github/actions/molecule/action.yaml
```diff
@@ -39,8 +39,8 @@ runs:
         uv sync --dev
         echo ${{ github.workspace }}/.venv/bin >> $GITHUB_PATH
       env:
-        UV_PYTHON_DOWNLOADS: never
-        UV_PYTHON_PREFERENCE: only-system
+        UV_PYTHON_DOWNLOADS: auto
+        UV_PYTHON_PREFERENCE: system
```
mnaser: This is using Python 3.12 and deploying it, any reason for this?

images/openstack-venv-builder/Dockerfile
```diff
@@ -32,6 +32,7 @@ EOF
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
 COPY --from=requirements --link /upper-constraints.txt /upper-constraints.txt
 RUN --mount=type=cache,target=/root/.cache/uv <<EOF bash -xe
+uv cache clean uwsgi
```
mnaser, line 35: What's the point of this change?  If we need to use a newer version of `uwsgi`, why don't we add version matchers to `uwsgi` (such as `>2.1.3` or something like that)?

releasenotes/notes/bump-to-noble-image-391514753483295e.yaml
```diff
@@ -0,0 +1,11 @@
...
+features:
+  - |
+    Atmosphere now use Ubuntu Noble as default image for OpenStack images.
+    OpenStack services now uses newer kernel, standard library and python.
+    Which brings better security and performance overall.
+  - |
+    The default running python moved from python3.10 to python 3.12.
+    Python 3.12 brings up overall speed increases over python 3.10
+    with better cache efficiency.
+    Note that Atmosphere still supports python 3.10.
```
mnaser: I think this is mixing up Python (for CI and for Ansible) vs Python (in the Docker image).  We should fix the features here to accomodate for this.

Please ask Claude to help you rewrite this :)

mnaser (changes requested):
@ricolin @yaguangtang good stuff, almost there.

.github/actions/molecule/action.yaml
ricolin: As we use 22.04 environment, python3.12 is not there for uv to fetch, and it need to get it from somewhere else.
otherwise it will raise error when it can't find python3.12

.python-version
ricolin: I think this file is somehow referenced when images are build or run in CI against new python3.12 images. I can surely remove this if that works for CI :)

images/openstack-venv-builder/Dockerfile
ricolin: For this I think we hit similar issue like https://github.com/astral-sh/uv/issues/12463
where uv cache seems not cleanup properly between build.

.github/actions/molecule/action.yaml
mnaser: It's complaining there is no Python 3.12 because of the fact that you changed the `.python-version` file to say Python 3.12.

PR #2783, opened 2025-06-20
ci: update CSI job to use ubuntu-latest

mnaser:
## Summary
- Update CSI job runner from ubuntu-22.04 to ubuntu-latest
- Remove .python-version file as it's no longer needed

## Test plan
- [ ] CI tests pass with the new runner configuration
- [ ] CSI tests run successfully on ubuntu-latest

🤖 Generated with [Claude Code](https://claude.ai/code)


mnaser:
So I think the CSI job is running the Ceph backend no matter what so we should fix that. 

Also, we should bump the version here to avoid the bug with AppArmor..

PR #2793, opened 2025-06-27
Mising no_log for create user and password

ricolin:
Change-Id: I9b60ca505d7fd6ab6462fcc0825bbfbee4de2d9f


mnaser (changes requested):
@ricolin I think this should include a release note.

ricolin:
> @ricolin I think this should include a release note.

updated

PR #2805, opened 2025-07-07
Fix cinder auth in nova conf

ricolin:
Change-Id: Id69d17f0b007711a58b307a187a2dab4c660154d

### DIFF
--- a/.charts.yml
+++ b/.charts.yml
@@ -161,6 +161,10 @@ charts:
     version: 0.3.46
     repository: *openstack_helm_repository
     dependencies: *openstack_helm_dependencies
+    patches:
+      gerrit:
+        review.opendev.org:
+          - 954211
   - name: octavia
     version: 0.2.9
     repository: *openstack_helm_repository
--- a/charts/nova/values.yaml
+++ b/charts/nova/values.yaml
@@ -1435,6 +1435,7 @@ conf:
       auth_type: password
       auth_version: v3
     cinder:
+      auth_type: password
       catalog_info: volumev3::internalURL
     database:
       max_retries: -1
@@ -1856,6 +1857,22 @@ endpoints:
     port:
       api:
         default: 9292
+        public: 80
+  volumev3:
+    name: cinderv3
+    hosts:
+      default: cinder-api
+      public: cinder
+    host_fqdn_override:
+      default: null
+    path:
+      default: '/v3/%(tenant_id)s'
+      healthcheck: /healthcheck
+    scheme:
+      default: http
+    port:
+      api:
+        default: 8776
         public: 80
   compute:
     name: nova
--- a/roles/nova/vars/main.yml
+++ b/roles/nova/vars/main.yml
@@ -61,8 +61,6 @@ _nova_helm_values:
         barbican_endpoint_type: internal
       cache:
         backend: oslo_cache.memcache_pool
-      cinder:
-        auth_type: password
       conductor:
         workers: 8
       compute:


releasenotes/notes/fix-cinder-auth-in-nova-adead5a3a276781a.yaml
```diff
@@ -0,0 +1,8 @@
+---
+fixes:
+  - |
+    Fix cinder authentication configurations in Nova.
+    This will honor authentication configurations under
+    OpenStack helm endpoints like
+    `openstack_helm_endpoints_nova_region_name`.
+    To able to override default configurations.
```
mnaser: This will not render properly.  First of all, our release notes use reStructuredText so the ` will not actually give you what you think it does.  Also, even if it did, you should write this properly, since this will evaluate like this:

> Fix cinder authentication configurations in Nova. This will honor authentication configurations under OpenStack helm endpoints like `openstack_helm_endpoints_nova_region_name`. To able to override default configurations.

You can see that is not easy to read, and you will need to use `` for monospace text as well.  I suggest using your AI of choice with the following prompt: "Can you rewrite this release note using proper grammer and the Microsoft Writing Style Guide and in reStructuredText format, keeping the same briefness as it will be added using reno" which for me using ChatGPT gives me this:

```
---
fixes:
  - |
    Corrected Cinder authentication configuration handling in Nova.
    Nova now respects authentication overrides defined in OpenStack Helm endpoints,
    such as ``openstack_helm_endpoints_nova_region_name``.
```

I'm not sure if that passes Vale, but you can pass the Vale errors after and keep going from there. :)

mnaser (changes requested):
Other than the release note, can you confirm this was tested at your side (I think @yaguangtang reported this?)

ricolin:
> Other than the release note, can you confirm this was tested at your side (I think @yaguangtang reported this?)

Yes, all tested on my side

yaguangtang (approved):

yaguangtang:
> Other than the release note, can you confirm this was tested at your side (I think @yaguangtang reported this?)

I tested with a custom region name with atmosphere 2.4 applied this patch, and it fixes the issue. 
 
nova@nova-api-osapi-65775d6d4-pv4fk:/etc/nova$ cat nova.conf |grep -B 5 -A 8 os_region_name
max_retries = -1
[cinder]
auth_type = password
auth_url = http://keystone-api.openstack.svc.cluster.local:5000/
catalog_info = volumev3::internalURL
os_region_name = us-central-1
password = hWJoyrUS6Wuv8cVBcnjj3Ybk9Ya7iRaW
project_domain_name = service
project_name = service

ricolin:
> > Other than the release note, can you confirm this was tested at your side (I think @yaguangtang reported this?)
> 
> I tested with a custom region name with atmosphere 2.4 applied this patch, and it fixes the issue.
> 
> nova@nova-api-osapi-65775d6d4-pv4fk:/etc/nova$ cat nova.conf |grep -B 5 -A 8 os_region_name max_retries = -1 [cinder] auth_type = password auth_url = http://keystone-api.openstack.svc.cluster.local:5000/ catalog_info = volumev3::internalURL os_region_name = us-central-1 password = hWJoyrUS6Wuv8cVBcnjj3Ybk9Ya7iRaW project_domain_name = service project_name = service

Nice!

Here also post the conf in main branch deployment  with scenario `no custom region override` from my earlier test
```
[cinder]
auth_type = password
auth_url = http://keystone-api.openstack.svc.cluster.local:5000/
catalog_info = volumev3::internalURL
os_region_name = RegionOne
password = uUlzgPwa7u9Mj6q9BT5hiPEm8vBoNShi
project_domain_name = service
project_name = service
user_domain_name = service
username = cinder-RegionOne

```

yaguangtang (approved):

mnaser (approved):

PR #2814, opened 2025-07-09
bump ovn chart version to 2025.1.0

okozachenko1203:


mnaser (changes requested):
The change looks good initially but we will need to bump `helm-toolkit` to land this.  Also, I would like to know what the state of the DaemonSets looks like before upgrade and after upgrade to this revision.

Will there be an outage happening?  Since I think the DaemonSet name will change as well to `ovn-controller-default-xxx` or something perhaps?  Can you share the output of `kubectl get pods` before and after?

mnaser:
@okozachenko1203 Can we please get this fixed up please to be able to merge today?

yaguangtang:
@mnaser  Corvex is using 5.0.2 which is stable/2024.2, means this 2025.1 helm chart need to backport to stable/2024.2 or upgrade their atmosphere to 6.0.0

mnaser:
> @mnaser we is using 5.0.2 which is stable/2024.2, means this 2025.1 helm chart need to backport to stable/2024.2 or upgrade their atmosphere to 6.0.0

I don't think we want to upgrade to 6.0.0 so we need to land the `helm-toolkit` backports and then backport this all the way to their release.

PR #2815, opened 2025-07-10
[helm-toolkit][bump chart] 2025.1.8

ricolin:
[helm-toolkit][bump chart] 2025.1.8

Change-Id: I19027d448fb0eaf423a2f02150ebafb5e41ab79a


ricolin:
As vale is fail because we have too many files in pr here, I create another draft pr for validate vale for release note https://github.com/vexxhost/atmosphere/pull/2822

charts/ironic/values.yaml
```diff
@@ -107,7 +107,6 @@ conf:
       api_url: null
     database:
       connection: null
-      max_retries: -1
```
mnaser, line 110: accidental delete?  or we need to maybe move this into our vars?

ricolin: it got deleted when purge Priorityclassname patch file.
Added to https://review.opendev.org/c/openstack/openstack-helm/+/955851 and reference it from the new commit, so we don't need to carry this for long.

roles/ceph_provisioners/vars/main.yml
```diff
@@ -30,3 +30,6 @@ _ceph_provisioners_helm_values:
     job_namespace_client_key_cleaner: false
     job_namespace_client_key: false
     storageclass: false
+  pod:
+    labels:
+      include_app_kubernetes_io: false
```
mnaser: To keep things consistent, the pod section should be above the conf section like the others.

roles/memcached/vars/main.yml
```diff
@@ -19,3 +19,6 @@ _memcached_helm_values:
   monitoring:
     prometheus:
       enabled: true
+  pod:
+    labels:
+      include_app_kubernetes_io: false
```
mnaser, line 21: Let's move this under images.

roles/tempest/vars/main.yml
```diff
@@ -13,6 +13,9 @@
 # under the License.
 
 _tempest_helm_values:
+  pod:
+    labels:
+      include_app_kubernetes_io: false
```
mnaser: Move to above conf to stay consistent

roles/staffeln/vars/main.yml
mnaser, line 1: It seems the staffeln chart did not use an updated helm-toolkit?

mnaser (changes requested):
You forgot the `openvswitch` role to drop the labels for.. also, for the unit tests, i"ll probably revise them with a seperate test because we're not longer testing "default values" by injecting values after :)

mnaser:
@ricolin i was using the new UI and itw as limited to 300 files, so sorry missed aw hole bunch of stuff first time around :)

roles/staffeln/vars/main.yml
ricolin: yes, I plan to do it in another PR as current PR already overwhelmed

charts/patches/heat/0003-htk-job_ks_user-to-create-multiple-users.patch
okozachenko1203, line 1: @ricolin do we need this patch really?
https://github.com/openstack/openstack-helm/commit/a7f921d10b9a52c2ed29519e0e97c9931150698b
this change is backward compatible so as long as we use serviceUser, then single ks user creation should work.

charts/patches/manila/0003-fix-manila-manifests.patch
okozachenko1203, line 1: @ricolin i'd make another PR for this patch
https://github.com/openstack/openstack-helm/commit/c249c90bdbeb3aebdfe8744862cdb016c6404e62

PR #2817, opened 2025-07-10
[stable/2024.2] fix magnum trusts

ricolin:
ref: https://review.opendev.org/c/openstack/magnum/+/954554


images/magnum/patches/magnum/0001-Fix-trust-create.patch
mnaser, line 1: No need for this, just bump to latest Magnum version because this is already merged.

releasenotes/notes/fix-magnum-trusts-ee4824c8e6e3610d.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    backport fixes for Magnum on Keystone RBAC failure.
+    That caused by unexpected parameter.
```
mnaser, line 5: This is again very unhelpful, this would make more sense:

Resolved regression causing Magnum clusters failing to create caused by the usage of new version of OpenStack SDK

Something along those lines, what you said reading that alone literally gives no context for the user.

PR #2823, opened 2025-07-17
chore: support redhat family os

yaguangtang:


roles/cluster_issuer/handlers/main.yml
```diff
@@ -12,6 +12,13 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Install openssl-perl package
+  when: ansible_facts['os_family'] == 'RedHat'
+  ansible.builtin.package:
+    name: openssl-perl
```
fitbeard, line 18: I explicitly run environments on rhel family distros. This is not needed. Please do testing on Rocky Linux9/Alma Linux9/Redhat Linux 9 and not on Fedora because there is no real chance for Fedora in production.

roles/openstack_cli/templates/openrc.j2
```diff
@@ -12,7 +12,11 @@ export OS_PROJECT_DOMAIN_NAME=Default
 export OS_PROJECT_NAME=admin
 
 {% if cluster_issuer_type is defined and cluster_issuer_type in ('self-signed', 'ca', 'venafi') %}
+{% if ansible_os_family == 'Debian' %}
```
fitbeard: Non of this is needed because this is an abstraction from container perspective and `OS_CACERT` is always mounted under `/usr/local/share/ca-certificates/atmosphere.crt` path

roles/openstacksdk/templates/clouds.yaml.j2
```diff
@@ -9,7 +9,11 @@ clouds:
       project_domain_name: Default
     region_name: "{{ openstack_helm_endpoints_keystone_region_name }}"
 {% if cluster_issuer_type is defined and cluster_issuer_type in ('self-signed', 'ca', 'venafi') %}
+    {% if ansible_os_family == 'Debian' %}
```
fitbeard: Same here as above. The magic happened here: https://github.com/vexxhost/atmosphere/blob/main/roles/openstack_cli/templates/atmosphere.sh.j2#L4

roles/cluster_issuer/handlers/main.yml
yaguangtang: the c_rehash is from this pacakge, I tested in rocky 9.4 cloud image.

fitbeard: Ah sorry. My bad. In rhel this not mandatory but let's keep it for consistency.

roles/openstacksdk/templates/clouds.yaml.j2
yaguangtang: so this is needed during the installazation, this credential for the openstack client is used both in the Host OS and the container, and the host OS is redhat familiy , but the container is ubuntu.

roles/openstacksdk/templates/clouds.yaml.j2
```diff
@@ -9,7 +9,11 @@ clouds:
       project_domain_name: Default
     region_name: "{{ openstack_helm_endpoints_keystone_region_name }}"
 {% if cluster_issuer_type is defined and cluster_issuer_type in ('self-signed', 'ca', 'venafi') %}
+{% if ansible_os_family == 'Debian' %}
     cacert: "/usr/local/share/ca-certificates/atmosphere.crt"
+{% elif ansible_os_family == 'RedHat' %}
```
fitbeard, line 14: Not needed.

yaguangtang: From my testing, this is needed during the installazation, this credential for the openstack client is used both in the Host OS and the container, and the host OS is redhat familiy , but the container is ubuntu.

fitbeard (approved):

PR #2854, opened 2025-08-04
Fix mulecule ha cases

ricolin:
Fixing two tasks:
* molecule login is currently broken as we didn't provide docker plugin.
* bash could be broken by the script  on different OS version

Skip-Release-Notes

Change-Id: I13afb238b8f9dbb66bff7ae7c7db893d73b2a798


ricolin:
recheck

PR #2900, opened 2025-09-02
Make manila share BFV

### LINKED ISSUE #2869: Implement BFV for Manila (or use containers)
At the moment, there is no clean and clear way of failing over Manila VMs.  It would be a lot more beneficial to move towards a model that either uses OVN and mounts the pods directly into the customer network, or use BFV on the VMs in case the hypervisor data gets lost.

ricolin:
Make manila shares boot from volume.
ref: https://review.opendev.org/c/openstack/manila/+/959016

Change-Id: I7b622c39909fa43a022366113b2e49c99633750e
Depends-On: https://github.com/vexxhost/manila-service-images/pull/12
fixes: [issues/2869](https://github.com/vexxhost/atmosphere/issues/2869)

### DIFF
--- a/roles/AGENTS.md
+++ b/roles/AGENTS.md
@@ -1,9 +1,10 @@
 # roles/AGENTS.md
 
-## Cluster API version changes
+## Review guidelines
 
-When changing Cluster API, kubeadm bootstrap/control-plane, `clusterctl`, or
-CAPO/OpenStack provider versions under `roles/`:
+Apply these rules both when making changes and when reviewing PRs that change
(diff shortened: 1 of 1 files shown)


mnaser:
@ricolin why does it need to use base volume? why does it not use boot from volume using an image as a base instead and Cinder will simply copy that image in?

ricolin:
We also need to either change manila image to raw format or  turn `image_conversion_disable` to false otherwise will hit this issue when using image as source for BFV

PR #2901, opened 2025-09-03
deps: upgrade k8s collection to 2.3.2

okozachenko1203:
This PR is based on the assumption that k8s collection 2.3.2 will be released after this PR is merged https://github.com/vexxhost/ansible-collection-kubernetes/pull/194

We don't need to keep local patch for capo image build so this PR uses the upstream image

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -6,28 +6,29 @@ variable "TAG" {
     default = "main"
 }
 
-target "cluster-api-provider-openstack-source" {
-    context = "images/source-patch"
-    platforms = ["linux/amd64", "linux/arm64"]
+# NOTE(oleksandr1203): We keep this commented out for now as we'll need patch in the future.
+# target "cluster-api-provider-openstack-source" {
+#     context = "images/source-patch"
+#     platforms = ["linux/amd64", "linux/arm64"]
 
-    contexts = {
-        "git" = "https://github.com/kubernetes-sigs/cluster-api-provider-openstack.git#v0.11.6"
-        "patches" = "patches/kubernetes-sigs/cluster-api-provider-openstack"
-    }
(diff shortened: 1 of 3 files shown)


docker-bake.hcl
```diff
@@ -6,28 +6,29 @@ variable "TAG" {
...
+#     platforms = ["linux/amd64", "linux/arm64"]
+
+#     contexts = {
+#         "source" = "target:cluster-api-provider-openstack-source"
+#     }
+
+#     tags = [
+#         "${REGISTRY}/capi-openstack-controller:${TAG}"
+#     ]
+# }
```
mnaser: Let's remove this instead of commenting it out, we can always check Git history to bring it back.

docker-bake.hcl
```diff
@@ -294,7 +295,7 @@ target "openstack" {
 
 group "default" {
     targets = [
-        "cluster-api-provider-openstack",
+        # "cluster-api-provider-openstack",
```
mnaser: Remove this too

mnaser (changes requested):
This needs a release note.

releasenotes/notes/bump-k8s-collection-6803673d8e60f667.yaml
```diff
@@ -0,0 +1,6 @@
+---
+upgrade:
+  - |
+    Bump Kubernetes collection from 2.0.1 to 2.3.2 and the Cluster
+    API driver for Magnum has been bumped to 0.31.2 to improve
+    stability, fix bugs and add new features.
```
mnaser: Please make this into two different release note files.

mnaser (changes requested):

Issue #2920, opened 2025-09-05
Cinder failed to gain/drop privileges


ricolin:
This happens when try to create cinder boot volume with image convert.
``` 
   File "/var/lib/openstack/lib/python3.12/site-packages/cinder/volume/flows/manager/create_volume.py", line 897, in _prepare_image_cache_entry
     model_update = self._create_from_image_cache_or_download(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/var/lib/openstack/lib/python3.12/site-packages/cinder/volume/flows/manager/create_volume.py", line 981, in _create_from_image_cache_or_download
     with image_utils.TemporaryImages.fetch(
   File "/usr/lib/python3.12/contextlib.py", line 137, in __enter__
     return next(self.gen)
            ^^^^^^^^^^^^^^
   File "/var/lib/openstack/lib/python3.12/site-packages/cinder/image/image_utils.py", line 1394, in fetch
     fetch_verify_image(context, image_service, image_id, tmp)
   File "/var/lib/openstack/lib/python3.12/site-packages/cinder/image/image_utils.py", line 882, in fetch_verify_image
     data = get_qemu_data(image_id, has_meta, format_raw,
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/var/lib/openstack/lib/python3.12/site-packages/cinder/image/image_utils.py", line 721, in get_qemu_data
     data = qemu_img_info(dest,
            ^^^^^^^^^^^^^^^^^^^
   File "/var/lib/openstack/lib/python3.12/site-packages/cinder/image/image_utils.py", line 168, in qemu_img_info
     format_name = cinder.privsep.format_inspector.get_format_if_safe(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/var/lib/openstack/lib/python3.12/site-packages/oslo_privsep/priv_context.py", line 261, in _wrap
     self.start()
   File "/var/lib/openstack/lib/python3.12/site-packages/oslo_privsep/priv_context.py", line 275, in start
     channel = daemon.RootwrapClientChannel(context=self)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2921, opened 2025-09-05
Fix cinder volume container permission

ricolin:
Cinder hit FailedToDropPrivileges error when it try to fetch and verify image from Qemu.
Following error will occur.
"oslo_privsep.daemon.FailedToDropPrivileges: Privsep daemon failed to start" Fix this issue by make container run as root user.

Fixes: https://github.com/vexxhost/atmosphere/issues/2920
Change-Id: Ia230af2d12643513ae798bb0e6aa95d5e01df2e9


okozachenko1203 (approved):

ricolin:
What I mean is we can just bump chart to includes https://review.opendev.org/c/openstack/openstack-helm/+/959677 if we still need this PR

ricolin:
sorry the second close/reopen was a mistake

ricolin:
> I think we can remove the override here https://github.com/vexxhost/atmosphere/blob/main/roles/cinder/vars/main.yml#L34-L38,
> 
> the correct config should be
> 
> ```
>  pod:
>     security_context:
>       cinder_volume:
>         container:
>           cinder_volume:
>             privileged: true
> ```

added https://github.com/vexxhost/atmosphere/pull/2921/commits/a6c48e579cf589831d6384ff4b85e4dcea18fcfc

PR #2950, opened 2025-09-10
ensure necessary kernel modules are loaded

yaguangtang:
kernel module nvme are used in latest os_brick library 
nbd module is required for Trilio snapshot mount and file search

### DIFF
--- a/playbooks/openstack.yml
+++ b/playbooks/openstack.yml
@@ -67,6 +67,10 @@
   hosts: controllers:computes
   become: true
   roles:
+    - role: kernel_modules
+      tags:
+        - kernel_modules
+
     - role: lpfc
       tags:
         - lpfc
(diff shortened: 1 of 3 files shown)


yaguangtang:
> I think it makes more sense to take an approach similar to this:
> 
> https://github.com/openstack/openstack-helm/blob/4172872b086b1e3bedd872c100fe620f5bb59c8a/openvswitch/templates/bin/_openvswitch-vswitchd-init-modules.sh.tpl#L20
> 
> This makes sure that everything continues to be "cloud-native".

yeah, for those tightly required by the openstack services,it's better to load in the chart. 
for this case NBD module is required by Trilio "mount snapshot" and "file search" feature. in compute nodes.
nvme module is required by cinder iCSCI library os_brick

PR #2983, opened 2025-09-16
feat: add manila share type configuration

fitbeard:
Related to: https://github.com/vexxhost/magnum-cluster-api/pull/795/commits/da57db95b933af340fb559b8e170fc92f7502ac9

Also handy if we need more extra specs like:
```
openstack share type create manila-generic-share true \
    --extra-specs share_backend_name=ANOTHER_BACKEND
```

### DIFF
--- a/releasenotes/notes/add-manila-share-types-12fa414e15ef0b6b.yaml
+++ b/releasenotes/notes/add-manila-share-types-12fa414e15ef0b6b.yaml
@@ -1,0 +1,4 @@
+---
+features:
+  - Add support for Manila share types configuration with default extra
+    specs for snapshot support.
--- a/roles/manila/tasks/main.yml
+++ b/roles/manila/tasks/main.yml
@@ -52,3 +52,53 @@
     gigabytes: -1
     security_group: -1
     security_group_rule: -1
+
+- name: Create share types
+  when: manila_share_types | length > 0
+  block:
+    - name: Wait until share service ready
+      kubernetes.core.k8s_info:
+        api_version: apps/v1
+        kind: Deployment
+        name: manila-api
+        namespace: openstack
+        wait_sleep: 10
+        wait_timeout: 600
+        wait: true
+        wait_condition:
+          type: Available
+          status: true
+
+    - name: Create share types
+      changed_when: false
+      ansible.builtin.shell: |
+        set -o posix
+        source /etc/profile.d/atmosphere.sh
+        openstack share type show {{ item.name }} ||
+        openstack share type create \
+          --public {{ item.is_public | default(true) }} \
+          {{ item.name }} {{ item.driver_handles_share_servers | default(true) }}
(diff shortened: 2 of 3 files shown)


roles/manila/tasks/main.yml
```diff
@@ -52,3 +52,53 @@
...
+          status: true
+
+    - name: Create share types
+      changed_when: false
+      ansible.builtin.shell: |
+        set -o posix
+        source /etc/profile.d/atmosphere.sh
+        openstack share type show {{ item.name }} ||
+        openstack share type create \
+          --public {{ item.is_public | default(true) }} \
```
ricolin: nit: is this necessary or optional when creating type? as we gonna do it in line 86 below

releasenotes/notes/add-manila-share-types-12fa414e15ef0b6b.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - Add support for Manila share types configuration with default extra
+    specs for snapshot support.
```
ricolin, line 4: would be nice to mention a default public share type will also be created

fitbeard: Default share type is created by chart during bootstrap https://github.com/vexxhost/atmosphere/blob/main/charts/manila/templates/bin/_bootstrap.sh.tpl#L54
and we also rely on it by default: https://github.com/vexxhost/atmosphere/blob/main/charts/manila/values.yaml#L651

roles/manila/tasks/main.yml
fitbeard: Fixed

fitbeard: @ricolin @mnaser

roles/manila/tasks/main.yml
```diff
@@ -52,3 +52,52 @@
...
+      args:
+        executable: /bin/bash
+      environment:
+        OS_CLOUD: atmosphere
+      loop: "{{ manila_share_types }}"
+      register: configure_manila_share_types
+      retries: 60
+      delay: 5
+      until: configure_manila_share_types.rc == 0
+      failed_when: configure_manila_share_types.rc != 0
```
mnaser: I don't feel great about this, while we're done this a few times, I fear that it's gonna become "normal" practice.

Instead, why don't we create a module called `share_type` that is probably going to be not too far out from https://opendev.org/openstack/ansible-collections-openstack/src/branch/master/plugins/modules/volume_type.py in terms of code -- we can push that upstream, add some simple tests, and once the tests pass upstream, we can embed it into this repository while we wait for it to make it's way out upstream.

fitbeard: We did it! https://review.opendev.org/c/openstack/ansible-collections-openstack/+/963259

PR #2990, opened 2025-09-19
feat: upgrade htk version to 2025.1

okozachenko1203:


mnaser (changes requested):
https://github.com/vexxhost/atmosphere/pull/2815 maybe we can continue to work on that one if you can review it?

PR #2991, opened 2025-09-19
fix pyopenssl dependency

ricolin:
Change-Id: I669341a7b5cc9560a4831080a203586dad5c908a


mnaser (changes requested):
I'm not a big fan of this, this has possibility of breaking the node entirely that it will uninstall other things.

Why don't we remove the whole PyOpenSSL upgrade entirely?

PR #3000, opened 2025-09-19
enable neutron taas service plugin for port mirroring

yaguangtang:


roles/neutron/vars/main.yml
```diff
@@ -83,6 +83,17 @@ __neutron_helm_values:
...
+        taas:
+          enabled: true
+    taas_plugin:
+      service_providers:
+        service_provider: >-
+          {% if atmosphere_network_backend == 'ovn' %}
+          TAAS:TAAS:neutron_taas.services.taas.service_drivers.ovn.taas_ovn.TaasOvnDriver:default
+          {% else %}
+          TAAS:TAAS:neutron_taas.services.taas.service_drivers.taas_rpc.TaasRpcDriver:default
+          {% endif %}
```
mnaser: Instead of Jinja, please use the `TaasRpcServer` here by default and add the `taas` specific configuration for OVN below.

mnaser (changes requested):

roles/neutron/vars/main.yml
```diff
@@ -114,6 +120,12 @@ __neutron_ovn_helm_values:
         ml2:
           type_drivers: flat,vlan,geneve
           tenant_network_types: geneve
+      taas:
+        taas:
+          enabled: true
```
mnaser: Since this section gets merged into the original one, having this here is redundant.

The usage of this variable is not one or the other but it starts with the first and overlays this if OVN is in use.

mnaser (changes requested):

roles/neutron/vars/main.yml
yaguangtang: indeed

roles/neutron/vars/main.yml
```diff
@@ -114,6 +120,9 @@ __neutron_ovn_helm_values:
         ml2:
           type_drivers: flat,vlan,geneve
           tenant_network_types: geneve
+    taas_plugin:
+      service_providers:
+        service_provider: TAAS:TAAS:neutron_taas.services.taas.service_drivers.taas_rpc.TaasRpcDriver:default
```
mnaser: I think this should be the ovn driver no?

mnaser (changes requested):

roles/neutron/vars/main.yml
yaguangtang: My bad, copied without changing:(

mnaser (approved):

PR #3035, opened 2025-09-24
[stable/2025.1]fix: update neutron chart to fix openvswitch config missing

yaguangtang:
fix extra neutron config files aren't loaded using uwsgi start script for openvswitch network backend


mnaser:
@yaguangtang is this a backport of a change that was pushed into `main`?

PR #3080, opened 2025-10-09
feat: bump rabbitmq resources

fitbeard:
Note: In RabbitMQ `4.x` `HA` queues are completely removed BUT we never used them because...:) BUT some charts has default policies under `rabbitmq.policies` with settings for `ha_queues` which are not valid for `4.x` and prevents init jobs to finish. We need to override defaults and get rid of settings related to ha_queues.

### DIFF
--- a/releasenotes/notes/fix-octavia-amphora-alert-exclude-deleted-status-12138982a4f3a4a1.yaml
+++ b/releasenotes/notes/fix-octavia-amphora-alert-exclude-deleted-status-12138982a4f3a4a1.yaml
@@ -1,0 +1,4 @@
+---
+fixes:
+  - |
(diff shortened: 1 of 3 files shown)


releasenotes/notes/bump-rabbitmq-components-2fa5862e1e5be302.yaml
mnaser, line 1: - Upgrades from RabbitMQ 3.13.x are supported to both RabbitMQ 4.0.x and RabbitMQ 4.1.x.

We should be good then.

fitbeard: Added

PR #3108, opened 2025-10-14
feat: allow set custom args for openstack exporter

fitbeard:
We have customers who wants to disable some services, slow metrics or enable more extra arguments.


fitbeard:
@mnaser switched to `latest` tag to enable `--cache` support. By default https://github.com/openstack-exporter/openstack-exporter/blob/main/main.go#L110 metric collecting happens every 150s (`cache-ttl/2`) and cache flush every 300s (`cache-ttl`) when `--cache-ttl` is unset.

PR #3143, opened 2025-10-21
feat: update image dependencies for 2025.2

fitbeard:


mnaser (changes requested):
This needs to be pushed to main first and then backported to stabel/2025.2

PR #3173, opened 2025-10-31
[Cinder] Bump cinder to 2025.2.0

ricolin:
Change-Id: Ief6b754b49fb11964731231ec41c15baf131baab


roles/cinder/vars/main.yml
```diff
@@ -71,6 +71,14 @@ __cinder_helm_values:
           priority: 0
           apply-to: all
           pattern: '^(?!(amq\.|reply_)).*'
+  secrets:
```
ricolin, line 67: without these we got error

```
"Error: UPGRADE FAILED: rendered manifests contain a resource that already exists. Unable to continue with update: could not get information about the resource Secret \"\" in namespace \"openstack\": resource name may not be empty"
```

roles/openstack_helm_endpoints/vars/main.yml
```diff
@@ -68,6 +68,9 @@ _openstack_helm_endpoints_identity:
         region_name: "{{ openstack_helm_endpoints_keystone_region_name }}"
         username: "admin-{{ openstack_helm_endpoints_keystone_region_name }}"
         password: "{{ openstack_helm_endpoints_keystone_admin_password }}"
+        project_name: admin
+        user_domain_name: default
+        project_domain_name: default
```
ricolin, line 73: without these we got error

"Error: template: cinder/templates/secret-keystone.yaml:28:42: executing \"cinder/templates/secret-keystone.yaml\" at <include \"helm-toolkit.snippets.keystone_secret_openrc\">: error calling include: template: cinder/charts/helm-toolkit/templates/snippets/_keystone_secret_openrc.tpl:23:62: executing \"helm-toolkit.snippets.keystone_secret_openrc\" at <b64enc>: invalid value; expected string",

ricolin: due to commit bfbf0c3b6db5824394bfec3f25887f89e5730fe0

roles/cinder/vars/main.yml
ricolin: this could be something starts since helm v2

ricolin:
Closing as obsolete: Cinder 2025.1/2025.2 image digests and chart configurations have already landed on main.

PR #3176, opened 2025-11-03
support FwaaS plugin for neutron with OVN

yaguangtang:


mnaser:
replaced by https://github.com/vexxhost/atmosphere/pull/3712

PR #3186, opened 2025-11-04
fix stable/2025.2 aio ci job

ricolin:


mnaser:
lets use https://github.com/vexxhost/atmosphere/pull/3223

PR #3214, opened 2025-11-06
fix: Deploy ORC components in control-plane

larainema:


releasenotes/notes/deploy-orc-components-in-control-plane-7f6e32c0b8c14c67.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    Deploy OpenStack resource controller in control-plane.
```
mnaser: Can you make this a bit more clear?  Imagine how this would be listed here:

https://vexxhost.github.io/atmosphere/releasenotes/2025.2.html

PR #3252, opened 2025-11-10
fix: [valkey]fix missing resources limit

ricolin:
Change-Id: Ia8a19560f697ef04ce8ad1ec443004f13adda792

### DIFF
--- a/roles/valkey/vars/main.yml
+++ b/roles/valkey/vars/main.yml
@@ -33,6 +33,9 @@ _valkey_helm_values:
       repository: "{{ atmosphere_images['valkey_sentinel'] | vexxhost.kubernetes.docker_image('path') }}"
       tag: "{{ atmosphere_images['valkey_sentinel'] | vexxhost.kubernetes.docker_image('tag') }}"
     primarySet: atmosphere
+    resources:
+      limits:
+        cpu: 1000m
   tls:
     enabled: true
     existingSecret: "{{ valkey_server_certificate_secret_name }}"
@@ -45,6 +48,9 @@ _valkey_helm_values:
       registry: "{{ atmosphere_images['redis_exporter'] | vexxhost.kubernetes.docker_image('domain') }}"
       repository: "{{ atmosphere_images['redis_exporter'] | vexxhost.kubernetes.docker_image('path') }}"
       tag: "{{ atmosphere_images['redis_exporter'] | vexxhost.kubernetes.docker_image('tag') }}"
+    resources:
+      limits:
+        cpu: 1000m
   primary:
     resources:
       requests:


yaguangtang (approved):

fitbeard (approved):

roles/valkey/vars/main.yml
```diff
@@ -33,6 +33,13 @@ _valkey_helm_values:
       repository: "{{ atmosphere_images['valkey_sentinel'] | vexxhost.kubernetes.docker_image('path') }}"
       tag: "{{ atmosphere_images['valkey_sentinel'] | vexxhost.kubernetes.docker_image('tag') }}"
     primarySet: atmosphere
+    resources:
+      requests:
+        cpu: 500m
+        memory: 512Mi
+      limits:
+        cpu: 1000m
```
mnaser, line 38: Can you test increasing the cpu limit only and nothing else?  The requests don't really matter in this case, and the memory limit is not really an issue.

roles/valkey/vars/main.yml
```diff
@@ -45,6 +52,13 @@ _valkey_helm_values:
       registry: "{{ atmosphere_images['redis_exporter'] | vexxhost.kubernetes.docker_image('domain') }}"
       repository: "{{ atmosphere_images['redis_exporter'] | vexxhost.kubernetes.docker_image('path') }}"
       tag: "{{ atmosphere_images['redis_exporter'] | vexxhost.kubernetes.docker_image('tag') }}"
+    resources:
+      requests:
+        cpu: 500m
+        memory: 512Mi
+      limits:
+        cpu: 1000m
```
mnaser, line 53: Same here.

mnaser (changes requested):
Just need to make sure all this applies

mnaser (commented):
Can you see if we have metrics-server running on an environment so that we can get long term statistics of the ceiling of this component?  Alternatively, we actually can probably pull it from Prometheus to see how much CPU usage it does, specifically in busy environments.

ricolin:
recheck

ricolin:
> Can you see if we have metrics-server running on an environment so that we can get long term statistics of the ceiling of this component? Alternatively, we actually can probably pull it from Prometheus to see how much CPU usage it does, specifically in busy environments.

I checked 10 days of Prometheus data across three Valkey pods. The workload was light, so this does not establish an exact production ceiling. The proposed 1000m limit is a conservative burst ceiling matching the Valkey containers, while requests remain unchanged.
```
   Container           Average CPU     p99 CPU    Maximum CPU    p99 throttled periods    Current limit
  ━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━  ━━━━━━━━━━  ━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━
   Sentinel                 41–42m      42–43m            50m                   23–24%             150m
  ──────────────────  ─────────────  ──────────  ─────────────  ───────────────────────  ───────────────
   Metrics exporter       1.9–2.6m    2.1–2.9m           5.7m                   35–47%             150m

```

PR #3258, opened 2025-11-13
chore: tunning neutron format/config

ricolin:
Correct some typo when makeing chart patch.
Also correct config location for VPN service_providers config

Change-Id: I4ed1eb7f0c244c5da2a3e57b847d7699f7f53f77


ricolin:
recheck

PR #3297, opened 2025-11-21
[A8E-62] Fix openstacksdk install issue with pip

larainema:


roles/openstacksdk/tasks/main.yml
```diff
@@ -12,10 +12,37 @@
...
+    (ansible_facts['distribution'] == 'Ubuntu' and ansible_facts['distribution_major_version'] is version('22', '<=')) or
+    (ansible_facts['distribution'] == 'Debian' and ansible_facts['distribution_major_version'] is version('12', '<=')) or
+    (ansible_facts['distribution'] in ['AlmaLinux', 'RedHat', 'Rocky'] and ansible_facts['distribution_major_version'] is version('8', '<='))
+  block:
+    - name: Install PIP
+      ansible.builtin.package:
+        name:
+          - python3-pip
+          - python3-cryptography
+
```
mnaser: No need for this

roles/openstacksdk/tasks/main.yml
```diff
@@ -12,10 +12,37 @@
...
-- name: Install openstacksdk
-  ansible.builtin.pip:
-    name: openstacksdk
-    version: "{{ openstacksdk_version | default(omit) }}"
+- name: Install openstacksdk using PIP
+  when: >
+    (ansible_facts['distribution'] == 'Ubuntu' and ansible_facts['distribution_major_version'] is version('22', '<=')) or
+    (ansible_facts['distribution'] == 'Debian' and ansible_facts['distribution_major_version'] is version('12', '<=')) or
+    (ansible_facts['distribution'] in ['AlmaLinux', 'RedHat', 'Rocky'] and ansible_facts['distribution_major_version'] is version('8', '<='))
+  block:
```
mnaser: Which means no need for a block

roles/openstacksdk/tasks/main.yml
```diff
@@ -12,10 +12,37 @@
...
+- name: Install openstacksdk using package manager
+  when: >
+    (ansible_facts['distribution'] == 'Ubuntu' and ansible_facts['distribution_major_version'] is version('22', '>')) or
+    (ansible_facts['distribution'] == 'Debian' and ansible_facts['distribution_major_version'] is version('12', '>')) or
+    (ansible_facts['distribution'] in ['AlmaLinux', 'RedHat', 'Rocky'] and ansible_facts['distribution_major_version'] is version('8', '>'))
+  block:
+    - name: Enable EPEL repository
+      ansible.builtin.package:
+        name: epel-release
+      when: ansible_facts['os_family'] == "RedHat"
```
mnaser: No need for this

releasenotes/notes/fix-openstacksdk-install-issue-73ee7bf931a2bdad.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Fix ``openstacksdk`` install issue with pip, using ``Ansible.builtin.package``
+    instead of ``Ansible.builtin.pip``.
```
mnaser: We should explain the why/what this is happening, and you can update the dictionary to have ansible with lower case so it shows the right thing.

mnaser (changes requested):

mnaser:
> None of this will work without installing additional repository for Openstack packages. And if we do then we will back again from where we started - additional apt/rpm repo management and package conflicts. Plus additional Ansible tasks for packages which needs to be absent durring major upgrades. There hould be another way. Please please do not go back.

How much packages are we actually adding if we are just installing `python-openstacksdk`?  Also, do we really need an extra repository?  It's my assumption that we wouldn't need much more if it's a new enough version of Debian or Ubuntu?

We don't really have much of an alternative in the short term to resolve this I think?

fitbeard:
> > None of this will work without installing additional repository for Openstack packages. And if we do then we will back again from where we started - additional apt/rpm repo management and package conflicts. Plus additional Ansible tasks for packages which needs to be absent durring major upgrades. There hould be another way. Please please do not go back.
> 
> How much packages are we actually adding if we are just installing `python-openstacksdk`? Also, do we really need an extra repository? It's my assumption that we wouldn't need much more if it's a new enough version of Debian or Ubuntu?
> 
> We don't really have much of an alternative in the short term to resolve this I think?

Debian 13: https://gist.github.com/fitbeard/fdf1b6ba5b73c804c760963a9f96e749 <- no need for an extra repo.
RHEL: https://gist.github.com/fitbeard/cebe54310ca55ae6ade03c07f7351fa8 <- official way using meta rpm package looks horrible + you have to select which OS release you want to use. This can be mitigated by using copy of created openstack.repo file. In this case this stupid depencencie chain could be skipped:
```
Dependencies resolved.
============================================================================================================================================================================================================================================================================================================================
 Package                                                                                          Architecture                                                         Version                                                                   Repository                                                            Size
============================================================================================================================================================================================================================================================================================================================
Installing:
 centos-release-openstack-dalmatian                                                               noarch                                                               1-1.el9                                                                   extras                                                               7.9 k
Installing dependencies:
 centos-release-ceph-reef                                                                         noarch                                                               1.0-1.el9                                                                 extras                                                               7.4 k
 centos-release-cloud                                                                             noarch                                                               1-1.el9                                                                   extras                                                               7.9 k
 centos-release-messaging                                                                         noarch                                                               1-4.el9                                                                   extras                                                               8.4 k
 centos-release-nfv-common                                                                        noarch                                                               1-6.el9                                                                   extras                                                               7.8 k
 centos-release-nfv-openvswitch                                                                   noarch                                                               1-6.el9                                                                   extras                                                               7.6 k
 centos-release-rabbitmq-38                                                                       noarch                                                               1-4.el9                                                                   extras                                                               7.4 k
 centos-release-storage-common                                                                    noarch                                                               2-5.el9                                                                   extras                                                               8.4 k
```

fitbeard:
> @fitbeard @mnaser , I updated the PR for rhel related env still use pip , if this is doable?

LGTM:
We droped RHEL8 support even before we started official rhel support.
RHEL10 still not fully tested. Me personally never ran full blown Atmosphere installation on RHEL10 and i'm sure there could be more issues than just a single package installation issue.

ricolin (approved):

mnaser:
I'm pulling this into https://github.com/vexxhost/atmosphere.common/pull/114

PR #3314, opened 2025-12-02
feat(A8E-72): add initial roles for cloudnative-pg

fitbeard:
I need initial review here.

What's missing:

- Finalize Cluster CR
- Write documentation about backup/restore
- Decide about reconciliation vs Ansible management for databases and users


larainema:
@fitbeard , I reviewed this against the Keycloak PostgreSQL migration work. I think the managed-CNPG path needs a follow-up before it can be used for Keycloak:

- choose the Kubernetes/CNPG support baseline; Atmosphere currently defaults to Kubernetes 1.28.13, while current supported CNPG releases require newer Kubernetes
- fix the operator/plugin version and kubectl-cnpg checksum alignment
- make CNPG deployment opt-in; the current infrastructure playbook creates a generic openstack/cnpg-db cluster even while Keycloak remains on MySQL/PXC
- finalize Keycloak database/user/secret ownership and add backup/restore resources and documentation

Can you help confirm whether the next managed-CNPG follow-up should target a post-Kubernetes-upgrade baseline, and whether database/users should be reconciled through CNPG resources or Ansible? I suggest keeping this PR as the CNPG foundation and proceeding with the external PostgreSQL migration in #4134 independently.

fitbeard:
@larainema many of these questions are mentioned here long time ago: https://vexxhost.atlassian.net/browse/A8E-72?focusedCommentId=29340 . All this will be addressed later i think when we have PXC -> PG migration plan. By the way nice finding about Kubernetes version too low.

PR #3348, opened 2025-12-09
Implement A8E-64 Service Account Support

mnaser:
Enable service account authentication for the Keystone-Keycloak identity driver using the existing `keystone` Keycloak client.

Changes:
- Enable service accounts on the keystone Keycloak client
- Assign realm-management roles to the service account for user management
- Update identity driver config to use client credentials auth

[A8E-64]

[A8E-64]: https://vexxhost.atlassian.net/browse/A8E-64?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


roles/keystone/tasks/main.yml
```diff
@@ -83,13 +83,36 @@
     realm: "{{ item.keycloak_realm }}"
     client_id: "{{ item.keycloak_client_id }}"
     secret: "{{ item.keycloak_client_secret }}"
+    service_accounts_enabled: true
```
fitbeard, line 88: You need to set more options for this to work:
```
    client_authenticator_type: "client-secret"
    public_client: false
    service_accounts_enabled: true
    direct_access_grants_enabled: false
```
Without explicitly disabling public client with API calls Keycloak UI will not show Credentials tab of a client. To remove this confusion set public client to false (enable `Client authentication`).
We want to make this secure as much as possible - we don't need direct grants here which are enabled by default.

roles/keystone/tasks/main.yml
```diff
@@ -83,13 +83,36 @@
...
+    auth_username: "{{ item.keycloak_admin_user }}"
+    auth_password: "{{ item.keycloak_admin_password }}"
+    validate_certs: "{{ cluster_issuer_type != 'self-signed' }}"
+    # Role mapping settings
+    realm: "{{ item.keycloak_realm }}"
+    service_account_user_client_id: "{{ item.keycloak_client_id }}"
+    client_id: realm-management
+    roles:
+      - name: view-users
+      - name: query-groups
```
fitbeard: Role `query-groups` is not available for client `realm-management` and it's not needed. Role `view-users` from client `realm-management` is enough to get user/group info.

fitbeard:
@mnaser almost:)

PR #3389, opened 2026-01-04
fix cinder volume container permission issue

yaguangtang:
https://github.com/openstack/openstack-helm/blob/master/cinder/values.yaml#L161-L172


ricolin (approved):
This also already fixed in https://github.com/openstack/openstack-helm/blob/master/cinder/values.yaml#L172 and https://github.com/vexxhost/atmosphere/pull/3173
which we get the fix by update cinder chart and remove this override.
But before that's merged, this fix seems quick and valid

PR #3386, opened 2025-12-19
fix: improve ironic and fix flat network

ricolin:
This PR introduce: 
- Add common kernal params: Add nofb, console=ttyS0,115200 to ironic kernel_append_params
- Fix flat ironic network config for multi-tenant network
- Allow using system scope authentication from OS-CLOUD

PR #3465, opened 2026-01-22
fix: Ironic image should be public

ricolin:
This change sets public visibility only for the Ironic Python Agent deploy images uploaded by the `roles/ironic` role.

Affected deploy images from config:

- `ironic_python_agent_deploy_kernel_name`
- `ironic_python_agent_deploy_ramdisk_name`

This does not make tenant instance images public. It only makes the operator-owned IPA deploy kernel and ramdisk visible so Ironic can boot the agent during deploy, cleaning, inspection, rescue, and clean-step reboot flows.

Change-Id: I852d4a75fe289efdc3f7d6e4f1f237c8292ed3a2

### DIFF
--- a/.github/styles/config/vocabularies/Base/accept.txt
+++ b/.github/styles/config/vocabularies/Base/accept.txt
@@ -5,6 +5,7 @@ ACME
 ALLOCATED
 AMD
 AVX
+Agent
 Ansible
 BGP
 Bitnami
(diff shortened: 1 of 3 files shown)


yaguangtang (approved):

mnaser (changes requested):
@ricolin how much it needs to be public:?  In theory, the image should be uploaded to where the Ironic user lives as it is the only consumer of it.

ricolin:
I added the solution to https://github.com/vexxhost/atmosphere/pull/3386 so find a better solution to change the image owner to "service" so ironic don't get the issue to find that image

PR #3469, opened 2026-01-28
fix(horizon): bump image to include port security fix

larainema:
## Summary
- Update horizon image to include the upstream fix for public port creation
- The previous image was built on Dec 2, 2025, before the fix was merged to docker-horizon on Dec 3, 2025

## Root Cause Analysis
The atmosphere stable/2024.2 branch was using a horizon image that was built before the upstream fix (Dont send enable_port_security when disallowed by policy) was included in docker-horizon.

The fix prevents sending enable_port_security when disallowed by policy, allowing non-admin users to create ports without policy errors.

Closes A8E-90

🤖 Generated with [Claude Code](https://claude.ai/code)


larainema:
> I think when there is a new stable release of atmosphere, those image hash will update to the latest automatically.

Thanks @yaguangtang , but I can see the hash didn't be changed for example from release 7.1.0 to 7.2.0

PR #3587, opened 2026-02-09
fix: [VPNaaS][OVN] fix neutron ovn vpn agent frequently restart

ricolin:


roles/neutron/vars/main.yml
yaguangtang, line 1: I think it's better to 
`deployment_rpc_server: true`
so that for existing env, an upgrade will take effect.

guilhermesteinmuller: Actually, that is needed for environments where the fix is not shipped. If the user wantsto workaround, they should explicitly override it to true

mnaser (approved):

PR #3638, opened 2026-02-17
fix(kube_prometheus_stack): add burn-rate inhibition rules for NGINX Ingress alerts

mnaser:
When multiple burn-rate tiers fire simultaneously for the same service (e.g., Critical and High both firing for `nova-api`), operators receive redundant notifications.

This adds Alertmanager inhibition rules so that higher-tier burn-rate alerts suppress lower-tier ones for the same service, reducing alert noise.

Also updates the alerts instructions to document that inhibition rules should be considered whenever creating tiered or graduated alert sets.

### DIFF
--- a/roles/kube_prometheus_stack/vars/main.yml
+++ b/roles/kube_prometheus_stack/vars/main.yml
@@ -42,18 +42,21 @@ _kube_prometheus_stack_helm_values:
             - alertname =~ "NginxIngress(High|Moderate|Low)ErrorBudgetBurn"
           equal:
             - service
+            - namespace
         - source_matchers:
             - alertname = "NginxIngressHighErrorBudgetBurn"
           target_matchers:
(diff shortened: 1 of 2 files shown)


roles/kube_prometheus_stack/vars/main.yml
```diff
@@ -35,6 +35,25 @@ _kube_prometheus_stack_helm_values:
...
+          target_matchers:
+            - alertname =~ "NginxIngress(Moderate|Low)ErrorBudgetBurn"
+          equal:
+            - service
+        - source_matchers:
+            - alertname = "NginxIngressModerateErrorBudgetBurn"
+          target_matchers:
+            - alertname = "NginxIngressLowErrorBudgetBurn"
+          equal:
+            - service
```
mnaser, line 58: @copilot i believe these rules should maybe also include namespace in the matchers so if a service has the same name in two diff namespaces it wouldn't be surpressed.

PR #3672, opened 2026-02-20
feat(monitoring): add transceiver-exporter role

yaguangtang:
## Summary

Add a new Ansible role to deploy the transceiver-exporter Helm chart, a Prometheus exporter for network transceivers (SFP, SFP+, QSFP). The chart deploys a DaemonSet on all servers with hostNetwork access and NET_ADMIN capability to read transceiver EEPROM data.

## Changes

- Vendored `transceiver-exporter` Helm chart (v0.1.2) from [yaguangtang/transceiver-exporter](https://github.com/yaguangtang/transceiver-exporter)
- Created `roles/transceiver_exporter` role following existing patterns (goldpinger, ingress_nginx)
- Added image reference to `roles/defaults/vars/main.yml`
- Added role to `playbooks/monitoring.yml` with conditional deployment

## Usage

Deployment is **disabled by default**. To enable:

```yaml
atmosphere_transceiver_exporter_enabled: true
```

When enabled, the exporter runs as a DaemonSet on all servers, exposing optical power and transceiver health metrics on port 9458. A ServiceMonitor is enabled by default for Prometheus scraping.

## Jira

A8E-85


mnaser (commented):
I checked this against current `main`: Atmosphere already enables the `node-exporter` `ethtool` collector, and that data is shipped through the existing kube-prometheus-stack node-exporter scrape path.

That said, I do not think the existing collector makes this exporter redundant. The `node_exporter` `ethtool` collector in the version currently pinned by Atmosphere collects driver info, link capabilities/advertised speeds/port/autonegotiation data, and driver stats via ethtool. It does not read module EEPROM/DDM data or export the optical transceiver telemetry this exporter exposes, such as module temperature, module voltage, laser bias, RX/TX optical power, thresholds, vendor identity/serial/OUI, wavelength, and supported link lengths.

So if the intended value here is optical transceiver health, this PR is adding data we do not get from the existing `node-exporter` `ethtool` collector. If the only desired signals were generic driver/link/stat counters, then the existing collector would be enough.

Separate from that comparison, if this moves forward the image reference still needs the usual digest pinning in `roles/defaults/vars/main.yml`.

PR #3697, opened 2026-02-26
Add IPMI SEL alerts for memory and CPU errors

noskovao:
**Add IPMI SEL alerts for memory and CPU errors**

- Add SEL event matchers for uncorrectable memory and CPU errors.
- Create P1 alerts for those SEL conditions.
- Add alert tests and documentation entries.

### DIFF
--- a/.github/styles/config/vocabularies/Base/accept.txt
+++ b/.github/styles/config/vocabularies/Base/accept.txt
@@ -15,8 +15,10 @@ CRD
 CSI
 Cinder
 CoreDNS
+DIMM
 DNS
 Dell
+ECC
 FQDN
 FQDNs
 Galera
@@ -56,6 +58,7 @@ READY
 RGW
 RPC
 RabbitMQ
+SEL
 SLOs?
 SMART
 SQLAlchemy
--- a/doc/source/admin/monitoring.rst
+++ b/doc/source/admin/monitoring.rst
@@ -908,6 +908,82 @@ experiencing network issues.
      kubectl debug node/<affected-node> -it --image=busybox -- \
        cat /host/var/log/syslog | tail -100
 
+``IpmiUncorrectableMemoryError``
+================================
+
+This alert fires when the Intelligent Platform Management Interface (IPMI)
+exporter reports a recent ``uncorrectable_memory_error`` System Event Log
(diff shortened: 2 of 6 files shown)


.github/styles/Microsoft/Acronyms.yml
mnaser, line 1: Instead of exceptions, we should add the words to the `accept.txt` instead.

roles/ipmi_exporter/tasks/main.yml
```diff
@@ -40,19 +40,23 @@
...
-                config-hash: "{{ ipmi_exporter_config | to_yaml | hash('md5') }}"
+                config-hash: "{{ ipmi_exporter_config | to_nice_yaml(indent=2) | hash('md5') }}"
               labels:
                 application: ipmi-exporter
                 job: ipmi
             spec:
               containers:
                 - name: exporter
+                  args:
+                    - --config.file=/config.yml
```
mnaser, line 51: Is this needed? does that mean it never actually used a config file before? o_O

mnaser (changes requested):

roles/ipmi_exporter/tasks/main.yml
noskovao: Yes, we need it. I believe in your image, you have it in docker file as well as runAsUser and runAsGroup

.github/styles/Microsoft/Acronyms.yml
noskovao: I was pretty sure that these are acronyms (an abbreviation formed from the initial letters of other words and pronounced as a word)

mnaser: Yeah, however this file is actually the list of words that come directly from the Microsoft technical writing terms, so we're taking them "as-is" since they may be overwritten.

mnaser (changes requested):

.github/styles/Microsoft/Acronyms.yml
noskovao: Ok I've moved it to accept.txt

noskovao:
Hey @mnaser, can you review this change again?
Do you think there is a benefit to having this merged?

PR #3712, opened 2026-03-03
feat(neutron): add support for Neutron FWaaS

yaguangtang:
- Enabled Neutron FWaaS by default for OVN deployments instead of introducing a separate role-level feature flag.
- Wired FWaaS through the Neutron OVN defaults:
  - adds `firewall_v2` to the default OVN service plugins
  - configures the OVN FWaaS driver
  - mounts `neutron_fwaas.conf` for neutron-server, neutron-rpc-server, and db-sync
  - runs the `neutron-fwaas` DB migrations
- Enabled the Horizon FWaaS dashboard by default so the UI matches the backend behavior.
- Added tests for the Neutron role/chart wiring and Horizon panel defaults.
- Added deployment docs and a release note for the new default behavior.

## Related

- Docker image support: https://github.com/vexxhost/docker-neutron/pull/634


roles/neutron/defaults/main.yml
```diff
@@ -42,3 +42,6 @@ neutron_policy_server_integration_enabled: true
 #   - name: switch1_key_file_name
 #     key: <private ssh key content>
 neutron_genericswitch_keys: []
+
+# Enable Neutron Firewall as a Service (FWaaS)
+neutron_fwaas_enabled: false
```
mnaser: Is there a reason not to just make it enable out of the box without a way to disable it?

mnaser (changes requested):
small comment but cuirious why there is a bunch of other changes, we should fix that

yaguangtang:
The CI fails due to https://github.com/vexxhost/docker-neutron/pull/1962 needed.

yaguangtang:
@mnaser  The CI job fails due to neutron plugins aren't updated for a long time,  https://github.com/vexxhost/docker-neutron/pull/1327

yaguangtang:
Depends on https://github.com/vexxhost/docker-tempest/pull/198 to pass tempest neutron plugin test cases.

PR #3784, opened 2026-04-01
ci(keycloak): migrate keycloak molecule job to Zuul

larainema:
## Summary

The keycloak molecule job was running on a self-hosted GitHub Actions runner (`vexxhost-ubuntu-22.04-4`) that has been unavailable, causing all CI runs to be stuck in 'queued' state indefinitely since March 17, 2026.

This migrates the keycloak molecule job to Zuul, following the same pattern used for AIO and CSI molecule jobs (PR #3671).

## Changes

- **Removed** the `keycloak` job from `.github/workflows/ci.yaml` (no longer depends on the unavailable self-hosted runner)
- **Added** `atmosphere-molecule-keycloak` job to `.zuul.yaml` with appropriate single-node resource reduction vars
- **Added** the new job to both `check` and `gate` pipelines

## Impact

This will unblock all GitHub Actions CI runs that were previously stuck waiting for the self-hosted runner. The keycloak molecule test will continue to run, but now via Zuul infrastructure.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -274,7 +274,6 @@
     name: atmosphere-molecule-keycloak
     parent: atmosphere-molecule
     timeout: 7200
-    nodeset: atmosphere-ubuntu-jammy-single-node-16
     vars:
       molecule_scenario: keycloak
       csi_driver: local-path-provisioner
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -273,7 +273,6 @@
 - job:
     name: atmosphere-molecule-keycloak
     parent: atmosphere-molecule
-    timeout: 7200
     vars:
       molecule_scenario: keycloak
       csi_driver: local-path-provisioner
@@ -292,9 +291,6 @@
         controller:
           config:
             worker-processes: 2
-      valkey_helm_values:
-        replica:
-          replicaCount: 1
 
 - project:
     check:
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -270,6 +270,33 @@
     vars:
       atmosphere_network_backend: ovn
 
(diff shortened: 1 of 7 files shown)


larainema:
recheck

mnaser:
@larainema this looks good, the only challenge is that its using a 16c/64g vm which is _overkill_ .. can you check what other labels are available in zuul, and maybe i might need to add a new one?

larainema:
> Dong Ma (@larainema) this looks good, the only challenge is that its using a 16c/64g vm which is _overkill_ .. can you check what other labels are available in zuul, and maybe i might need to add a new one?

I think there is default ubuntu-jammy label can be used

.zuul.yaml
```diff
@@ -270,16 +270,44 @@
...
+        pod:
+          replicas:
+            api: 1
+      ingress_nginx_helm_values:
+        controller:
+          config:
+            worker-processes: 2
+      valkey_helm_values:
+        replica:
+          replicaCount: 1
```
mnaser: @copilot Zuul layers variables one on top of the other, since the parent of this job is `atmosphere-molecule` and it has a ton of variables already, can you make sure which ones in here are actually necessary?

also, is the timeout reasonable here?  maybe we can drop it, the job finished in around 20 minutes.

mnaser (changes requested):
small cleanups, but it is passing.

.zuul.yaml
mnaser: @copilot aren't those variables pre-existing in the parent atmosphere-molecule job?

mnaser:
recheck

larainema:
recheck

PR #3839, opened 2026-04-16
fix(openvswitch): add /var/tmp mount to ovs-vswitchd container

larainema:
Backport upstream change [966771](https://review.opendev.org/c/openstack/openstack-helm/+/966771) to add `/var/tmp` mount to the `ovs-vswitchd` container. This ensures proper operation when using DPDK with the Mellanox driver.

### DIFF
--- a/.charts.yml
+++ b/.charts.yml
@@ -187,6 +187,10 @@ charts:
     version: 1.2.0
     repository: *vexxhost_openstack_helm_repository
     dependencies: *vexxhost_openstack_helm_dependencies
+    patches:
+      gerrit:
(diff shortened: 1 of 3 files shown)


ricolin:
You need to make this PR to https://github.com/vexxhost/openstack-helm and bump version here after it got release there

ricolin:
> You need to make this PR to https://github.com/vexxhost/openstack-helm and bump version here after it got release there

so a version bump in vexxhost/openstack-helm (which can be manually PR specific for this or there is a CI PR), and then bump version here

PR #3841, opened 2026-04-17
feat(deploy): parallel deployment orchestrator with intra-component parallelism and early CI deps

ricolin:
## Summary

Combined PR stacking all parallel deployment orchestrator and CI perf work on top of `main`:

| PR | Description |
|----|-------------|
| #3818 | Parallel deployment orchestrator with DAG-based component scheduling |
| #3834 | Intra-component parallelism via pre-roles (magnum, octavia) |
| #3835 | Pre-install CI dependencies early to reduce setup overhead |
| #3844 | Break three config-only DAG edges for parallelism |
| #3872 | Skip Molecule `idempotence` step in Zuul check pipeline |
| #3873 | Skip molecule jobs for doc-only and meta-only changes via `irrelevant-files` |

### What's included

1. **Go-based parallel deployer** (`atmosphere deploy`) — 42 components across 11 parallel waves via DAG-based dependency graph (#3818)
2. **Intra-component parallelism** — `PreRoleName` runs pre-roles (image uploads, TLS certs) concurrently with main Helm installs (#3834)
3. **DAG optimization** — removes the magnum→octavia edge plus three additional config-only edges (neutron→nova, magnum→barbican/heat, rook-ceph-cluster→barbican) for extra critical-path parallelism (#3834, #3844)
4. **Early CI dependency installation** — system packages, Helm, Go, Python venv pre-installed in Zuul pre-run (#3835)
5. **Skip idempotence in check pipeline** — the AIO Molecule scenario no longer runs the `idempotence` step in check; gate and periodic runs keep it (#3872)
6. **Path-conditional molecule jobs** — documentation, release-note, and repo-meta-only PRs no longer trigger the full molecule job set via Zuul `irrelevant-files` (#3873)

Note: #3835 already includes the container image pre-pull work from #3843.

Supersedes #3818, #3834, #3835, #3844, #3872, #3873.

PR #3834, opened 2026-04-15
feat(deploy): add intra-component parallelism with pre-roles and optimize DAG dependencies

ricolin:
## Summary

Builds on top of the parallel deploy orchestrator (#3818) to add **intra-component parallelism** via pre-roles and **optimize the DAG dependency graph**, further reducing critical-path deployment time.

### Intra-component parallelism with pre-roles

Introduces a `PreRoleName` field on `Component`, enabling independent tasks (e.g. image uploads, cert creation) to run **in parallel** with the main Helm deploy within a single component slot.

The deployer launches the pre-role and main role as concurrent `ansible-playbook` processes via `errgroup`. A `_pre_role_active: true` variable signals the main role to skip tasks already handled by the pre-role.

Applied to two critical-path components:

| Component | Pre-role tasks | Estimated savings |
|-----------|---------------|-------------------|
| **magnum** | Glance image uploads (~150s) | ~120–150s (overlaps with Helm install) |
| **octavia** | cert-manager Certificates + admin quota setup | ~15s |

### DAG dependency optimization

Removes `octavia` from `magnum`'s dependency list. Magnum's Helm install only needs the octavia endpoint URL (a deterministic string from `openstack_helm_endpoints`), not a running octavia service. This lets magnum start after `barbican` + `heat` complete instead of waiting for `octavia`, saving ~4.5 min on the critical path.

### Pre-role fixes

- Add `openstack_helm_endpoints` role dependency to `octavia_pre` and `magnum_pre` (`meta/main.yml`)
- Add `README.md` to pre-role directories (required by CI)
- Add octavia TLS defaults (`octavia_tls_*`) to `octavia_pre/defaults/main.yml`
- Add magnum image vars (`_magnum_images`) and format defaults to `magnum_pre`

## Changed files

| File | Change |
|------|--------|
| `internal/deploy/component.go` | Add `PreRoleName` field, remove magnum→octavia dep |
| `internal/deploy/deployer.go` | Parallel pre-role + main role execution via errgroup |
| `internal/deploy/deployer_test.go` | Tests for pre-role playbook rendering |
| `roles/magnum_pre/` | **New** — pre-role for image uploads |
| `roles/magnum/tasks/main.yml` | Skip image uploads when `_pre_role_active` |
| `roles/octavia_pre/` | **New** — pre-role for certs + quotas |
| `roles/octavia/tasks/main.yml` | Skip cert/quota tasks when `_pre_role_active` |

PR #3818, opened 2026-04-13
feat(deploy): add parallel deployment orchestrator

mnaser:
## Summary

Adds a Go binary (`cmd/atmosphere`) that deploys Atmosphere components in parallel waves using a DAG-based dependency graph, targeting a reduction from ~60 minutes to ~22 minutes for full deployments.

## Architecture

| Package | Purpose |
|---------|---------|
| `pkg/dag/` | Generic `Graph[T]` library — Kahn's topo sort, subgraph extraction, parallel wave execution via errgroup |
| `internal/deploy/` | 42-component registry, `Deployer` interface + `AnsibleDeployer`, 3-mode orchestrator |
| `cmd/atmosphere/` | cobra CLI: `atmosphere deploy --inventory <inv> [--tags <t>]` |

## Three Operating Modes

1. **No tags** (full deploy): All 42 components run across 11 parallel waves
2. **Single tag** (`--tags keystone`): Pass-through to `ansible-playbook site.yml --tags keystone` — identical to today
3. **Multiple tags** (`--tags nova,keystone`): DAG-aware subgraph with parallel waves

## Wave Schedule (Full Deploy)

```
Wave 0:  ceph, kubernetes                          (2 parallel)
Wave 1:  csi                                       (1)
Wave 2:  cert-manager, memcached, rook-ceph, ...   (19 parallel)
Wave 3:  cluster-issuer, rabbitmq-op, pxc-op, ...  (6 parallel)
Wave 4:  ingress-nginx, pxc, ceph-provisioners     (3 parallel)
Wave 5:  keycloak                                  (1)
Wave 6:  keystone                                  (1)
Wave 7:  barbican, glance, cinder, placement, ...  (10 parallel)
Wave 8:  nova, staffeln                            (2 parallel)
Wave 9:  octavia                                   (1)
Wave 10: openstack-exporter                        (1)
```

## How It Works

- For `PlaybookType` components (ceph, kubernetes, csi): runs the full playbook file directly
- For `RoleType` components (everything else): generates a minimal single-role playbook at runtime and pipes it via `/dev/stdin` to `ansible-playbook`, avoiding multi-play parsing overhead
- Output from concurrent processes is streamed with `[component-name]` prefixes
- A prerequisite step runs `openstacksdk` once before parallel waves to avoid concurrent `pip install` / `clouds.yaml` writes

## Backwards Compatibility

- **Zero changes** to existing Ansible roles, variables, or playbooks
[...]

PR #3844, opened 2026-04-20
feat(deploy): break three config-only DAG edges for parallelism

ricolin:
## Summary

Audit of install-time behaviour vs. declared dependencies in the parallel deployment orchestrator found three edges where the declared dependency is a **configuration-only reference** (endpoint URL stored in a Helm values template) rather than a real install-time API call. Removing them lets the orchestrator schedule components earlier.

### Edges removed

| Component | Old deps | New deps | Why the removed edge is safe |
|---|---|---|---|
| `neutron` | keystone, **nova**, ovn, coredns | keystone, ovn, coredns | `roles/neutron/tasks/main.yml` only makes Neutron API calls (`openstack.cloud.network`, subnets) and does not touch Nova at install time. |
| `magnum` | **barbican**, **heat** | keystone | `roles/magnum/vars/main.yml` references `barbican_client` / `heat_client` in `magnum.conf`. These strings are only dereferenced when a user later creates a cluster. |
| `rook-ceph-cluster` | rook-ceph, ceph, **barbican** | rook-ceph, ceph, **keystone** | The role creates an identity user and catalog entry for the RGW. The previous Barbican dependency was mis-declared — the real dependency is Keystone. |

### Readiness check added

Along with the Keystone dependency, `roles/rook_ceph_cluster/tasks/main.yml` now waits for the `keystone-api` Deployment to be `Available` before the subsequent `openstack.cloud.*` calls, preventing a race with Keystone's rollout. This mirrors the cert-manager secret wait added for Octavia in #3834.

The other two removals (`neutron -> nova`, `magnum -> barbican/heat`) don't need new gates: audit of the affected `tasks/` files confirms they make no install-time API calls against the removed dependencies.

## Measured impact

On `atmosphere-molecule-aio-ovn` the current critical path is:

```
... -> keystone -> Wave6 -> nova -> neutron -> (octavia || manila) = ~12 min tail
```

After these changes the expected critical path is:

```
... -> keystone -> Wave6 -> (nova || neutron) -> (octavia || manila) = ~8.5 min tail
```

**Expected saving: ~3 min 20 s on the critical path** (roughly 9% of the measured 34m 40s non-prepull deploy time in #3841).

## Validation

- `go test ./pkg/dag/ ./internal/deploy/` passes.
- `go build ./cmd/atmosphere` succeeds.
- Release note passes `vale` with zero errors / warnings / suggestions.
- DCO signed.

## Stacking

This PR is based on #3818 (`feat/parallel-deploy-orchestrator`) so it is self-contained once #3818 lands on main. #3841 will be rebuilt as #3834 + #3835 + this PR on top of `main`.

Signed-off-by: Rico Lin <rlin@vexxhost.com>
Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/internal/deploy/component.go
+++ b/internal/deploy/component.go
@@ -263,7 +263,7 @@ var Components = []Component{
 		Type:        RoleType,
 		RoleName:    "rook_ceph_cluster",
 		Hosts:       "controllers[0]",
-		DependsOn:   []string{"rook-ceph", "ceph", "barbican"},
+		DependsOn:   []string{"rook-ceph", "ceph", "keystone"},
 		Environment: cephEnvironment,
 	},
 	{
@@ -380,7 +380,7 @@ var Components = []Component{
 		Type:      RoleType,
 		RoleName:  "magnum",
 		Hosts:     "controllers[0]",
-		DependsOn: []string{"octavia", "barbican", "heat"},
+		DependsOn: []string{"keystone", "glance"},
 	},
 	{
 		Name:      "manila",
--- a/releasenotes/notes/dag-edge-removal-phase1-439694be3dac4d84.yaml
+++ b/releasenotes/notes/dag-edge-removal-phase1-439694be3dac4d84.yaml
@@ -1,0 +1,17 @@
+---
+features:
+  - |
+    The parallel deployment orchestrator now runs Magnum in parallel with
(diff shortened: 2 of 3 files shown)


ricolin:
Squash-merged into #3818 (feat/parallel-deploy-orchestrator). Closing.

PR #3872, opened 2026-04-23
ci(molecule): skip idempotence step in check pipeline

ricolin:
## Summary

Idempotence re-runs the full AIO converge (~20-30 minutes) to assert that every role converges with zero changed tasks on a second run. Running it on **every** check build is wasteful: it roughly doubles the wall time of the AIO job and the same guarantee is re-verified in the gate pipeline before merge.

## Change

- Make the AIO scenario's `test_sequence` explicit in `molecule/aio/molecule.yml` (otherwise Molecule uses its built-in default).
- In `test-playbooks/molecule/pre.yml`, drop the `idempotence` line from that sequence when `zuul.pipeline == 'check'`.
- Gate and periodic runs are untouched — they still execute the full `molecule test` including `idempotence`.

## Expected saving

On a clean check run of `atmosphere-molecule-aio-*`, the idempotence converge accounts for **~20-30 minutes of a ~60 minute job**. Most PRs push multiple times before merging, so this compounds across the life of a change.

## Stacking

Branched from `feat/parallel-deploy-orchestrator` (#3818) per request. Once #3818 lands, this PR's diff against main will just be the 3 files above.

The same change will also be rolled into #3841.

## Caveats

- If a role regresses idempotence on a `main` branch that has no direct gate run (e.g. scheduled jobs), the periodic job will still catch it.
- Check-pipeline reviewers no longer get a per-push signal that a role is idempotent. That is a deliberate trade-off: gate blocks merge anyway.

Signed-off-by: Rico Lin <ricolin@ricolky.com>
Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/molecule/aio/molecule.yml
+++ b/molecule/aio/molecule.yml
@@ -12,3 +12,18 @@ ansible:
     args:
       ansible_playbook:
         - --inventory=${MOLECULE_PROJECT_DIRECTORY}/inventory.yaml
+
+scenario:
+  test_sequence:
+    - dependency
+    - cleanup
+    - destroy
+    - syntax
+    - create
+    - prepare
+    - converge
+    - idempotence
+    - side_effect
+    - verify
+    - cleanup
+    - destroy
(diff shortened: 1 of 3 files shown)


ricolin:
recheck

PR #3875, opened 2026-04-24
feat(deploy): schedule DAG with per-node readiness instead of wave barriers

ricolin:
## Summary

Replace the wave-barrier scheduler in `pkg/dag/dag.go` with an event-driven (per-node readiness) one. Each component starts as soon as its own `DependsOn` nodes are complete, instead of waiting for the entire Kahn wave to drain.

## Motivation

Critical-path analysis of #3841 showed a **340s idle gap** on wave 2 of an AIO check build:

- `cert-manager` ends at 253s.
- `percona-xtradb-cluster-operator` only declares `DependsOn: ["cert-manager"]`, so it *should* start immediately at 253s.
- Instead it starts at 593s, because `cert-manager` and `ceph` both have `kubernetes` as their only dep and therefore share a wave — and `ceph` is the long pole of that wave.

That 340-second stall isn't in any `DependsOn` declaration. It's purely an artifact of the `Waves()`-driven loop in `Run`:

```go
for _, wave := range waves {
    ...
    if err := eg.Wait(); err != nil { return err }  // ← barrier
}
```

## Change

`Run` now:

- Spawns one goroutine per node up front.
- Each goroutine `select`s on the `done` channels of its direct deps (or `ctx.Done()`).
- Acquires the global concurrency semaphore (`concurrency` parameter, if > 0).
- Calls `fn`, and on success closes its own `done` channel to unblock dependents.
- On error, leaves its `done` channel **open** so dependents exit via the errgroup context cancellation.

Cycle detection is preserved by calling `Waves()` purely as a validation step before scheduling.

## Correctness

- Partial order from `DependsOn` is fully preserved (each node waits on every declared dep).
- Failures propagate: failed node's dependents are cancelled, not hung.
- `concurrency` becomes a **global** in-flight cap rather than per-wave. That matches what we actually want to bound (how many playbooks run at once on the executor).

## Tests

- `TestRunShortNodeNotBlockedByUnrelatedLongNode` — regression test that locks in the new behavior: a short node starts before an unrelated long node in the same Kahn wave finishes.
- `TestRunStopsDependentsOnError` — dependents of a failed node never run.
- `TestRunConcurrencyCap` — global cap bounds in-flight across the whole graph.
- All existing tests (ordering, parallelism, subgraph, reverse) still pass.
- `go test ./pkg/dag/ ./internal/deploy/ -race` is green.

## Expected CP impact

On the measured #3841 build, the single direct saving on the critical path is bounded by where the pxc chain ultimately re-joins ceph via `csi`. But the **scheduler change is infrastructure, not a point fix**: any future build where a short node is wave-pinned behind a long unrelated node benefits immediately, without touching the DAG.

Next PRs in this series (nova/neutron split, `glance_pre`, etc.) rely on this scheduler to realize their full savings.

## Stacking

Branched from `feat/parallel-deploy-orchestrator` (#3818) because it builds on that orchestrator. Targets `main` so that once #3818 merges, this PR's diff collapses to just the three files above.

## Files

- `pkg/dag/dag.go` — event-driven `Run`.
- `pkg/dag/dag_test.go` — three new tests.
- `releasenotes/notes/dag-event-driven-scheduler-*.yaml` — release note.


ricolin:
Folding this commit back into #3818 since the scheduler change belongs in the same perf stack. The event-driven scheduler commit (8b3471cf) is now on feat/parallel-deploy-orchestrator.

Perf validation (aio-ovn, #3818 latest vs this scheduler change):
- Deploy tail: 2404s → 2096s (**-308s / -12.8%**)
- Job wall: 3633s → 3308s (**-8.9%**)
- No component regressions
See `/home/rico/notes/logs/atmosphere/260424.perf-3875-vs-3818.md`.

PR #3876, opened 2026-04-24
feat(deploy): prioritise ready DAG nodes by critical path under concurrency cap

ricolin:
Stacked on top of #3818.

## Problem

The parallel deploy orchestrator introduced in #3818 starts a goroutine per DAG node and serialises admission through a plain channel semaphore when a global concurrency cap is set. Channel semaphores are FIFO-ish but not priority-aware, so when multiple ready nodes compete for the last free slot, admission order is effectively arbitrary. In practice that means a short-chain node can be admitted ahead of a long-chain node that still has a deep tail of dependents behind it — the worst possible choice for makespan.

## Change

Replace the channel semaphore in `pkg/dag.Graph.Run` with a priority-aware scheduler. Priority = length of each node's longest downstream chain (a unit-weighted critical path), precomputed once per `Run` via the new `CriticalPath()` method. When capacity frees up, the waiter with the highest priority is admitted next; ties fall back to arrival order so ordering is deterministic. When `concurrency <= 0` the scheduler is skipped entirely and behaviour is unchanged.

This is the classic list-scheduling / HEFT heuristic. It costs O(V+E) at startup and a bounded heap during the run, and it preserves the partial order established by `DependsOn`.

## Tests

- `TestCriticalPath` verifies the depth computation on a branching graph.
- `TestRunPriorityUnderConcurrencyCap` drives the scheduler directly: primes the one-slot capacity, queues waiters in a non-priority order, then releases; asserts admission happens in priority-descending order.
- `TestRunPriorityDoesNotBlockIndependentChains` guards against the scheduler refusing to admit anything when there's capacity.
- Existing DAG tests (ordering, parallelism, wave-gap closure, error propagation, concurrency cap) still pass under `-race -count=10`.

## Target

Base is \`main\` on purpose — this change does not depend on anything specific to #3818 once merged, but is stacked on #3818 so the diff shows the intended final state. Once #3818 merges, this PR will rebase cleanly.

## Notes

- Scheduler uses `container/heap`; no new deps.
- Priorities are static (unit weights). A future PR could feed in measured p50 durations for better ordering, gated on the per-component timing metrics tier-2 item.


ricolin:
Rolled into #3818 via commits 8db74c62 (critical-path scheduling) and 743ca1a1 (release note refresh). Closing this draft.

PR #3888, opened 2026-04-28
feat(monitoring): add Geneve and bond MTU drift alerts

larainema:
Adds a focused Geneve transmit-error alert and the node-exporter metrics needed to identify Geneve tunnel interfaces without relying on interface names.

## Changes

- Adds `GeneveTransmitErrors`, a Prometheus alert for sustained transmit errors on devices matched through `node_ethtool_info{driver="geneve"}`.
- Enables the `node-exporter` `ethtool` collector.
- Updates the node-exporter network collector filters so Geneve tunnel interfaces, including `cilium_geneve`, can be exported by the `netdev` collector.
- Adds a runbook entry for `GeneveTransmitErrors` in the monitoring documentation.
- Adds focused reno notes for the `ethtool` collector change and the Geneve transmit-error alert.

## Motivation

A compute host ran with an underlay MTU mismatch for an extended period. The downstream symptoms showed up as tenant overlay traffic problems, including TLS handshake stalls and noisy live-migration cutovers. Sustained Geneve transmit errors are the strongest metric signal for that class of failure because the Linux Geneve transmit path increments errors when encapsulation or underlay transmit fails.

This intentionally does not add a separate bond MTU inconsistency alert because existing MTU consistency coverage already handles that class of drift, and it does not alert on transmit drops because the Geneve drop counters are weaker and less actionable in this path.

## Verification

- `CGO_ENABLED=0 go test -v -run "TestPrometheusRules/GeneveTransmitErrors" ./roles/kube_prometheus_stack/` passes.
- Full `TestPrometheusRules` was run; `GeneveTransmitErrors` passed and the suite only failed on existing unrelated `MysqlClusterDown` fixtures.
- Vale was run on the touched monitoring docs and release notes; only existing `AlertManager` terminology findings remain in `doc/source/admin/monitoring.rst`.
- In a live deployment, `node_ethtool_info{driver="geneve"}` returns both `cilium_geneve` and `genev_sys_6081` devices.
- In a live deployment, the joined alert query executes successfully and currently returns zero transmit-error rate for `genev_sys_6081`; the thresholded alert expression executes and returns no firing series.


mnaser (approved):
Reviewed the alert rules, runbook entries, and promtool tests for the Geneve transmit error/drop and bond MTU drift alerts. The new alert cases passed locally via `nix-shell -p go prometheus.cli --run 'CGO_ENABLED=0 go test -v -run TestPrometheusRules ./roles/kube_prometheus_stack/'`.

The full suite still has two existing `MysqlClusterDown` fixture failures, but the new `GeneveTransmitErrors`, `GeneveTransmitDrops`, and `NodeBondMTUInconsistent` tests all passed.

PR #3903, opened 2026-05-05
feat(monitoring): add HostNICSpeedUnknown alert

donnyao:
## What

Adds a new Prometheus alert `HostNICSpeedUnknown` that fires when `node_network_speed_bytes < 0` for more than 5 minutes. `node_exporter` emits a negative value when the kernel cannot read the interface speed via ethtool (`Speed: Unknown!`).

## Why

This catches a class of silent data-plane failures on bond slaves where:

- The link stays administratively up (carrier on, MII up)
- But the transceiver / cable / switch port is degraded so no link speed can be negotiated
- And `node_exporter` therefore reports a negative speed

`NodeBondingDegraded` does not catch this because the bond driver still considers the slave active (MII up). In a recent customer incident, a control-node bond ran on a single healthy slave for 9+ days without paging, because the broken slave's MII never went down.

`HostNICSpeedUnknown` would have paged immediately on the same condition.

## Changes

- `roles/kube_prometheus_stack/files/jsonnet/mixins.libsonnet` — add `HostNICSpeedUnknown` rule to the existing `node-exporter-extras` group (severity: warning → P3)
- `roles/kube_prometheus_stack/files/jsonnet/tests.yml` — positive test (negative speed → alert fires) and negative test (normal speed → no alert)
- `doc/source/admin/monitoring.rst` — runbook entry with diagnostic and remediation steps
- `releasenotes/notes/host-nic-speed-unknown-alert-*.yaml` — feature release note

## Testing

```
$ cd roles/kube_prometheus_stack && go test -run TestPrometheusRules -v
...
--- PASS: TestPrometheusRules/HostNICSpeedUnknown (0.39s)
--- PASS: TestPrometheusRules/HostNICSpeedUnknown#01 (0.38s)
```

(Pre-existing `MysqlClusterDown` failures on `main` are unrelated to this change.)

### DIFF
--- a/.github/styles/config/vocabularies/Base/accept.txt
+++ b/.github/styles/config/vocabularies/Base/accept.txt
@@ -88,6 +88,7 @@ backport
 boolean
 cgroup
 etcd
+ethtool
 image_manifest
 kek
 libvirt
--- a/doc/source/admin/monitoring.rst
+++ b/doc/source/admin/monitoring.rst
@@ -1044,6 +1044,62 @@ the risk of a full outage if another node fails.
 
      kubectl -n openstack get pxc-backup
 
+``HostNICSpeedUnknown``
+=======================
+
+This alert fires when ``node_network_speed_bytes`` reports a negative value for
+a network interface for at least 5 minutes. ``node_exporter`` returns a
+negative value when the kernel cannot read the interface speed via ethtool
+(``Speed: Unknown!``), which typically indicates a silent data-plane failure:
+the OS keeps the link administratively up but the transceiver, cable, or
+switch port is degraded so no link speed can be negotiated. Bonded interfaces
+will not detect this state because MII status remains up, so the underlying
+slave can stay in the bond as a "zombie" carrying no traffic.
+
+**Likely root causes:**
+
+- Failed or marginal transceiver, DAC/AOC cable, or fiber patch
+- Switch port disabled, errdisabled, or in an unexpected operational state
+- NIC firmware bug after a host or switch reboot
+- NIC hardware fault (queue or PHY failure) that does not raise a carrier event
+
+**Diagnostic and remediation steps:**
+
+1. Identify the affected host and interface from the alert labels (``instance``
+   and ``device``).
+
+2. Confirm the speed is unknown on the host:
+
+   .. code-block:: console
+
+      ethtool <device> | grep -E "Speed|Link detected"
+
+3. Check whether the interface is part of a bond and how traffic is split
+   across slaves:
+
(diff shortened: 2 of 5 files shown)


mnaser (changes requested):
Thanks for adding coverage for this failure mode. I validated the rule shape against a live Atmosphere deployment and checked the Linux/node_exporter behavior. I think the failure mode is real, but the current expression is too broad and would page on unrelated cases.

Live validation:

- The PR expression, `node_network_speed_bytes{job="node-exporter"} < 0`, currently returns sustained matches in a live deployment. With `for: 5m`, this would fire immediately if deployed there.
- The matches are not all the intended failure mode:
  - virtual tunnel interfaces can report `-125000` because their speed is unknown
  - disconnected or unused physical Ethernet ports can also report `-125000`, but with `carrier=0` and `operstate=down`
- In those cases, `/sys/class/net/<dev>/speed = -1`, but that alone does not mean a bond slave is silently failed while still considered usable.

Historical incident context:

- A previous customer incident did have the condition this alert is trying to catch: a physical Ethernet bond slave reported `Speed: Unknown!` / `node_network_speed_bytes = -125000` while the link was still detected and the interface remained relevant to the bond.
- So the signal is possible and useful, but only when combined with enough state to exclude expected virtual-device behavior and simple link-down ports.

Kernel/node_exporter feasibility:

- Linux exposes speed and carrier independently. `/sys/class/net/<iface>/speed` can be `SPEED_UNKNOWN` (`-1`) while `/sys/class/net/<iface>/carrier` is still `1`.
- node_exporter’s netclass collector converts Mbps to bytes/sec, so `-1 Mbps` becomes `-125000` bytes/sec.

Could you update the alert to target the specific useful condition and avoid standing false positives? For example:

```promql
node_network_speed_bytes{job="node-exporter"} < 0
  and on(instance, device) node_network_protocol_type{job="node-exporter"} == 1
  and on(instance, device) node_network_carrier{job="node-exporter"} == 1
```

This should exclude virtual devices and physical interfaces that are simply disconnected/down. Please also update the tests to include those negative cases:

- virtual interface with negative speed should not fire
- Ethernet interface with negative speed and `carrier=0` should not fire
- Ethernet interface with negative speed and `carrier=1` should fire

Separately, the alert annotations and docs ordering need follow-up to satisfy `.github/instructions/alerts.instructions.md`:

- summary should use `<Component>: <Impact statement>`
- description should include `{{ $value }}`, threshold, and normal behavior
- `HostNICSpeedUnknown` should be alphabetically placed before the `Ipmi...` entries in the alert reference

PR #3906, opened 2026-05-07
feat: switch token validation to use oauth2 and jwks endpoint per realm

fitbeard:
Improve security and flexibility for more possible integrations.

Depends on: https://github.com/vexxhost/docker-keystone/pull/753


fitbeard:
Initial CI tests will fail with current images on Keystone bootstrap.

PR #3915, opened 2026-05-11
fix(deploy): four critical bugs in parallel orchestrator (follow-up to #3818)

ricolin:
## Summary

Follow-up to #3818 — fixes four critical bugs identified during code review of the parallel deployment orchestrator.

> **Base note:** branch sits on top of `feat/parallel-deploy-orchestrator` (#3818). PR target is `main` so the diff visible here will only be meaningful **after** #3818 lands. Until then, review by looking at the latest commit on this branch only.

## Bugs fixed

| ID | Bug | Fix |
|----|-----|-----|
| **C1** | `atmosphere deploy --tags neutron,nova` (or any multi-tag where a selected component has `PreRoleDependsOn` outside the subgraph) **hangs forever** because the pre-role goroutine waits on completion channels that never close. | In `deployMultipleTags`, allocate `completionTracker` channels only for components in the subgraph. `tracker.Wait` already skips unknown names, which now matches the `--tags` semantic that everything not selected is assumed already deployed. |
| **C2** | Ctrl-C **orphans `ansible-playbook`/`helm`/`ssh` subprocesses** because `cmd/atmosphere` never wraps the root context with `signal.NotifyContext`. Combined with `Setpgid: true`, children survive the parent. | Wrap `cmd.Context()` with `signal.NotifyContext(os.Interrupt, syscall.SIGTERM)` in `newDeployCmd.RunE`. |
| **C3** | `defer tracker.MarkDone(id)` runs **even on failure**, briefly unblocking downstream pre-role gates before errgroup cancellation propagates. | Move `MarkDone` to the success path (after `Deployer.Deploy` returns nil). |
| **C4** | `BuildGraph` **silently ignores** the error from `g.AddNode`, masking duplicate-name regressions. | Check the error and propagate it. |

## Why the existing tests didn't catch C1

The mock deployers in `orchestrator_test.go` drop the `preGate` callback. The new `gateAwareDeployer` in `orchestrator_subgraph_pregate_test.go` invokes it, mirroring `AnsibleDeployer.Deploy`. With the old code, the new test deadlocks until the 5 s ctx timeout fires; with the fix, it returns immediately.

Verified:

```
$ git stash && go test -run TestDeployMultipleTags_PreRoleDependencyOutsideSubgraph -v ./internal/deploy
--- FAIL: TestDeployMultipleTags_PreRoleDependencyOutsideSubgraph (5.00s)
    Deploy returned error (likely a hang regression): component neutron failed: context deadline exceeded

$ git stash pop && go test -run TestDeployMultipleTags_PreRoleDependencyOutsideSubgraph -v ./internal/deploy
--- PASS: TestDeployMultipleTags_PreRoleDependencyOutsideSubgraph (0.00s)
```

## Tests added

- `TestDeployMultipleTags_PreRoleDependencyOutsideSubgraph` — neutron+nova (C1 regression guard)
- `TestDeployMultipleTags_KeystoneAlone` — keystone variant of C1
- `TestDeployFailure_DoesNotMarkDone` — C3 regression guard

## Files

- `cmd/atmosphere/deploy.go` — C2
- `internal/deploy/component.go` — C4
- `internal/deploy/orchestrator.go` — C1, C3
- `internal/deploy/orchestrator_subgraph_pregate_test.go` (new) — regression tests


ricolin:
Superseded by #3818 — all commits (C1–C4 + Keystone variant test fix + Wave 2 apiserver concurrency cap) have been squashed onto feat/parallel-deploy-orchestrator. See https://github.com/vexxhost/atmosphere/pull/3818 for the consolidated change and the AIO validation report.

PR #3924, opened 2026-05-13
fix(deploy): mask host libvirt sockets before libvirt

ricolin:
## Summary

- add a `host_libvirt_sockets` role that stops and masks host libvirt/virtlogd/virtlockd socket-activated units
- wire the role into the parallel deploy DAG as a dependency of `libvirt`
- run the same host prep in the tagged `openstack.yml` path so `--tags libvirt` preserves the ordering

## Notes

This PR is based on #3818 and targets `main`. It addresses the AIO race where host `libvirtd.socket` can bind `/run/libvirt` before the Kubernetes libvirt DaemonSet starts.

## Testing

- `go test ./internal/deploy ./pkg/dag`
- `ANSIBLE_ROLES_PATH=roles ansible-playbook --syntax-check -i localhost, /dev/stdin` with a minimal play invoking `host_libvirt_sockets`

Full `playbooks/openstack.yml` syntax-check could not complete in this local checkout because Ansible cannot resolve an existing dependency role, `atmosphere.common.secretgen_controller`, from `roles/octavia/meta/main.yml`.


mnaser:
@ricolin we never want to have libvirt installed on the host, how did this env that you're testing end up with libvirt on the host?

We want to eliminate Ansible as much as possible

ricolin:
okay should mark this as draft as just experiment. let's drop this

PR #3925, opened 2026-05-14
feat(neutron): add neutron ironic agent support

ricolin:
## Summary

- add Neutron Helm values for the baremetal ML2 backend and `[ironic]` credentials
- append those values only when the final Helm values enable `manifests.deployment_ironic_agent`
- source Ironic region, username, and password from `openstack_helm_endpoints` instead of hardcoding Keystone auth details

### short version for ironic-neutron-agent  and baremetal driver intro

  ironic-neutron-agent tells Neutron which physical networks each Ironic bare metal node can reach. That information lets Neutron
  place bare metal ports on the right network segment, VLAN, provider network, or routed provider network.

  This PR enables two required pieces together:

  - ironic-neutron-agent syncs Ironic node and port network information into Neutron.
  - baremetal in network.backend enables the ML2 binding path for Neutron ports using vnic_type=baremetal.

  Without the baremetal backend, the agent can report the topology, but Neutron may still fail to bind bare metal ports because ML2
  does not have the driver for that binding path.

  This stays opt-in through manifests.deployment_ironic_agent, so the default Neutron deployment remains unchanged.

### Long version for ironic-neutron-agent  and baremetal driver intro

ironic-neutron-agent is a Neutron-side service used with OpenStack Ironic bare-metal nodes.

Its main job is to tell Neutron which physical networks each bare-metal node can reach. This allows Neutron to correctly bind bare-metal ports to the right network segment, VLAN, provider network, or routed network.

In practice, it helps with:

* Mapping Ironic nodes to physical networks
* Supporting bare-metal Neutron port binding
* Enabling tenant, provisioning, cleaning, and rescue network transitions
* Helping ML2/switch drivers configure the correct switch ports or VLANs
* Supporting routed provider networks
* Improving OVN bare-metal networking features in newer deployments

It does not power on servers, install operating systems, run inside the bare-metal node, or directly manage hardware. Those tasks are handled by Ironic, ironic-python-agent, and hardware drivers.

In short: ironic-neutron-agent lets Neutron understand and manage networking for Ironic bare-metal nodes.

Adding some context on why this PR enables the baremetal ML2 backend together with the Neutron Ironic agent:

The Neutron Ironic agent is only the service that reports Ironic bare metal port and physical network information back to Neutron. That gives Neutron the inventory and reachability data it needs, but it does not by itself teach ML2 how to bind a Neutron port whose binding:vnic_type is baremetal.

For the bind flow to work end to end, Neutron also needs the baremetal mechanism/backend enabled in network.backend. That backend is what lets ML2 recognize bare metal ports, use the Ironic-side physical network information, and produce a valid binding for provider or routed provider networks. Without it, we can deploy ironic-neutron-agent, but Neutron can still reject or fail to bind bare metal ports because the ML2 backend list does not include the driver responsible for that binding path.

So these two pieces cover different sides of the same feature:

ironic-neutron-agent syncs Ironic node/port network information into Neutron.
baremetal in network.backend enables the ML2 binding path for Neutron ports using vnic_type=baremetal.
That is why this change appends baremetal alongside the existing Atmosphere network backend only when operators opt in by enabling manifests.deployment_ironic_agent. It keeps the default Neutron deployment unchanged, while making the opt-in Ironic networking path complete.

## Notes

This intentionally ports only the `roles/neutron/vars/main.yml` and `roles/neutron/tasks/main.yml` portions from `ricolin/atmosphere#11`. It does not set `manifests.deployment_ironic_agent`; operators still opt in through `neutron_helm_values`.

## Validation

- `python3` YAML parse for `roles/neutron/vars/main.yml` and `roles/neutron/tasks/main.yml`
- local `ansible-playbook` smoke test for false/true `deployment_ironic_agent` conditional behavior
- `git diff --check`

### DIFF
--- a/roles/neutron/tasks/main.yml
+++ b/roles/neutron/tasks/main.yml
@@ -57,7 +57,8 @@
     _neutron_helm_values: "{{ _neutron_helm_values | combine(__neutron_generic_switch_keys_helm_values, recursive=True) }}"
 
 - name: Append Helm values (ironic agent)
-  when: (_neutron_helm_values | combine(neutron_helm_values, recursive=True)).manifests.deployment_ironic_agent | bool
+  when: >-
+    (((_neutron_helm_values | combine(neutron_helm_values, recursive=True)).manifests | default({})).deployment_ironic_agent | default(false)) | bool
   ansible.builtin.set_fact:
     _neutron_helm_values: "{{ _neutron_helm_values | combine(__neutron_ironic_helm_values, recursive=True) }}"
 
--- a/roles/neutron/tasks/main.yml
+++ b/roles/neutron/tasks/main.yml
@@ -56,6 +56,11 @@
   ansible.builtin.set_fact:
     _neutron_helm_values: "{{ _neutron_helm_values | combine(__neutron_generic_switch_keys_helm_values, recursive=True) }}"
 
+- name: Append Helm values (ironic agent)
+  when: (_neutron_helm_values | combine(neutron_helm_values, recursive=True)).manifests.deployment_ironic_agent | bool
+  ansible.builtin.set_fact:
+    _neutron_helm_values: "{{ _neutron_helm_values | combine(__neutron_ironic_helm_values, recursive=True) }}"
+
 - name: Deploy Helm chart
   run_once: true
   kubernetes.core.helm:
--- a/roles/neutron/vars/main.yml
+++ b/roles/neutron/vars/main.yml
@@ -55,6 +55,9 @@ __neutron_helm_values:
       nova:
         live_migration_events: true
       ironic:
+        region_name: "{{ openstack_helm_endpoints_ironic_region_name }}"
+        username: "ironic-{{ openstack_helm_endpoints_ironic_region_name }}"
+        password: "{{ openstack_helm_endpoints_ironic_keystone_password }}"
         valid_interfaces: internal
       placement:
         region_name: "{{ openstack_helm_endpoints_neutron_region_name }}"
@@ -157,6 +160,12 @@ __neutron_ovn_helm_values:
     daemonset_ovn_metadata_agent: true
     daemonset_ovn_vpn_agent: true
     daemonset_ovs_agent: false
+
+__neutron_ironic_helm_values:
+  network:
+    backend:
+      - "{{ atmosphere_network_backend }}"
+      - baremetal
 
 __neutron_policy_server_helm_values:
   conf:
--- a/doc/source/user/index.rst
+++ b/doc/source/user/index.rst
@@ -6,3 +6,4 @@ User Guide
    :maxdepth: 2
 
    auth
+   ironic
--- a/doc/source/user/ironic.rst
+++ b/doc/source/user/ironic.rst
@@ -1,0 +1,87 @@
+######
+Ironic
+######
+
+***********************************************************
+Make Neutron aware of bare metal physical network topology
+***********************************************************
+
+.. note::
+
+   This feature requires an environment with the Neutron Ironic agent.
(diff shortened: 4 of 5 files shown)


roles/neutron/tasks/main.yml
```diff
@@ -31,6 +31,14 @@
   ansible.builtin.set_fact:
     _neutron_helm_values: "{{ _neutron_helm_values | combine(__neutron_policy_server_helm_values, recursive=True) }}"
 
+- name: Append Helm values (ironic agent)
+  when: >-
+    ((_neutron_helm_values | combine(neutron_helm_values, recursive=True))
```
ricolin: just set fact for this "((_neutron_helm_values | combine(neutron_helm_values, recursive=True))" before task "Append Helm values (ironic agent)" so we can reuse this in that task and also in followup tasks.
I htink

roles/neutron/vars/main.yml
```diff
@@ -139,6 +137,19 @@ __neutron_ovn_helm_values:
...
 
+__neutron_ironic_helm_values:
+  network:
+    backend:
+      - "{{ atmosphere_network_backend }}"
+      - baremetal
+  conf:
+    neutron:
+      ironic:
+        region_name: "{{ openstack_helm_endpoints['identity']['auth']['ironic']['region_name'] }}"
```
ricolin: I'm expect something like openstack_helm_endpoints_neutron_region_name
so as username and password below

roles/neutron/tasks/main.yml
```diff
@@ -31,6 +31,17 @@
...
     _neutron_helm_values: "{{ _neutron_helm_values | combine(__neutron_policy_server_helm_values, recursive=True) }}"
 
+- name: Merge Helm values
+  ansible.builtin.set_fact:
+    _neutron_helm_values: "{{ _neutron_helm_values | combine(neutron_helm_values, recursive=True) }}"
+
+- name: Append Helm values (ironic agent)
+  when: >-
+    (_neutron_helm_values.get('manifests', {})
+    .get('deployment_ironic_agent', false)) | bool
```
ricolin: we always have manifests.deployment_ironic_agent so no need to use get and default here

roles/neutron/tasks/main.yml
```diff
@@ -31,6 +31,15 @@
   ansible.builtin.set_fact:
     _neutron_helm_values: "{{ _neutron_helm_values | combine(__neutron_policy_server_helm_values, recursive=True) }}"
 
+- name: Merge Helm values
```
ricolin: just remove this Merge Helm values as we need  neutron_helm_values to be the final override power

ricolin:
Adding some context on why this PR enables the `baremetal` ML2 backend together with the Neutron Ironic agent:

The Neutron Ironic agent is only the service that reports Ironic bare metal port and physical network information back to Neutron. That gives Neutron the inventory and reachability data it needs, but it does not by itself teach ML2 how to bind a Neutron port whose `binding:vnic_type` is `baremetal`.

For the bind flow to work end to end, Neutron also needs the `baremetal` mechanism/backend enabled in `network.backend`. That backend is what lets ML2 recognize bare metal ports, use the Ironic-side physical network information, and produce a valid binding for provider or routed provider networks. Without it, we can deploy `ironic-neutron-agent`, but Neutron can still reject or fail to bind bare metal ports because the ML2 backend list does not include the driver responsible for that binding path.

So these two pieces cover different sides of the same feature:

- `ironic-neutron-agent` syncs Ironic node/port network information into Neutron.
- `baremetal` in `network.backend` enables the ML2 binding path for Neutron ports using `vnic_type=baremetal`.

That is why this change appends `baremetal` alongside the existing Atmosphere network backend only when operators opt in by enabling `manifests.deployment_ironic_agent`. It keeps the default Neutron deployment unchanged, while making the opt-in Ironic networking path complete.

PR #3933, opened 2026-05-18
fix: update cert-manager webhook infoblox wapi chart repo

yaguangtang:
The luisico repo chart image is built with k8s.io/client-go@v0.21.3, which tries to watch flowcontrol.apiserver.k8s.io/v1beta1 resources (FlowSchema and PriorityLevelConfiguration). These v1beta1 APIs were removed in Kubernetes 1.26, therefore the webhook fails to run in current 1.28.x environment.
This patch update the chart to use a fork version of the luisico repo.

### DIFF
--- a/.charts.yml
+++ b/.charts.yml
@@ -38,9 +38,9 @@ charts:
     repository:
       url: https://snowdrop.github.io/godaddy-webhook
   - name: cert-manager-webhook-infoblox-wapi
-    version: 1.5.2
+    version: 2.0.0
     repository:
-      url: https://luisico.github.io/cert-manager-webhook-infoblox-wapi
+      url: https://sarg3nt.github.io/cert-manager-webhook-infoblox-wapi
   - name: cinder
     version: 0.3.23
     repository: *openstack_helm_repository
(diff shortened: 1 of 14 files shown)


mnaser (changes requested):
The old project owner suggested this project:

https://github.com/TazTheManiac/cert-manager-webhook-infoblox

Any reason why we wouldn't pick that one, @yaguangtang ?

yaguangtang:
> The old project owner suggested this project:
> 
> https://github.com/TazTheManiac/cert-manager-webhook-infoblox
> 
> Any reason why we wouldn't pick that one, Yaguang Tang (@yaguangtang) ?

For TazTheManiac/cert-manager-webhook-infoblox, the solverName changes from infoblox-wapi to infoblox, which I think it's not compatible with the existing customer env to upgrade.

mnaser:
Out of curiosity, is the solver name something that is hard coded in it or we can change it in the chart values?

yaguangtang:
> Out of curiosity, is the solver name something that is hard coded in it or we can change it in the chart values?

It's hard coded 
https://github.com/sarg3nt/cert-manager-webhook-infoblox-wapi/blob/main/main.go#L117
https://github.com/TazTheManiac/cert-manager-webhook-infoblox/blob/master/main.go#L36

PR #3975, opened 2026-05-21
Update Neutron server URLs to use openstack.svc.cluster.local

mpiscaer:
In our current setup we use a Windows DNS as upstream DNS server. When a non-existing TLD gets asked, a ServFail returns.

To solve this issue, the suffix openstack.svc.cluster.local gets added. This also makes the query more performant (it does not use the DNS domain search).

### DIFF
--- a/releasenotes/notes/update-neutron-policy-02161e0eb01b7c95.yaml
+++ b/releasenotes/notes/update-neutron-policy-02161e0eb01b7c95.yaml
@@ -1,0 +1,4 @@
+---
+fixes:
+  - |
+    To solve this issue, the suffix openstack.svc.cluster.local gets added. This also makes the query more performant (it does not use the DNS domain search).
--- a/roles/neutron/vars/main.yml
+++ b/roles/neutron/vars/main.yml
@@ -161,9 +161,9 @@ __neutron_ovn_helm_values:
 __neutron_policy_server_helm_values:
   conf:
     policy:
-      delete_port: "((rule:admin_only) or (rule:service_api) or role:member and rule:network_owner or role:member and project_id:%(project_id)s) and http://neutron-server:9697/port-delete"
-      update_port:mac_address: "((rule:admin_only) or (rule:service_api) or role:manager and project_id:%(project_id)s) and http://neutron-server:9697/port-update"
-      update_port:fixed_ips: "((rule:admin_only) or (rule:service_api) or role:manager and project_id:%(project_id)s or role:member and rule:network_owner) and http://neutron-server:9697/port-update"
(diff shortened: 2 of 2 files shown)


mnaser:
@mpiscaer Are you able to do the same test with `neutron-server.` as the hostname, I think the trailing dot will avoid doing DNS search

mpiscaer:
@mnaser If I query `neutron-server.` coreDNS will always fail with a error (NXdomain or ServFail). coreDNS always expect a full FQDN. The pod is responsible to use the search domain to fill in the rest of the FQDN.

mnaser:
I wonder what about neutron.openstack.svc.

Since the cluster local is actually configurable .. last shot :)

I still think trailing dot even with cluster local is good as well

fitbeard:
I agree. cluster.local should be last resort. Jwt token content relies on this in some places.

mpiscaer:
@mnaser @fitbeard I don't understand what issue you both bring up?

Is it about that in the PR the full FQDN is used and so you are not flexable in chainging it when the installed k8s uses a different suffix? because this is ussed multiple times for example https://github.com/vexxhost/atmosphere/blob/main/roles/neutron/vars/main.yml also set it.

Or do I miss something?

fitbeard:
@mpiscaer Let me explain. Example from Neutron vars is also not perfect (all places where we use hardcoded `.cluster.local`). Because if we introduce new configurable option for Kubernetes domain configuration `--cluster-domain=custom-domain.internal` everything will break. Not only this logic for Neutron Policy server. So for now the best way to start changing existing things is start using `pod.namespace.svc` (for services) form.

mnaser:
yeah, i agree with @fitbeard -- if we have full .cluster.local usages, i think it would be wise to start slowly cleaning those up to just be .svc.

mpiscaer:
Personally, I never liked the DNS search system, I think it introduces a security and latency issue.

So I would like to avoid too use of DNS search. I can imagine using a suffix variable to complete the FQDN.

when something like neutron-server.openstack.svc is used, then client dns seach system still needs to do multiple DNS query's to coreDNS to get a working answer.

When the pod /etc/resolv.conf has `search openstack.svc.cluster.local svc.cluster.local cluster.local localdomain.tld`

It will query for:
1. `neutron-server.openstack.svc`
2. `neutron-server.openstack.svc.openstack.svc.cluster.local`
3. `neutron-server.openstack.svc.svc.cluster.local`
4. `neutron-server.openstack.svc.cluster.local`

And if it gets decided to use the short name and let the DNS search do the rest, I would like to vote to use `neutron-server` as it right now. This will generate fewer extra DNS queries.

PR #3984, opened 2026-05-24
fix(alerts): correct node network alert annotations

yaguangtang:
Node receive and transmit alert rules evaluate an error ratio. The old annotations described $value as an absolute count. That could show "0 errors" even when the ratio threshold fired.

Override both alerts in the Atmosphere node mixin. Report the current error rate with humanizePercentage. Use a shared networkErrorRateThreshold in both rule and text.

Add focused promtool regression tests for both alerts. Add a reno note describing the user-visible fix.


yaguangtang:
sample alerts before this fix
```
  Annotations:
   - description = prod1 interface bond0 has encountered 0 receive errors in the last two minutes.
   - summary = Network interface is reporting many receive errors.
```

PR #3997, opened 2026-05-28
feat(monitoring): run ceph exporter as daemonset

larainema:
## Summary
- run ceph-exporter as an optional DaemonSet in the monitoring namespace, scheduled to Ceph nodes with the Kubernetes-native `ceph=enabled` node label
- scrape ceph-exporter through a ServiceMonitor and Kubernetes service discovery
- keep the existing Ceph mgr ServiceMonitor enabled as the perf-counter fallback for clusters that do not yet run ceph-exporter
- label Ceph inventory nodes through the kubernetes_node_labels role and document the monitoring change with Reno

## Validation
- git diff --check
- YAML parse for touched role files
- GOCACHE=/private/tmp/go-build-cache go test ./roles/kube_prometheus_stack -run "Test(CreateKeycloakRealmTask|AddClientRolesInIdTokenTask|CreateKeycloakClientsTask|CreateKeycloakRolesTask)"
- AIO test: temporary ceph-exporter DaemonSet scraped metrics successfully, including `ceph_osd_op_w_in_bytes`

### DIFF
--- a/releasenotes/notes/ceph-exporter-monitoring-fallback-cc10d38474f13f08.yaml
+++ b/releasenotes/notes/ceph-exporter-monitoring-fallback-cc10d38474f13f08.yaml
@@ -1,0 +1,9 @@
+---
+features:
+  - |
+    Added optional Prometheus scrape configuration for ``ceph-exporter``
+    targets in the kube-prometheus-stack role. The existing Ceph mgr
+    ServiceMonitor remains enabled by default so deployments can continue to
+    rely on mgr perf-counter collection while operators roll out the exporter.
+    Operators can enable the exporter scrape by populating
+    ``kube_prometheus_stack_ceph_exporter_http_sd_configs``.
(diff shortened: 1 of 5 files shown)


mnaser (changes requested):
I think this is the right initial way however what I would prefer is that we run `ceph-exporter` directly inside of the `monitoring` namespace.

This would give us the benefit of being able to use native service discovery using Kubernetes to run and scrape this, so I think what we need to do is add a "ceph" label or something more Kubernetes native for nodes which are running Ceph, and then run it as a `DaemonSet` only on those systems and scrape them using Prometheus.

larainema:
Addressed the requested refactor in the latest commits:

- ceph-exporter now runs directly in the `monitoring` namespace as a DaemonSet.
- Ceph nodes are labeled with `ceph=enabled` through `kubernetes_node_labels`, and the DaemonSet uses that nodeSelector.
- Prometheus scrapes the exporter through a ServiceMonitor/Kubernetes service discovery.
- The existing Ceph mgr ServiceMonitor remains enabled as the perf-counter fallback for deployments that need more time before ceph-exporter is available.
- AIO validation confirmed the exporter exposes the expected perf-counter metrics, including `ceph_osd_op_w_in_bytes`.

PR #4004, opened 2026-06-08
feat(monitoring): add libvirt stale mount alerts

larainema:
## Summary
- Add libvirt Prometheus alerts for repeated pod restarts and duplicate Ceph-related bind mounts.
- Add a node-exporter textfile sidecar that exports libvirt/Ceph stale mount gauges from host mountinfo.
- Add Alertmanager inhibition, runbook documentation, tests, and a release note.

## Testing
- jsonnet -J files/jsonnet -J files/jsonnet/vendor files/jsonnet/rules.jsonnet >/private/tmp/atmosphere-rules.json
- git diff --check --cached
- CGO_ENABLED=0 go test -v -run 'TestPrometheusRules/Libvirt' ./roles/kube_prometheus_stack/

### DIFF
--- a/doc/source/admin/monitoring.rst
+++ b/doc/source/admin/monitoring.rst
@@ -1057,6 +1057,51 @@ intervention and may precede a crash.
 5. If the error recurs, schedule hardware maintenance and replace the
    affected CPU or motherboard as needed.
 
+``LibvirtPodRestarts``
+======================
+
+This alert fires when a libvirt pod's main container restarts more than three
+times on the same node within one hour and the condition remains true for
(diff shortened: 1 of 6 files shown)


larainema:
recheck

mnaser (changes requested):
[Dong Ma (@larainema)](https://github.com/larainema) if we're going to have coverage for this, then we should try and figure out a way to actually use existing metrics for mounts and track anything that has a large # of mounts, we should be able to look in the historical customer data of prometheus to see if something in there can give us that data.

PR #4006, opened 2026-06-10
feat(monitoring): add native incident.io alerting support

guilhermesteinmuller:


doc/source/admin/monitoring.rst
mnaser, line 1: Historically, we have not really created new variables for the specific provider and just inlined them since they are not really "secrets".  So we should get rid of the `incident_io` variables OR decide if we want to start making this a native integration so instead it would be:

```
kube_prometheus_stack_incident_io_...: 
```

And when they are set, the config automatically gets set/enabled.

roles/defaults/vars/main.yml
mnaser, line 1: Are all the images updated her eto match the new version of the kube-prometheus-stack?

mnaser (changes requested):

roles/defaults/vars/main.yml
guilhermesteinmuller: yeah, based on my tests, these are the versions necessary. Actually. for alertmanager, incident.io notifier is available on v0.27.0, however, why not just push to the latest? 

For prometheus, i really pushed the exact version the notifier is available, not the latest. Thats were all the chart updates are coming from

doc/source/admin/monitoring.rst
guilhermesteinmuller: yeah i mean, we can remove it, not sure putting a vendor monitoring as native would be a good thing ? Just actuallty wanted to put in there becasue opsgenie requires just the api_key, whereas incident.io require 4 keys for both notifier and heartbeats... 

I guess then we are converging on getting rid of those in docs. Please confirm and i can adjust the pr

PR #4010, opened 2026-06-15
fix(ci): install docutils only for Vale runs

ricolin:
## Summary

Make the Vale job prepare its documentation lint dependencies only when there are documentation or release-note files to lint.

## Changes

- fetch full history for the Vale checkout so the job can diff against the base range
- find changed files under `doc` and `releasenotes`
- run `apt-get update` before installing `docutils` to avoid stale hosted-runner apt metadata
- install `docutils` and run Vale only when matching files changed
- add the required release note for the CI cleanup

## Validation

- `git diff --check`
- parsed `.github/workflows/ci.yaml` and the release note with PyYAML
- `vale releasenotes/notes/vale-docutils-conditional-install-47e25a92f4efab30.yaml`
- `uvx reno report --output /tmp/atmosphere-pr4010-reno.rst`

Context: repeated Vale failures on hosted runners were caused by stale apt metadata while installing `docutils`; this keeps the apt refresh and avoids the install entirely when Vale has no files to lint.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -56,10 +56,29 @@ jobs:
           egress-policy: audit
 
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
-      - run: sudo apt-get install -y docutils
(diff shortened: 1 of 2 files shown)


mnaser:
Dropped Vale for now, we can steer better with an `AGENTS.md`.

PR #4049, opened 2026-06-15
ci(zuul): use shared molecule templates

mnaser:
The molecule jobs now come from the shared `atmosphere-molecule-jobs`
template, so the local job definitions and dedicated pre/post playbooks
are no longer needed.


mnaser:
recheck

PR #4066, opened 2026-06-16
feat(monitoring): add host network failure alerts

ricolin:
## Summary

- tune `NodeBondingDegraded` to alert after 1 minute with a production runbook
- add `NodePhysicalNetworkTransmitErrors` for physical NIC transmit errors using existing `node-exporter` and `ethtool` metrics
- add promtool coverage, monitoring documentation, and a reno release note
- fix two existing MySQL alert fixtures so the full rule suite passes with sustained `for: 5m` conditions

## Why

Bond failure can precede Kubernetes node unreachable alerts by roughly 16 minutes. These alerts are vendor-neutral and fit all production environments because they use the kube-prometheus-stack/node-exporter path already deployed by Atmosphere.

## Validation

- `go test ./roles/kube_prometheus_stack`
- `git diff --check HEAD~1..HEAD`

### DIFF
--- a/doc/source/admin/monitoring.rst
+++ b/doc/source/admin/monitoring.rst
@@ -710,6 +710,125 @@ talking with ``etcd`` with the following commands:
       --from-file=/etc/kubernetes/pki/etcd/healthcheck-client.crt \
       --from-file=/etc/kubernetes/pki/etcd/healthcheck-client.key
 
+``NodeBondingDegraded``
+=======================
+
+This alert fires when ``node-exporter`` reports that a bonded interface has
+fewer active slaves than configured slaves for more than 1 minute:
+
+.. code-block:: text
+
+  node_bonding_slaves - node_bonding_active != 0
+
+This is an early host-network signal. If the affected bond carries management,
+storage, or tenant underlay traffic, it can precede downstream alerts such as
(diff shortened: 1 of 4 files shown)


ricolin:
Scenario validation result for `ef209b2c`:

- `go test ./roles/kube_prometheus_stack -count=1` passed.
- `git diff --check HEAD~1..HEAD` passed.
- Installed the PR collection into an existing AIO environment and ran a `kube-prometheus-stack` tagged update.
- First tagged update completed with `ok=36 changed=1 failed=0` and exit `0`.
- Re-ran the same tagged update for repeat-update/idempotency coverage; it completed with `ok=36 changed=0 failed=0` and exit `0`.
- Runtime validation after the update:
  - `kube-prometheus-stack` Helm release remained `deployed`
  - monitoring pods remained ready
  - `NodeBondingDegraded` and `NodePhysicalNetworkTransmitErrors` were present in the rendered `PrometheusRule`
  - both rules were mounted into Prometheus rule files
  - Prometheus `/api/v1/rules` reported both rules loaded and inactive before scenario triggering
- Triggered both alerts with a temporary synthetic `node-exporter` scrape target, without changing real host NIC or bond state:
  - `NodeBondingDegraded` reached `firing`
  - `NodePhysicalNetworkTransmitErrors` reached `firing`
- Removed the temporary scrape target resources and confirmed the synthetic alert instances cleared from Prometheus afterward.

Issue #2671, opened 2025-05-20
Keycloak upgrade challenges


fitbeard:
I will list here the discoveries I made while configuring high reliability for Keycloak.

A few details about the current configuration:

- We are using Keycloak with mysql driver/backend
- We are using PXC which by default do not allows tables without indexes (temporary setting PXC  strict mode to `PERMISSIVE`)

The last nuance made me anxious when testing updates and clean installations of newer versions:)

1. Both upgrades from 24.x and clean installs directly to 26.1.x and 26.2.x ended with something like this:

` WSREP has not yet prepared node for application use [Failed SQL: (1047) DELETE FROM keycloak.USER_CONSENT WHERE ID IN (SELECT ID FROM TEMP_USER_CONSENT_IDS)]`

This is message form PXC engine in Keycloak's bootstrap logs. PXC cluster hanged and pods restarted. Only manual internevntion helped restore PXC cluster to normal state.

One thin I noticed is that 26.x.y is using newer liquidbase for migrations. Keycloak =< 25.x is using `4.25.1` and starting 26.0.x liquidbase is switching to `4.29.1`

This data can be retireved from table `DATABASECHANGELOG`

Something is so bad that PXC with PERMISSINE and DISABLED strict mode hangs. Actually dies.
We need to investigate this in a future.

2. Upgrade from 24.x and clean install directly to 25.x works as expected.
3. To upgrade and use Keycloak 25.x we need to:

  - Bump `community.general` collection at least to `10.6.0`: https://github.com/ansible-collections/community.general/pull/9976
  - Bump `python-keycloak` and test dependencies for https://github.com/vexxhost/keystone-keycloak-backend with Keycloak 25.x . Right now backend is not working because something was changed in Keycloak 25.x related to refresh sessions. Sorry i forgot collect logs from Keystone.

fitbeard:
@mnaser that's was my second plan to try.

fitbeard:
@mnaser bad news. With latest 8.0 https://github.com/vexxhost/atmosphere/pull/2675 situation is the same - crashed PXC.

mnaser:
@fitbeard this is really stinky, do you have the traceback somewhere that we can report to pxc upstream?  seems like crashing a pxc isn't.. ideal

fitbeard:
Low chance of getting an answer but I will try to crash my lab PXC again and collect evidence for a bug report.

fitbeard:
[...]

PR #4096, opened 2026-06-22
fix(keycloak): upgrade bundled server to 26.6.3

larainema:
## Summary
- bump the vendored Bitnami Keycloak chart from 21.4.1 to 22.2.3
- bump the bundled upstream Keycloak image from 24.0.5 to 26.6.3, pinned by registry manifest digest
- update Keycloak proxy configuration for the newer server
- expose the Keycloak management port for health and metrics
- move the kube-prometheus-stack Keycloak scrape endpoint to the management metrics port

## Upgrade path
This PR targets a direct Keycloak upgrade from 24.0.5 to 26.6.3.

Recommended operator flow:

1. Take a PXC backup before upgrading Keycloak.
2. Apply the Atmosphere upgrade.
3. Verify Keycloak readiness and OIDC discovery.
4. Verify Keystone federation against the upgraded Keycloak realm.
5. Verify Keycloak health and metrics on the management port, 9000.
6. Verify PXC is ready and strict mode has been restored.

Issue #2671 reported PXC failures with earlier 26.x migrations, so operators should keep the pre-upgrade PXC backup until post-upgrade Keycloak and Keystone checks are complete.


mnaser (commented):
Thanks for pushing this forward. We do need to get off Keycloak 24.x for security reasons, but I do not think we should merge this as a direct `24.0.5 -> 26.6.3` upgrade against the current shared PXC-backed database path until the PXC/Liquibase risk has a concrete answer.

Context for the next reviewer/operator:

- The main blocker is tracked in https://github.com/vexxhost/atmosphere/issues/2671. Tadas reproduced failures upgrading from 24.x and also with clean installs of 26.1.x / 26.2.x using the current MySQL/PXC backend. The failure was around the Keycloak 26 Liquibase migration path, including `TEMP_USER_CONSENT_IDS` / `USER_CONSENT` SQL, and resulted in PXC hanging/restarting rather than a normal failed Keycloak rollout.
- We already tried the obvious PXC component bump. https://github.com/vexxhost/atmosphere/pull/2675 updated the PXC operator/engine path, and Tadas reported back on #2671 that the situation was the same: latest 8.0 still crashed PXC.
- This was also reported upstream to Percona as https://perconadev.atlassian.net/browse/PXC-4689. Percona closed it as `Cannot Reproduce`, not as fixed. The follow-up https://perconadev.atlassian.net/browse/PXC-4824 was closed as a duplicate. So from our side this remains an unresolved operational risk, not an upstream-resolved bug.
- https://github.com/vexxhost/atmosphere/pull/2669 is relevant but separate: it enables HA/scaling for the current 24.x line and references the same upgrade findings. It does not solve the 26.x database migration problem.
- There is also WIP branch work toward introducing CloudNativePG/PostgreSQL resources: https://github.com/vexxhost/atmosphere/tree/feat/add_postgres_resources. That branch adds CloudNativePG/Barman charting and roles plus a `cnpg-db` cluster, but it is broad/WIP and does not yet appear to provide a completed Keycloak data migration path.

The specific concern with this PR is that the current operator flow still relies on taking a PXC backup, applying the Keycloak upgrade, and then verifying PXC recovered. That is a reasonable recovery measure, but it is not enough evidence that the upgrade is safe to ship by default. The issue we saw was not just Keycloak failing readiness; it was a database-cluster destabilization scenario.

Before merging this, I think we need one of these paths documented and validated:

1. A successful direct `24.0.5 -> 26.6.3` upgrade in an Atmosphere-like environment with existing Keycloak data on PXC, including PXC health/replication checks before and after.
2. A tested maintenance procedure that temporarily sets `pxc_strict_mode=PERMISSIVE` consistently via PXC configuration across all nodes, confirms every node is in that mode, performs the upgrade, validates Keycloak/Keystone/PXC, and restores `MASTER` afterward. The current runtime `SET GLOBAL` approach in the role is probably not sufficient as proof because it may not be cluster-wide or persistent across node restarts.
3. Preferably, split Keycloak away from the shared OpenStack PXC backend and migrate it to a dedicated supported database backend, likely PostgreSQL/CloudNativePG, then run the Keycloak 26 upgrade there.

So my recommendation is: keep this PR as the version-upgrade vehicle, but do not merge it until the PXC risk is either reproduced-and-cleared in lab, handled by a tested temporary maintenance flow, or avoided by moving Keycloak to PostgreSQL.

larainema:
@mnaser Thanks for the detailed context. I agree we should not merge this as a plain Keycloak version bump while #2671 remains unresolved.

I did a stronger follow-up validation on the AIO environment. The earlier successful AIO upgrade should not be treated as proof that the PXC path is safe, because it did not exercise the problematic consent migration path.

On the AIO PXC cluster, with `pxc_strict_mode=MASTER`, the exact Keycloak 25.0.0 MySQL consent migration SQL shape fails with:

```text
ERROR 1105: Percona-XtraDB-Cluster prohibits use of CREATE TABLE AS SELECT with pxc_strict_mode = ENFORCING or MASTER
```

When temporarily switching PXC to PERMISSIVE, the same probe passes, and after restoring MASTER, PXC remained healthy.

So this confirms the blocker is real for the current shared PXC path.

@fitbeard could you please help with the follow-up and comment on the preferred path forward here? It sounds like the options are either:
1. prove and document a controlled PXC maintenance procedure around pxc_strict_mode=PERMISSIVE, including before/after PXC health checks, or
2. continue the PostgreSQL/CloudNativePG direction so Keycloak is no longer coupled to the shared OpenStack PXC cluster.

I’ll keep this PR blocked from merge until we have one of those paths validated.

fitbeard:
@larainema @mnaser hello. PXC never was and i think never will be supported backend for Keycloak: https://www.keycloak.org/server/db -> `'It is not a supported configuration if the underlying database specific Hibernate dialect allows the use of a version that differs from those shown.'` I think we were lucky and unlucky at the same time when introduced Keycloak with such as configuration at the beginning (we always played with set/unset PERMISSIVE). As Mohammed mentioned all collected facts - all previous attempts crashed. Two reproducible proofs registered in PXC community bug track with no luck. Another thing - we still rely on old Bitnami chart which never was design for HA. I think we have more to do here - migrate data to supported backend, migrate to Keyclaok official operator, introduce session cache using Infinispan or Redis(poc) backends. I vote for CloudnativePG here.

PS. I did rnd outside of our product how to migrate data:

  - Database 'transformation' from Mysql to Postgres is not possible
  - Got successful results by exporting and importing whole data. But i tested with very very small installation and amount of data:

```
docker run -ti --rm \
  --network keycloak-mysql-to-postgres_keycloak \
  -v $PWD/export:/tmp \
  -e KEYCLOAK_ADMIN=admin \
  -e KEYCLOAK_ADMIN_PASSWORD=password123 \
  -e KC_DB_USERNAME=keycloak \
  -e KC_DB_PASSWORD=password123 \
  -e KC_DB=mysql \
  -e KC_DB_URL="jdbc:mysql://keycloak-mysql:3306/keycloak" \
  quay.io/keycloak/keycloak:24.0.5-0 export --file /tmp/keycloak-export.json

docker run -ti --rm \
  --network keycloak-mysql-to-postgres_keycloak \
  -v $PWD/export:/tmp \
  -e KEYCLOAK_ADMIN=admin \
  -e KEYCLOAK_ADMIN_PASSWORD=password123 \
  -e KC_DB_USERNAME=keycloak \
  -e KC_DB_PASSWORD=password123 \
  -e KC_DB=postgres \
  -e KC_DB_URL="jdbc:postgresql://keycloak-postgres:5432/keycloak" \
  quay.io/keycloak/keycloak:24.0.5-0 import --file /tmp/keycloak-export.json
```

PR #4109, opened 2026-07-02
feat(storage): add NetApp support on stable/2025.2

larainema:
## Summary

This integration PR is based on `stable/2025.2` and cherry-picks the pending NetApp storage changes onto a released 7.x baseline.

Included changes:

- Trident CSI support for Kubernetes PVCs
- Trident Fibre Channel protocol support
- First-class NetApp ONTAP Cinder backend support via `type: netapp`

The intent is to provide a stable branch for validating NetApp-backed Kubernetes PVCs and OpenStack Cinder volumes before the individual feature PRs are merged.

## Usage

Use this branch directly for validation:

```bash
git fetch origin feat/netapp-storage-stable-2025.2
git switch --detach origin/feat/netapp-storage-stable-2025.2
```

Configure both storage layers in inventory:

- `csi_driver: trident` for Kubernetes PVCs
- `atmosphere_storage.volumes.backends.<name>.type: netapp` for OpenStack Cinder

### DIFF
--- a/doc/source/deploy/csi.rst
+++ b/doc/source/deploy/csi.rst
@@ -128,12 +128,28 @@ Trident CSI driver by updating your Ansible inventory as follows:
 
 Use ``trident`` as the Atmosphere ``csi_driver`` value.  The Kubernetes CSI
 provisioner remains ``csi.trident.netapp.io``, and the backend driver is
-selected separately with values such as ``ontap-san`` or ``ontap-nas``.
+selected separately with values such as ``ontap-san``,
+``ontap-san-economy``, or ``ontap-nas``.
 
 By default, Atmosphere configures Trident with the ``ontap-san`` storage driver
-and creates the ``general`` StorageClass using the
-``csi.trident.netapp.io`` provisioner. This is appropriate for
-``ReadWriteOnce`` PVCs backed by iSCSI LUNs.
+and ``trident_csi_access_protocol: iscsi``. It creates the ``general``
+StorageClass using the ``csi.trident.netapp.io`` provisioner. This is
+appropriate for ``ReadWriteOnce`` PVCs backed by iSCSI LUNs.
+
+To use Fibre Channel with the ``ontap-san`` driver, set
+``trident_csi_access_protocol: fc``:
+
+.. code-block:: yaml
+
+    csi_driver: trident
+    trident_csi_storage_driver: ontap-san
+    trident_csi_access_protocol: fc
+    trident_csi_management_lif: <FILL IN>
+    trident_csi_svm: <FILL IN>
+    trident_csi_username: <FILL IN>
+    trident_csi_password: <FILL IN>
+
+The ``ontap-san-economy`` driver supports iSCSI only.
 
 If you need shared ``ReadWriteMany`` PVCs, use an ONTAP NAS backend instead:
 
@@ -159,12 +175,20 @@ backend object, and StorageClass without waiting for a successful backend bind.
 .. admonition:: SAN multipathing
     :class: warning
 
-    For ``ontap-san`` backends, Trident requires working iSCSI and multipath
-    configuration on every Kubernetes node. Atmosphere installs the host
-    packages, ensures an iSCSI initiator name exists, and configures
-    ``find_multipaths no`` by default. You can disable the multipath edit by
-    setting ``trident_csi_configure_multipath: false`` if your hosts already
-    manage that configuration elsewhere.
+    For ``ontap-san`` backends, Trident requires working SAN host preparation
+    and multipath configuration on every Kubernetes node. Atmosphere
+    configures ``find_multipaths no`` by default.
+
+    For iSCSI, Atmosphere installs the Trident iSCSI host packages, starts the
+    required services, and ensures an initiator name exists.
+
+    For Fibre Channel, Atmosphere installs the FC helper and multipath
+    packages that Trident expects. FC switch zoning and host HBA presentation
+    still need to be in place before you deploy the role.
+
+    You can disable the multipath edit by setting
+    ``trident_csi_configure_multipath: false`` if your hosts already manage
+    that configuration elsewhere.
 
 *********
 IBM Block
--- a/doc/source/deploy/csi.rst
+++ b/doc/source/deploy/csi.rst
@@ -111,6 +111,61 @@ PowerStore configuration. This includes specifying the block protocol, which
 can either be Fibre Channel (FC) or iSCSI, depending on your network
 infrastructure.
 
+**************
+NetApp Trident
+**************
+
+For environments using NetApp ONTAP storage such as AFF arrays, configure the
+Trident CSI driver by updating your Ansible inventory as follows:
+
+.. code-block:: yaml
+
+    csi_driver: trident
+    trident_csi_management_lif: <FILL IN>
+    trident_csi_svm: <FILL IN>
+    trident_csi_username: <FILL IN>
+    trident_csi_password: <FILL IN>
+
+Use ``trident`` as the Atmosphere ``csi_driver`` value.  The Kubernetes CSI
+provisioner remains ``csi.trident.netapp.io``, and the backend driver is
+selected separately with values such as ``ontap-san`` or ``ontap-nas``.
+
+By default, Atmosphere configures Trident with the ``ontap-san`` storage driver
+and creates the ``general`` StorageClass using the
+``csi.trident.netapp.io`` provisioner. This is appropriate for
+``ReadWriteOnce`` PVCs backed by iSCSI LUNs.
+
+If you need shared ``ReadWriteMany`` PVCs, use an ONTAP NAS backend instead:
+
+.. code-block:: yaml
+
+    csi_driver: trident
+    trident_csi_storage_driver: ontap-nas
+    trident_csi_management_lif: <FILL IN>
+    trident_csi_data_lif: <FILL IN>
+    trident_csi_svm: <FILL IN>
+    trident_csi_username: <FILL IN>
+    trident_csi_password: <FILL IN>
+    trident_csi_fs_type: ""
+
+You can pass additional backend options through
+``trident_csi_backend_options`` and additional StorageClass parameters through
+``trident_csi_storage_class_parameters``.
+
+For test environments that do not have access to ONTAP yet, set
+``trident_csi_wait_for_backend: false`` so the role can validate the operator,
+backend object, and StorageClass without waiting for a successful backend bind.
+
+.. admonition:: SAN multipathing
+    :class: warning
+
+    For ``ontap-san`` backends, Trident requires working iSCSI and multipath
+    configuration on every Kubernetes node. Atmosphere installs the host
+    packages, ensures an iSCSI initiator name exists, and configures
+    ``find_multipaths no`` by default. You can disable the multipath edit by
+    setting ``trident_csi_configure_multipath: false`` if your hosts already
+    manage that configuration elsewhere.
+
 *********
 IBM Block
 *********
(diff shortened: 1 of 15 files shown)


doc/source/deploy/csi.rst
```diff
@@ -111,6 +111,85 @@ PowerStore configuration. This includes specifying the block protocol, which
...
+**************
+
+For environments using NetApp ONTAP storage such as AFF arrays, configure the
+Trident CSI driver by updating your Ansible inventory as follows:
+
+.. code-block:: yaml
+
+    csi_driver: trident
+    trident_csi_management_lif: <FILL IN>
+    trident_csi_svm: <FILL IN>
```
ntap-johanneswagner: Specifying the svm name is only necessary if cluster credentials are used and can even harm in some configurations (ex. Metrocluster). I don't see any need for it and would exclude this.

doc/source/deploy/csi.rst
```diff
@@ -111,6 +111,85 @@ PowerStore configuration. This includes specifying the block protocol, which
...
+appropriate for ``ReadWriteOnce`` PVCs backed by iSCSI LUNs.
+
+To use Fibre Channel with the ``ontap-san`` driver, set
+``trident_csi_access_protocol: fc``:
+
+.. code-block:: yaml
+
+    csi_driver: trident
+    trident_csi_storage_driver: ontap-san
+    trident_csi_access_protocol: fc
```
ntap-johanneswagner, line 154: FC is only supported if it is bare metal deployment, not sure whether you really want to offer this?

doc/source/deploy/csi.rst
```diff
@@ -111,6 +111,85 @@ PowerStore configuration. This includes specifying the block protocol, which
...
+    trident_csi_management_lif: <FILL IN>
+    trident_csi_svm: <FILL IN>
```
ntap-johanneswagner: Specifying the svm name is only necessary if cluster credentials are used and can even harm in some configurations (ex. Metrocluster). I don't see any need for it and would exclude this.

doc/source/deploy/csi.rst
```diff
@@ -111,6 +111,85 @@ PowerStore configuration. This includes specifying the block protocol, which
...
+
+If you need shared ``ReadWriteMany`` PVCs, use an ONTAP NAS backend instead:
+
+.. code-block:: yaml
+
+    csi_driver: trident
+    trident_csi_storage_driver: ontap-nas
+    trident_csi_management_lif: <FILL IN>
+    trident_csi_data_lif: <FILL IN>
+    trident_csi_svm: <FILL IN>
```
ntap-johanneswagner: Specifying the svm name is only necessary if cluster credentials are used and can even harm in some configurations (ex. Metrocluster). I don't see any need for it and would exclude this.

doc/source/deploy/csi.rst
```diff
@@ -111,6 +111,85 @@ PowerStore configuration. This includes specifying the block protocol, which
...
+The ``ontap-san-economy`` driver supports iSCSI only.
+
+If you need shared ``ReadWriteMany`` PVCs, use an ONTAP NAS backend instead:
+
+.. code-block:: yaml
+
+    csi_driver: trident
+    trident_csi_storage_driver: ontap-nas
+    trident_csi_management_lif: <FILL IN>
+    trident_csi_data_lif: <FILL IN>
```
ntap-johanneswagner: data lif is only necessary if multiple nfs lifs are present and a specific should be chosen. I usually recommend to avoid multiple data lifs at the svm or use the round robin. Defining it in the backend can result in less flexibility for storage maintenance/ tech refresh scenarios. 
In this case the config is for san/san-economy, where it is even not allowed to specify a data lif as it would kill multipathing.

doc/source/deploy/csi.rst
```diff
@@ -111,6 +111,85 @@ PowerStore configuration. This includes specifying the block protocol, which
...
+.. code-block:: yaml
+
+    csi_driver: trident
+    trident_csi_storage_driver: ontap-san
+    trident_csi_access_protocol: fc
+    trident_csi_management_lif: <FILL IN>
+    trident_csi_svm: <FILL IN>
+    trident_csi_username: <FILL IN>
+    trident_csi_password: <FILL IN>
+
```
ntap-johanneswagner, line 159: fstype is missing. While it's not necessary, as we fallback to ext4 as a default, it blocks the usage of fsGroup in K8s which can create issues.

doc/source/deploy/csi.rst
```diff
@@ -111,6 +111,85 @@ PowerStore configuration. This includes specifying the block protocol, which
...
+.. code-block:: yaml
+
+    csi_driver: trident
+    trident_csi_storage_driver: ontap-nas
+    trident_csi_management_lif: <FILL IN>
+    trident_csi_data_lif: <FILL IN>
+    trident_csi_svm: <FILL IN>
+    trident_csi_username: <FILL IN>
+    trident_csi_password: <FILL IN>
+    trident_csi_fs_type: ""
```
ntap-johanneswagner: Here should be either a "`<Fill IN>"` or "ext4". The available options are ext3, ext4 and xfs

doc/source/deploy/csi.rst
```diff
@@ -111,6 +111,85 @@ PowerStore configuration. This includes specifying the block protocol, which
...
+    trident_csi_svm: <FILL IN>
+    trident_csi_username: <FILL IN>
+    trident_csi_password: <FILL IN>
+
+Use ``trident`` as the Atmosphere ``csi_driver`` value.  The Kubernetes CSI
+provisioner remains ``csi.trident.netapp.io``, and the backend driver is
+selected separately with values such as ``ontap-san``,
+``ontap-san-economy``, or ``ontap-nas``.
+
+By default, Atmosphere configures Trident with the ``ontap-san`` storage driver
```
ntap-johanneswagner, line 134: How many pvcs would we expect? If its a larger amount, I would recommend to use the economy driver.

doc/source/deploy/csi.rst
```diff
@@ -111,6 +111,85 @@ PowerStore configuration. This includes specifying the block protocol, which
...
+    trident_csi_storage_driver: ontap-nas
+    trident_csi_management_lif: <FILL IN>
+    trident_csi_data_lif: <FILL IN>
+    trident_csi_svm: <FILL IN>
+    trident_csi_username: <FILL IN>
+    trident_csi_password: <FILL IN>
+    trident_csi_fs_type: ""
+
+You can pass additional backend options through
+``trident_csi_backend_options`` and additional StorageClass parameters through
```
ntap-johanneswagner, line 199: I would recommend to add the following two backend parameters as defaults:
1. useRest: false/true => Currently Trident decides based on ONTAP version whether to use REST or ZAPI. Unfortunately, REST is way slower than ZAPI. Because of this most of our customers don't want to have the automatic behavior but want to control it by this option.
2. custom naming template: 
`  defaults:
    nameTemplate: "{{ .volume.Name }}_{{ .volume.Namespace }}_{{ .volume.RequestName }}"`
By using a naming template like this, the volume name on ontap side is way better readable for humans. Per default trident names volumes like "trident_pvc_uuidofpvc". Using this naming template it would be pvc_k8snamespacewherethevolumeis_nameofthepvc which helps with troubleshooting.

roles/trident_csi/defaults/main.yml
```diff
@@ -0,0 +1,147 @@
...
+trident_csi_orchestrator_wait_retries: 120
+trident_csi_orchestrator_wait_delay: 10
+
+trident_csi_supported_storage_drivers:
+  - ontap-san
+  - ontap-san-economy
+  - ontap-nas
+  - ontap-nas-economy
+  - ontap-nas-flexgroup
+trident_csi_supported_access_protocols:
```
ntap-johanneswagner, line 48: we also allow NFS, SMB (windows nodes only) (all drivers that start with ontap-nas-*)and NVMe/TCP (ontap-san only)

roles/trident_csi/tasks/main.yml
```diff
@@ -0,0 +1,341 @@
...
+# under the License.
+
+---
+- name: Validate Trident CSI configuration
+  ansible.builtin.assert:
+    that:
+      - trident_csi_storage_driver in trident_csi_supported_storage_drivers
+      - trident_csi_access_protocol in trident_csi_supported_access_protocols
+      - trident_csi_management_lif is defined
+      - trident_csi_svm is defined
```
ntap-johanneswagner: Specifying the svm name is only necessary if cluster credentials are used and can even harm in some configurations (ex. Metrocluster). I don't see any need for it and would exclude this.

roles/trident_csi/templates/backend.yaml.j2
```diff
@@ -0,0 +1,23 @@
...
+  namespace: "{{ trident_csi_namespace }}"
+spec:
+  version: {{ trident_csi_backend_version }}
+  backendName: "{{ trident_csi_backend_name }}"
+  storageDriverName: "{{ trident_csi_storage_driver }}"
+  managementLIF: "{{ trident_csi_management_lif }}"
+{% if trident_csi_data_lif is defined and trident_csi_data_lif | string | length > 0 %}
+  dataLIF: "{{ trident_csi_data_lif }}"
+{% endif %}
+  svm: "{{ trident_csi_svm }}"
```
ntap-johanneswagner, line 20: Specifying the svm name is only necessary if cluster credentials are used and can even harm in some configurations (ex. Metrocluster). I don't see any need for it and would exclude this.

roles/trident_csi/templates/backend.yaml.j2
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: "{{ trident_csi_backend_config_name }}"
+  namespace: "{{ trident_csi_namespace }}"
+spec:
+  version: {{ trident_csi_backend_version }}
+  backendName: "{{ trident_csi_backend_name }}"
+  storageDriverName: "{{ trident_csi_storage_driver }}"
+  managementLIF: "{{ trident_csi_management_lif }}"
+{% if trident_csi_data_lif is defined and trident_csi_data_lif | string | length > 0 %}
+  dataLIF: "{{ trident_csi_data_lif }}"
```
ntap-johanneswagner, line 17: Especially as you currently use san backends only, this may not be defined and needs to be removed.

roles/trident_csi/templates/storageclass.yaml.j2
ntap-johanneswagner, line 1: There is a need to define the fsType in the StorageClass. While technically we will fall back to ext4 if nothing is defined, I still recommend to define it because fsGroup will not work without a fsType specified in the sc

ntap-johanneswagner (changes requested):
Reffering to the e-mail configuration with Dong Ma and Yaguang Tang, I've reviewed the code as far as I was able to and left some comments for changes.

doc/source/deploy/csi.rst
```diff
@@ -136,33 +136,59 @@ and ``trident_csi_access_protocol: iscsi``. It creates the ``general``
...
     csi_driver: trident
     trident_csi_storage_driver: ontap-nas
+    trident_csi_access_protocol: nfs
     trident_csi_management_lif: <FILL IN>
-    trident_csi_data_lif: <FILL IN>
-    trident_csi_svm: <FILL IN>
     trident_csi_username: <FILL IN>
     trident_csi_password: <FILL IN>
-    trident_csi_fs_type: ""
+    trident_csi_fs_type: ext4
```
ntap-johanneswagner: this needs to be fs_type: nfs . Setting ext4 would also work but as the FS for the ontap-nas driver with access protocol nfs will always be NFS no matter what you define here, the setting might confuse people.

ntap-johanneswagner (commented):
This looks good. All open comments can be marked as resolved. I tried to do this, however GitHub throws an error.

PR #4134, opened 2026-07-10
feat(keycloak): add PostgreSQL migration path

larainema:
## Summary
- add a PostgreSQL backend option for Keycloak using an externally managed application secret
- add a controlled MySQL/PXC-to-PostgreSQL export, import, and cutover playbook
- prevent a migrated deployment from being unintentionally returned to MySQL

## Follow-up
CloudNativePG installation, CRD lifecycle, backup/restore, and Kubernetes-version support remain in #3314. This change deliberately consumes a dedicated PostgreSQL service so it can be reviewed independently of that unfinished operator work.

This provides the prerequisite migration path for the Keycloak 26.6.3 upgrade in #4096.

### DIFF
--- a/playbooks/keycloak-migrate-to-postgresql.yml
+++ b/playbooks/keycloak-migrate-to-postgresql.yml
@@ -1,0 +1,18 @@
+# Copyright (c) 2026 VEXXHOST, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+- name: Migrate Keycloak from MySQL to PostgreSQL
+  hosts: controllers
+  become: true
+  pre_tasks:
+    - name: Require PostgreSQL as the configured Keycloak backend
+      ansible.builtin.assert:
+        that:
+          - keycloak_database_vendor | default('mysql') == 'postgres'
+        fail_msg: >-
+          Set keycloak_database_vendor to "postgres" in the deployment
+          configuration before running this migration playbook.
+  roles:
+    - role: keycloak_database_migration
+      tags:
+        - keycloak-database-migration
--- a/releasenotes/notes/keycloak-postgresql-migration-6450dd7cc07ef343.yaml
+++ b/releasenotes/notes/keycloak-postgresql-migration-6450dd7cc07ef343.yaml
@@ -1,0 +1,11 @@
+---
+features:
+  - Adds support for running Keycloak against a dedicated external PostgreSQL
+    database rather than the shared MySQL/PXC cluster.
+upgrade:
+  - Before upgrading Keycloak beyond 24.0.5, configure
+    ``keycloak_database_vendor`` as ``postgres`` along with an empty
+    PostgreSQL database and its application secret, then run
+    ``playbooks/keycloak-migrate-to-postgresql.yml``. The playbook stops
+    Keycloak, exports and imports the data with the image currently deployed,
+    and restores the MySQL-backed release if migration fails.
--- a/roles/keycloak/defaults/main.yml
+++ b/roles/keycloak/defaults/main.yml
@@ -28,7 +28,24 @@ keycloak_ingress_cluster_issuer: "{{ atmosphere_ingress_cluster_issuer }}"
 keycloak_admin_username: admin
 keycloak_admin_password: "{{ undef(hint='You must specify a Keycloak admin password using keycloak_admin_password') }}"
 
+keycloak_database_vendor: mysql
 keycloak_database_username: keycloak
 keycloak_database_password: "{{ undef('You must specify a Keycloak database password using keycloak_database_password') }}"
 keycloak_database_name: keycloak
+keycloak_postgresql_host: ""
+keycloak_postgresql_port: 5432
+keycloak_postgresql_secret_name: keycloak-postgresql
+keycloak_postgresql_secret_key: password
+keycloak_database_host: >-
+  {{ keycloak_postgresql_host if keycloak_database_vendor == 'postgres' else
+  openstack_helm_endpoints.oslo_db.hosts.default + '.openstack' }}
+keycloak_database_port: >-
+  {{ keycloak_postgresql_port if keycloak_database_vendor == 'postgres' else 3306 }}
+keycloak_database_jdbc_scheme: >-
+  {{ 'postgresql' if keycloak_database_vendor == 'postgres' else 'mysql' }}
+keycloak_database_existing_secret: >-
+  {{ keycloak_postgresql_secret_name if keycloak_database_vendor == 'postgres' else '' }}
+keycloak_database_password_secret_key: >-
+  {{ keycloak_postgresql_secret_key if keycloak_database_vendor == 'postgres' else 'db-password' }}
+keycloak_database_migration_marker_name: keycloak-postgresql-migration
 keycloak_host_tls_secret_name: "{{ openstack_helm_ingress_secret_name | default(keycloak_host + '-tls')}}"
--- a/roles/keycloak/tasks/main.yml
+++ b/roles/keycloak/tasks/main.yml
@@ -12,7 +12,45 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: Get the Kuberentes service for Percona XtraDB Cluster
+- name: Validate Keycloak database vendor
+  ansible.builtin.assert:
+    that:
+      - keycloak_database_vendor in ['mysql', 'postgres']
+    fail_msg: >-
+      keycloak_database_vendor must be either "mysql" or "postgres".
+
+- name: Validate PostgreSQL settings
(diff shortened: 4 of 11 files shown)


larainema:
@fitbeard , I reviewed #3314 as a follow-up to this work. The managed-CNPG path still needs a Kubernetes baseline decision, database/user ownership, and backup/restore design, so #4134 intentionally uses a dedicated external PostgreSQL service and does not depend on #3314.

Please help review this migration path and confirm the staged order: migrate Keycloak 24.0.5 to PostgreSQL first, validate the service and data, then apply the Keycloak 26.6.3 upgrade from #4096.

fitbeard:
@larainema we discussed with @mnaser about whole this thing. Let me finish testing in parallel data export/import in my test environment. Then i think it's good time to ditch Bitnami chart and switch to https://www.keycloak.org/operator/.

I think by letting users to choose backend vendor `keycloak_database_vendor` is not a good idea.

I think this should look like one major migration - new database, new operator, tests and cut off. The change should include cnpg+keycloak+operators.

larainema:
Thanks @fitbeard. I checked A8E-72 and comment 29340. The issue confirms the intended integrated migration to PostgreSQL, the official Keycloak Operator, and potentially the Infinispan Operator. However, comment 29340 raises the database/user reconciliation questions without recording a final decision.

I suggest using CloudNativePG Database resources and declarative managed roles, with the Keycloak credentials stored in a Kubernetes Secret used by both CNPG and the Keycloak CR. PostgreSQL TCP superuser access can then remain disabled. The backup/restore procedure must separately restore or rotate that Secret.

For the migration, I suggest retaining PostgreSQL 15 because the complete Keycloak 24.0.5 → PostgreSQL 15 → Keycloak 26.6.3 path has already been validated. PostgreSQL major upgrades can be handled separately.

Could you please share your export/import test results and confirm:

- The database/user reconciliation model.

- Whether the Infinispan Operator is still required.

- The Kubernetes upgrade target needed for a supported CNPG release.

- Whether #4134 should be reworked or replaced by the integrated migration.

Once confirmed, we can structure the code as stacked changes behind a disabled migration path, while performing the actual database and Keycloak cutover as one controlled maintenance operation.

fitbeard:
> Thanks Tadas Sutkaitis (@fitbeard). I checked A8E-72 and comment 29340. The issue confirms the intended integrated migration to PostgreSQL, the official Keycloak Operator, and potentially the Infinispan Operator. However, comment 29340 raises the database/user reconciliation questions without recording a final decision.
> 
> I suggest using CloudNativePG Database resources and declarative managed roles, with the Keycloak credentials stored in a Kubernetes Secret used by both CNPG and the Keycloak CR. PostgreSQL TCP superuser access can then remain disabled. The backup/restore procedure must separately restore or rotate that Secret.
> 
> For the migration, I suggest retaining PostgreSQL 15 because the complete Keycloak 24.0.5 → PostgreSQL 15 → Keycloak 26.6.3 path has already been validated. PostgreSQL major upgrades can be handled separately.
> 
> Could you please share your export/import test results and confirm:
> 
>     * The database/user reconciliation model.
> 
>     * Whether the Infinispan Operator is still required.
> 
>     * The Kubernetes upgrade target needed for a supported CNPG release.
> 
>     * Whether [feat(keycloak): add PostgreSQL migration path #4134](https://github.com/vexxhost/atmosphere/pull/4134) should be reworked or replaced by the integrated migration.
> 
> 
> Once confirmed, we can structure the code as stacked changes behind a disabled migration path, while performing the actual database and Keycloak cutover as one controlled maintenance operation.

@larainema now i have questions for you:)

  - I also like Kubernetes Secret used by both CNPG and the Keycloak CR...but user/role is part of CNPG Database CR which means secret lives in `openstack` namespace and we need this data later in `auth-system` namespace for Keycloak.
  - I don't know what to do with lowest supported Kubernetes version on Atmosphere (1.28.13). I think we can at least test incompatible versions. We have similar situation with PXC actually.

My poc results and observations:
  - I agree that we don't need superuser access using password and TCP.
  - I tested Keycloak with Postgres 18 - we are safe to go with newer version then 15.
  - We don't need dedicated Infinispan cluster - 26.x is using native `JDBC-PING` for that and for one cluster local discovery is more then enough.

larainema:
Thanks @fitbeard , these results resolve the superuser and Infinispan questions.

For the Secret namespace issue, I suggest deploying the Keycloak-specific CNPG Cluster, Database/role resources, and credential Secret in auth-system alongside the Keycloak CR. The CNPG operator can remain centrally installed. This gives CNPG and Keycloak access to one Secret without cross-namespace replication and keeps Keycloak independent from shared OpenStack databases.

For Kubernetes 1.28, we can test it for compatibility evidence, but I do not think we should enable an unsupported CNPG combination by default. Current CNPG releases do not support or test Kubernetes 1.28. We should either establish a newer Atmosphere Kubernetes baseline or retain an external PostgreSQL path for older deployments.

Could you also confirm whether your PostgreSQL 18 PoC covered the complete Keycloak 24.0.5/PXC export -> PostgreSQL 18 import -> Keycloak 26.6.x start and validation sequence? If so, I agree PostgreSQL 18 should be the target.

Based on this direction, I suggest dropping the database vendor option and Infinispan work from #4134, then replacing or reworking it around the integrated CNPG, official Keycloak Operator, and controlled migration path.

fitbeard:
@larainema thanks.

  - I tested Keycloak 24.0.5/PXC export -> PostgreSQL 18 import -> Keycloak 26.6.x start dozen of times. Zero issues observed.
  - I really like idea of having cloud native approach for databases -  one per namespace or even one per application.

Right now i pushed new changes for CNPG to see how things looks on Kubernetes 1.28 :https://github.com/vexxhost/atmosphere/pull/3314

Also i have working PoC of Keycloak operator managing Keycloak 26.6.4

@mnaser what do you think about dedicated PG instance for Keycloak in auth-system namespace? If this is way to go then which namespace is best to install cnpg-operator?

fitbeard:
@larainema i found set of cnpg components versions which work with 1.28.13 in our CI:

  - The latest version of chart which is not checking min version of Kubernetes is `0.27.1`
  - The highest version of operator which is supported by this chart version is `1.28.3` (there is a newer 'last release of 1.28' image `1.28.4` but this image is incompatible with chart 0.27.1 (`require EndpointSlice permissions`))
  - Barman plugin chart version `0.5.0` and binaries version `0.11.0` are supported with this cnpg version.

Theoretically support matrix mentions 'tested but not supported' versions for -2-3 versions from supported list.
So 1.29 and 1.28 is worth the risk to try.

What do you think?

larainema:
Thanks @fitbeard . The repeated full migration tests and Keycloak Operator PoC resolve the PostgreSQL 18 and Keycloak 26.6.4 questions.

I agree with using a dedicated PostgreSQL cluster in auth-system. I suggest installing the cluster-wide CNPG operator and Barman plugin in a neutral cnpg-system namespace, while keeping the Keycloak Cluster, Database/role resources, credentials, and Keycloak CR together in auth-system.

The Kubernetes 1.28 combination is worth testing, but I do not think we should use CNPG 1.28.3. CNPG 1.28.4 contains security and backup/recovery fixes. The missing EndpointSlice permissions can be added to chart 0.27.1, but we should also vendor the exact 1.28.4 CRDs and update the kubectl plugin/checksums.

Since Kubernetes 1.28 is outside the CNPG-tested matrix and CNPG 1.28 is now EOL, this should be documented as a temporary compatibility exception with a Kubernetes/CNPG upgrade path.

For #3314, I suggest limiting it to the operator foundation, removing the unconditional generic openstack/cnpg-db, and testing the patched 1.28.4 deployment on AIO. The dedicated Keycloak PostgreSQL cluster, backup/restore, official Keycloak Operator, and migration can then follow as the integrated cutover work.

fitbeard:
@larainema i like this idea very much! Here is PR dedicated for operators only: https://github.com/vexxhost/atmosphere/pull/4135 I'll leave old PR for later references(https://github.com/vexxhost/atmosphere/pull/3314). I think we can review bit by bit and in the end combine everything to one PR for operators, Keycloak's dedicated PG, Keycloak's operators and migration path. Sounds?

larainema:
@fitbeard , this direction sounds good. I agree with keeping #3314 for reference and using #4135 as the operator-only foundation.

I suggest keeping the implementation as stacked PRs, while treating the actual database and Keycloak migration as one controlled cutover.

I reviewed #4135 briefly and added a comment.

fitbeard:
@larainema new layer for stacked PR: https://github.com/vexxhost/atmosphere/pull/4136 and second one with fixes from your previous review: https://github.com/vexxhost/atmosphere/pull/4135

btw: nice team work!

PR #4135, opened 2026-07-14
feat: add cnpg operators

fitbeard:


larainema:
@fitbeard One version-alignment issue remains before we use this as the operator foundation.

The operator image is pinned to CNPG 1.28.4, but the vendored Helm chart is 0.27.1 with appVersion: 1.28.1. Its bundled CRDs are older than the [official CNPG 1.28.4 CRDs](https://github.com/cloudnative-pg/cloudnative-pg/blob/v1.28.4/releases/cnpg-1.28.4.yaml).

For example, the vendored CRDs use controller-gen v0.20.0 and do not include the 1.28.4 validation making cluster references immutable for resources such as Database, Pooler, and ScheduledBackup. The official 1.28.4 CRDs use controller-gen v0.21.0 and include those validation updates.

Could we add a chart-vendor patch that replaces the bundled CRDs with the exact 1.28.4 definitions? This would keep the Kubernetes-1.28-compatible chart templates while aligning the CRD schema, RBAC, kubectl plugin, and controller image on CNPG 1.28.4.

We should then validate both a fresh installation and an upgrade/reapply using server-side apply, to ensure the updated CRD schemas are accepted on Kubernetes 1.28.13.

fitbeard:
@larainema this is ready for a next review.

larainema:
@fitbeard Generally LGTM, the CRD alignment concern is addressed. The vendored CRDs match the official CNPG 1.28.4 definitions.

One remaining alignment issue: charts/cloudnative-pg/Chart.yaml still has appVersion: 1.28.1. As a result, Helm and Kubernetes labels report 1.28.1 while the controller image, plugin, and CRDs are 1.28.4; direct chart usage would also default to the 1.28.1 image. Could we include appVersion: 1.28.4 in the vendor patch?

fitbeard:
@larainema i think we need to stop using llms for simple comments:))) Everything looks so fake. No, i can't update Chart.yaml - chart-vendor pipeline is not happy about changes in this file. I tested and changed Chart.yaml.yaml before (llms suggestion btw) and then rolled back.

larainema:
> Dong Ma (@larainema) i think we need to stop using llms for simple comments:))) Everything looks so fake. No, i can't update Chart.yaml - chart-vendor pipeline is not happy about changes in this file. I tested and changed Chart.yaml.yaml before (llms suggestion btw) and then rolled back.

Thanks for clarifying. No further change requested from me.

larainema (approved):
lgtm

mnaser:
Let's get this landed inside https://github.com/vexxhost/atmosphere.common with CI there, so we can exercise CI for it only there.

Also, we're moving towards a pattern of using OCI Helm charts so we can store them there https://github.com/vexxhost/charts -- and maybe figure out a way we can patch it in CI and publish it there.

PR #4136, opened 2026-07-15
feat: install keycloak operator

fitbeard:
Related: https://github.com/vexxhost/atmosphere/pull/4134 and https://github.com/vexxhost/atmosphere/pull/4135

This will be part of stacked PR later.

PR #4137, opened 2026-07-16
feat: Keycloak migration

fitbeard:
Keycloak migration from PXC (24.0.5) to CNPG (26.6.4)

Included cherry-picks: https://github.com/vexxhost/atmosphere/pull/4135 https://github.com/vexxhost/atmosphere/pull/4136

Documentation addition pending with later commits


fitbeard:
@larainema please review when you have a time. `build` and `pre-commit` are failing for not related reasons. Zuul also eventually will fail on tasks with some `community.general.keycloak_xxx` modules (not fixed for now).

fitbeard:
recheck

fitbeard:
recheck

larainema:
Hi @fitbeard . I double-validated this PR at commit 6c28a257 on the AIO using an isolated Keycloak 24.0.5 deployment with users, credentials, roles, groups, events, offline sessions, a confidential client, and FGAP v1 permissions.

The main migration path works: Keycloak 26.6.4 completed the schema upgrades, became Ready, and preserved realms, users, credentials, roles, groups, clients, and password authentication. I also confirmed that the rescue block restores the legacy deployment when a failure occurs after cutover starts.

I found several items that should be addressed before merge:

Offline sessions and event history are not preserved. The source had 2 offline user sessions, 2 offline client sessions, 2 user events, and 7 admin events; all were zero in PostgreSQL after migration. A pre-migration offline token then failed with Offline user session not found, although the verifier passed.

A partially populated PostgreSQL target is not recoverable automatically. On rerun, the preflight assertion fails before entering the rescue block, leaving the legacy StatefulSet at zero replicas and ingress returning HTTP 503.

FGAP v1 behavior is lost. The migrated client’s management-permissions endpoint returned Feature not enabled because the operator-managed configuration enables only token-exchange.

Configurable database names need fixes. keycloak_database_name is also used as the Kubernetes Database.metadata.name, so names containing underscores are rejected. Additionally, spec.db.database is not set on the Keycloak CR, causing custom database names to fall back to keycloak.

I suggest fixing the database-name handling and interrupted-state recovery in this PR, and explicitly deciding how offline sessions, audit/event history, and FGAP v1 configurations will be preserved or handled operationally.

PR #4152, opened 2026-07-23
feat: selective-molecule-ci

fitbeard:

PR #4161, opened 2026-07-25
test(ci): exercise Keystone selective plan

fitbeard:
Temporary validation PR stacked on #4152. Expected selective result: run AIO Open vSwitch and Keycloak; skip AIO OVN, CSI local-path, and CSI RBD. This PR contains only an inert Keystone-path marker and will be closed after CI validation.


fitbeard:
recheck

PR #3811, opened 2026-04-10
feat(ironic): add novncproxy deployment for baremetal VNC console

larainema:
## Summary

Add support for the ironic-novncproxy service introduced in ironic 2025.1 for VNC console access to baremetal nodes.

### Helm chart (5 new files, 2 modified)
- `deployment-novncproxy.yaml`: Deployment with TCP health checks on port 6090
- `service-novncproxy.yaml`: Service for `baremetal_vnc_proxy` endpoint
- `bin/_ironic-novncproxy.sh.tpl`: Startup script
- `role-console-pods.yaml`: RBAC Role for managing console pods/secrets
- `rolebinding-console-pods.yaml`: RoleBinding for ironic-conductor SA
- `configmap-bin.yaml`: Added novncproxy script entry
- `values.yaml`: novncproxy labels, image, network, endpoints, pod config

### Ansible role (3 modified + 1 image config)
- `ironic_vnc_enabled` flag (disabled by default)
- `[vnc]` config section in ironic.conf (kubernetes container provider)
- Conditional VNC Proxy Ingress with WebSocket annotations (3600s timeouts)
- `ironic_novncproxy` image tag in defaults

## How to enable

Set `ironic_vnc_enabled: true` in your inventory. All manifests are disabled by default for backward compatibility.

## Dependencies

- vexxhost/docker-ironic#713 (adds `novnc` package to the ironic image)
- Upstream: https://review.opendev.org/c/openstack/openstack-helm/+/983954

Closes #3748

PR #4232, opened 2026-07-31
feat(ironic): complete Kubernetes console provider support

ricolin:
## Summary

- complete the Ironic Kubernetes VNC console provider integration
- stage noVNC browser assets without modifying application JavaScript at runtime
- provide the dynamic console Pod/Secret template, dedicated-namespace RBAC, and an enabled-by-default RFB NetworkPolicy when the feature is selected
- install `kubectl` in conductor Pods when the Kubernetes provider is enabled
- expose a distinct noVNC endpoint and role configuration through Atmosphere variables
- require operator-supplied console backend overrides to be complete digest references
- make dynamic Pod scheduling, image-pull, probe, resource, and scratch-volume controls configurable
- fail safely when disabling the feature while provider-owned Pods or Secrets remain, and remove the role-owned noVNC Ingress when disable is safe

The feature remains disabled by default. No image catalog entry or image build input is changed by this pull request.

## Why the original design is insufficient

The original provider integration was functional but did not establish a safe production contract:

1. Dynamic BMC credential Secrets and console Pods could fall back to the shared Helm release namespace. The chart now requires an explicit namespace; Atmosphere creates and selects `{{ ironic_helm_release_namespace }}-ironic-vnc` by default.
2. The RFB NetworkPolicy existed but the role did not enable it, leaving TCP/5900 reachable according to the cluster's ambient policy. It is now enabled by default whenever the Kubernetes console provider is enabled, with an explicit opt-out.
3. The noVNC hostname inherited the Ironic API hostname, allowing two Ingresses to claim the same host and path. The default is now derived as a distinct `baremetal-console` hostname and the role validates endpoint separation.
4. Pod existence did not prove that the backend was listening. The generic template now provides TCP startup and readiness probes on the RFB port.
5. Private registries and topology-constrained BMC networks could not be expressed. Image pull secrets, annotations, node selection, affinity, tolerations, topology spread, priority/runtime classes, resources, and scratch volumes are now configurable.
6. Disabling the feature only skipped the noVNC Ingress task. The role now reconciles that Ingress absent and refuses disable while labelled dynamic Pods or Secrets remain.
7. Digest validation previously accepted any string containing `@sha256:`. It now requires a complete 64-character lowercase SHA-256 reference for console backend and overridden browser-assets images.
8. The checked-in chart and its consolidated vendor patch could drift. The patch is regenerated from the chart baseline and verified to reproduce `charts/ironic` byte-for-byte.

The NetworkPolicy is deliberately ingress-only. It permits the Ironic noVNC proxy Pods to connect to dynamic RFB Pods; it does not define OpenStack console authorization or make a console read-only. OpenStack/Ironic authorization remains above this Kubernetes transport boundary.

## Stack and ownership

- this branch incorporates and supersedes the focused Ironic noVNC foundation from #3811 on current `main`
- chart implementation: https://review.opendev.org/c/openstack/openstack-helm/+/997340
- OpenStack-Helm owns the deployer-neutral `console.namespace` template contract
- Atmosphere owns namespace creation/defaulting, endpoint validation, lifecycle reconciliation, and values mapping
- upstream Ironic owns provider-side wait-for-listen and partial-apply cleanup; those are not implemented by this PR

## Validation

- `git diff --check`
- `helm lint charts/ironic`
- Helm 3 strict unit tests: 2 console tests, 2 priority-class tests, and 2 runtime-class tests passed
- default-disabled render contains no console Pod template, console RBAC, or console NetworkPolicy
- missing namespace is rejected when the Kubernetes provider is enabled
- separate-namespace render verifies Secret, Pod, Role, RoleBinding, and NetworkPolicy placement and the `application=ironic,component=novncproxy` source selector
- consolidated Ironic chart patch applies to its recorded baseline and reproduces the checked-in chart byte-for-byte
- chart image tags, Atmosphere image catalog inputs, and image-build inputs are unchanged
- `ansible-lint roles/ironic` reports no failures or warnings

The repository-wide Go test was also run locally. Changed packages passed; unrelated image-manifest tests assume a Linux target and fail on Darwin/arm64, and Prometheus rule tests require `promtool`, which was not installed on the workstation. CI remains authoritative for those platform/tool-dependent suites.


ricolin:
recheck

ricolin:
The latest commit hardens the generic provider contract without adding deployment-specific information or changing image inputs.

Key failure modes addressed are shared-namespace BMC credentials, a NetworkPolicy that was rendered but not selected by the role, API/noVNC hostname collision, lack of protocol-level readiness, missing private-registry and scheduling controls, stale ingress on disable, and weak digest validation. The disable path is intentionally fail-safe: it will not remove the public endpoint while labelled provider resources still exist.

Compatibility was checked explicitly. The feature remains opt-in; existing proxy priority-class and runtime-class tests still pass; default-disabled rendering emits no console provider resources; and the chart and Atmosphere image maps are unchanged. The consolidated vendor patch was regenerated and compared byte-for-byte with the checked-in chart.

The companion OpenStack-Helm review still needs a new Gerrit patch set for the explicit namespace contract. Provider-side wait-for-listen and partial-apply cleanup remain correctly owned by upstream Ironic rather than being hidden in this role.

ricolin:
recheck

PR #4234, opened 2026-08-03
fix: disable missing Ironic log config cleanly

ricolin:
## Category

Production compatibility fix. No emulation, credentials, or environment-specific data is included.

## Summary

- render disabled `log_config_append` as an empty value instead of YAML `null`
- prevent Helm from writing the literal `<no value>` filename into `ironic.conf`
- add role regression coverage and a release note

## Validation

- `helm lint charts/ironic`
- current Ironic runtime parses an empty `log_config_append` as disabled


ricolin:
@codex[agent] review this PR

PR #4237, opened 2026-08-05
feat(ironic): configure Redfish Virtual Media boot

ricolin:
## Summary

- enable the Redfish Virtual Media deployment configuration by default for Redfish-backed Ironic deployments
- advertise `redfish-virtual-media` while the feature is enabled
- keep boot-interface selection explicit per bare metal node, preserving existing PXE and iPXE selections
- provide `ironic_redfish_virtual_media_enabled: false` as the opt-out for deployments that do not enable Redfish
- retain runtime image ownership in the existing `atmosphere_image_overrides` contract

## Why

Redfish is part of the standard Ironic configuration, so its Virtual Media deployment path should work by default rather than require a second opt-in. Enabling the deployment configuration does not change the boot interface already selected on existing nodes; compatible nodes still select `redfish-virtual-media` individually.

While enabled, the role advertises the interface, disables Swift temporary URLs, permits the local media paths, and configures the x86_64 UEFI bootloader ESP and GRUB path. Deployments that do not use the Redfish hardware type can explicitly disable the feature.

Deployments remain responsible for selecting a compatible Ironic image and providing a BMC-reachable media publisher URL.

## Validation

- `git diff --check`
- local `pre-commit` command is unavailable; the PR CI pre-commit job provides that validation
- confirmed the diff contains no downstream repository, image, environment, address, or credential references

### DIFF
--- a/releasenotes/notes/enable-redfish-virtual-media-3da9fb17bb2d391e.yaml
+++ b/releasenotes/notes/enable-redfish-virtual-media-3da9fb17bb2d391e.yaml
@@ -1,0 +1,16 @@
+---
+features:
+  - |
+    Enables the Ironic configuration required for Redfish virtual-media
+    deployment. The role enables the ``redfish-virtual-media`` boot interface,
+    disables Swift temporary URLs, configures the x86_64 UEFI bootloader ESP,
+    and writes GRUB configuration to the Ubuntu signed GRUB prefix.
+
+    Runtime image selection remains managed through
+    ``atmosphere_image_overrides``. Deployments must select an Ironic image
+    that contains the configured bootloader ESP and required virtual-media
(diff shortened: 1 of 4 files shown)


mnaser:
@ricolin does it make sense to flip this to be the default when redfish is in the equation?

ricolin:
Correction: addressed in commit 1d03f043. The Redfish Virtual Media feature now defaults to enabled for Redfish-backed deployments, and both the boot-interface advertisement and the related publisher/UEFI settings follow the same feature flag. Deployments that do not enable Redfish can opt out with `ironic_redfish_virtual_media_enabled: false`. Nodes still select `redfish-virtual-media` individually, so existing PXE/iPXE selections are unchanged.

PR #4242, opened 2026-08-11
fix(octavia): retry transient quota requests

fitbeard:
## What changed

- retry the Octavia administrative compute quota update for up to five minutes;
- add a release note.

## Why

The quota request can race with OpenStack API readiness during deployment. A temporary API failure previously stopped Octavia even when the service recovered moments later.

This production fix is extracted from #4152 so it can be reviewed, released, and backported independently from selective CI.

## Validation

- `git diff --check` passed;


fitbeard:
recheck

PR #4244, opened 2026-08-11
fix(ceph-provisioners): create deployment namespace

fitbeard:
## What changed

- create the configured Ceph provisioners namespace before creating Kubernetes resources;
- add a release note.

## Why

The role assumed its namespace had been created by an earlier deployment step. That hidden ordering dependency breaks focused deployments and makes the role non-self-contained.

This production fix is extracted from #4152 so it can be reviewed, released, and backported independently from selective CI.

## Validation

- `git diff --check` passed;


fitbeard:
recheck

fitbeard:
recheck

fitbeard:
recheck

PR #4247, opened 2026-08-11
fix(rook-ceph-cluster): create service scope

fitbeard:
## What changed

- create the OpenStack `service` domain;
- create or reuse the `service` project before configuring the object storage user;
- add a release note.

## Why

The Rook Ceph cluster role assumed Keystone or another service had already created this identity scope. Focused deployments exposed that hidden ordering dependency.

This production fix is extracted from #4152 so it can be reviewed, released, and backported independently from selective CI. A separate SDK cleanup will follow after this prerequisite merges.

## Validation

- `git diff --check` passed;


fitbeard:
recheck

fitbeard:
recheck

PR #4250, opened 2026-08-11
feat(neutron): make Helm timeout configurable

fitbeard:
## What changed

- add `neutron_helm_timeout` with a five-minute default;
- use it for Neutron Helm operations;
- add a release note.

## Why

Deployments that need more time had no supported way to extend the Helm operation timeout.

This production improvement is extracted from #4152 so it can be reviewed, released, and backported independently from selective CI.

## Validation

- `git diff --check` passed;


fitbeard:
recheck

fitbeard:
recheck

PR #4335, opened 2026-08-22
fix(rook-ceph-cluster): manage identity with SDK

fitbeard:
## What changed

- replace shell-based service project creation with the OpenStack SDK project module;
- replace shell-based role assignment with the OpenStack SDK role-assignment module;
- use the shared service-domain lookup supported by the minimum openstack.cloud 2.0.0 requirement;
- add a concise release note.

## Why

The shell commands always reported no change and made repeated deployments less observable. The SDK modules provide idempotent state management and explicit domain selection.

This is the final production runtime concern extracted from #4152, as requested in review pullrequestreview-4900414469. It follows the service-scope prerequisite merged in #4247.

## Validation

- file hygiene hooks passed;
- Ansible syntax validation reached the Rook Ceph cluster role successfully and reported only the existing repository-wide galaxy no-changelog metadata finding;
- the module arguments were checked against the official openstack.cloud 2.0.0 source required by galaxy.yml;
- git diff --check passed;
- both commits carry the required DCO sign-off.

Depends-On: https://github.com/vexxhost/ansible-collection-ceph/pull/124

### DIFF
--- a/releasenotes/notes/rook-ceph-sdk-role-assignment-87e44518a8f73e7c.yaml
+++ b/releasenotes/notes/rook-ceph-sdk-role-assignment-87e44518a8f73e7c.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+    The Rook Ceph cluster role now manages object storage identity assignments
+    through the OpenStack API, making repeated deployments idempotent.
--- a/roles/rook_ceph_cluster/tasks/main.yml
+++ b/roles/rook_ceph_cluster/tasks/main.yml
@@ -115,11 +115,12 @@
 
 - name: Create OpenStack service project
   run_once: true
-  changed_when: false
-  ansible.builtin.command:
-    cmd: openstack project create --or-show --enable --domain service service
-  environment:
-    OS_CLOUD: atmosphere
+  openstack.cloud.project:
+    cloud: atmosphere
+    name: service
+    domain: service
+    enabled: true
+    state: present
 
 - name: Create OpenStack user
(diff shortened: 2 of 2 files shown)


fitbeard:
recheck

fitbeard:
recheck

fitbeard:
recheck

fitbeard:
CSI-RBD root cause update:

- this is deterministic: four corrected-patchset attempts failed at the same `vexxhost.ceph.mon` task, before Kubernetes, Rook, Keystone, or the modified SDK tasks ran;
- the August 11 passing job and all four failures used `vexxhost.ceph` 4.2.0 and the pinned `v20.2.3` image;
- the difference is the floating `quay.io/ceph/ceph:v20` tag. It moved on August 19 from the `v20.2.3` manifest to a new manifest;
- `cephadm bootstrap` was not receiving `CEPHADM_IMAGE`, so it followed floating `v20`, while the next `orch_apply` task explicitly used `v20.2.3`. That mixed runtime produced the keyring parse failure.

The independent dependency fix is [vexxhost/ansible-collection-ceph#124](https://github.com/vexxhost/ansible-collection-ceph/pull/124). It adds `CEPHADM_IMAGE` with the same resolved image while preserving `CEPH_CONTAINER_IMAGE` and `CEPH_CONTAINER_BINARY`.

I am not sending another blind recheck here. First I will validate the collection fix in its Ceph deployment matrix, then test this PR against that dependency.

fitbeard:
recheck

PR #4374, opened 2026-09-17
fix: use distro Python packages on newer Ubuntu and Debian

mnaser:
On Ubuntu 24.04, the deployment roles try to install OpenStackSDK and PyMySQL into the externally managed system Python with pip. Use the distribution packages on Ubuntu newer than 22 and Debian newer than 12, following the Kubernetes collection and the SDK approach in Atmosphere #3297 / atmosphere.common#114. Keep the existing pip path, version selection, and retries elsewhere.

This change targets `main`; backports can follow after merge. It changes the two role install blocks, documents that `openstacksdk_version` only controls pip installs, and adds a release note. It does not require the Common role migration or the pip overlay in #4139.

Validation:
- Ansible Core 2.18.6 condition evaluation: 10 OS/version cases for each role; exactly one install path selected in every case. Existing pip retries and default/explicit SDK version selection verified.
- Extracted installation tasks: Ansible syntax check and ansible-lint 25.6.1 passed.
- Repository whitespace/EOF/line-ending hooks and `git diff --check` passed.
- Full repository Ansible-lint did not complete: its `ansible-galaxy` dependency preparation stalled locally. CI still needs to run.
- Live package-installation test on Ubuntu 24.04.4 with Ansible 2.18.6: executed the four installation tasks extracted unchanged from this PR against actual host facts. Both distro packages installed and both pip tasks were skipped. System Python imported OpenStackSDK 3.0.0 and PyMySQL 1.0.2 from `/usr/lib/python3/dist-packages`. First successful run: `changed=2, failed=0, skipped=2`; rerun: `changed=0, failed=0, skipped=2`. An initial python3-apt bootstrap attempt hit a sandbox filesystem I/O error; the subsequent installation and rerun succeeded.
- Full OpenStack deployment, service API/database operations, rolling upgrade, and hosts with pre-existing pip overlays remain untested.

Pip audit of both stable/2024.2 and main:
- Replaced: the OpenStackSDK and Keycloak PyMySQL deployment installs.
- Retained: Earthfile and main's deployment instructions install into explicit virtualenvs.
- Retained: `terminate_project.yml` requires a specific ospurge Git revision; replacing it with a distro package would not preserve the dependency. Its Noble packaging needs a separate isolated-install follow-up.
- Retained: optional `keystoneauth-websso` CLI plugin instructions; no matching Noble package was found.
- Identified separately: Molecule's legacy `pip freeze` / uninstall-all cleanup is not an installation and should be reviewed before using that scenario on Noble.


mnaser:
recheck

PR #4379, opened 2026-09-17
fix: require atmosphere.common 0.7.1 for Secretgen rendering

mnaser:
Atmosphere currently permits `atmosphere.common` 0.7.0, which does not include the Secretgen controller manifest rendering fix. Raise the minimum to 0.7.1 so dependency resolution requires a release containing that fix.

Common 0.7.1 includes vexxhost/atmosphere.common#123 and has been published. This PR changes only the collection requirement and its release note, as part of the Ubuntu 24.04 (Noble) support work.

Validation: parsed both changed YAML files and verified the minimum dependency; `git diff --check` passed. Deployment validation is left to CI.


mnaser:
recheck

mnaser:
recheck

PR #4380, opened 2026-09-17
fix(nova): use eVMCS-enabled image and document Windows usage

mnaser:
Bumps the default Nova image on `main` to an image that supports opt-in eVMCS for Windows guests and that can load the os-vif OVS plugin, and documents the Windows workflow.

## Nova image

All 14 `nova_*` entries in `roles/defaults/vars/main.yml` move to:

```
ghcr.io/vexxhost/nova:main@sha256:7904b9c9ce8c07dcc075796237cc800fef0a4b8e2dab835302b86b88b01469e4
```

The Nova version is unchanged (`33.0.2+a8e.10.3`); only the digest moves. Two things are in the image:

- **eVMCS support** (Nova fork commits for `hw_hyperv_evmcs`).
- **`pyroute2`**, via vexxhost/docker-nova#1430.

The pyroute2 part matters because os-vif's code imports it (`os_vif/internal/ip/linux/impl_pyroute2.py`) while its published wheel does not declare it: pbr 7.0.3 parks environment-marked requirements in an `extras_require` pseudo-extra, and setuptools discards that mapping once a project has a `[project]` table, which os-vif gained in 5.0.0. `--constraint /upper-constraints.txt` pins `os_vif===5.1.0`, so the image previously got the wheel without the dependency. The consequence on a `main` deployment:

```
ERROR stevedore.extension [-] Could not load 'ovs': No module named 'pyroute2'
INFO os_vif [-] Loaded VIF plugins: noop
nova.virt.libvirt.vif._plug_os_vif -> KeyError: 'ovs'
Exhausted all hosts available for retrying build failures
```

Only `main` was exposed. Every stable branch pins os-vif 4.x or older, whose wheels still declare pyroute2.

## Documentation

Organized by audience, since the enabling step is a user action while the prerequisites are operator work:

| User guide | Administrator guide |
|---|---|
| Windows feature overview and benefits | Nova rollout and host prerequisites |
| Enabling/disabling eVMCS on an image | Intel VMX, nested virtualization, Placement checks |
| vTPM image properties and instance creation | Verifying vTPM support and configuring flavors |

The Windows user category links to both eVMCS and the shared vTPM guide; vTPM stays outside the Windows category because other guest OSes use it. User and admin pages cross-link, and the existing admin vTPM URL is preserved. Docs state that eVMCS is Intel-only, disclose the posted-interrupt trade-off with benchmarking advice, and note that MBEC is not included.

## Validation

- The published image was inspected directly: it contains `pyroute2-0.8.1.dist-info` next to `os_vif-5.1.0.dist-info`, where the previous image had os-vif alone. Both list the same `nova-33.0.2+a8e.10.3`.
- Two environments built from the same os-vif 5.1.0 wheel, differing only by pyroute2: without it `os_vif.initialize()` logs `Could not load 'ovs': No module named 'pyroute2'` then `Loaded VIF plugins: noop`; with it, `Loaded VIF plugins: noop, ovs`.
- vexxhost/docker-nova#1430 is merged, and the image was built and published on `main`; the build log shows `+ os-vif==5.1.0` and `+ pyroute2==0.8.1` for both architectures.
- All 14 Nova refs use the same digest; YAML parses; `git diff --check` clean.
- Focused Sphinx build of both navigation trees and the linked user/admin topics passes with warnings treated as errors. The full documentation build has pre-existing warnings in `admin/monitoring.rst` (heading underline) and `arch/images.rst` (three missing Dockerfile includes), unrelated to this change.

## Note for review

renovate PR #3501 (`chore(deps): update vexxhost images (main)`) also pins Nova to the previous digest `63053166`, which lacks pyroute2. Whichever of the two lands second should not reintroduce it; #3501 will need regenerating against the new digest.

The eVMCS feature itself has not been validated on real Windows/VBS workloads; that requires a canary on Intel hosts with `hw_hyperv_evmcs=true`.


mnaser:
recheck

PR #4381, opened 2026-09-17
fix(ceph): use the configured runtime image

mnaser:
The CSI-RBD molecule job bootstraps Ceph with the collection default runtime image while the rest of Atmosphere uses the configured Ceph image. This produces an incompatible keyring format. Pass the configured `atmosphere_images[ceph]` image to the Ceph collection so bootstrap and CSI clients use the same runtime.

Depends-On: https://github.com/vexxhost/atmosphere-zuul-jobs/pull/6


mnaser:
recheck

